PR #7, opened 2015-01-14
Fix ResourceLeak warning with CPython3

vmarkovtsev:
See https://bitbucket.org/bspeakmon/jira-python/issue/140/fix-resourceleak-warning-under-python-3

The fact that you closed that issue does not deny the fact that you do not know that requests.Session-s should be explicitly closed.


ssbarnea:
Thanks for the pull.

Issue #9, opened 2015-01-16
add_user is not creating users.


sargentstudley:
After revisting Issue #99 from the old bitbucket.org repository found at https://bitbucket.org/bspeakmon/jira-python/issue/99/add_user-is-not-working-with-ondemand, I've also discovered that this does not work with the server version either. 

I have created a unit test that recreates this issue and will be issuing a pull request shortly.

PR #10, opened 2015-01-16
ssbarnea

sargentstudley:
Related to issue #9 , Added unit test to reproduce the error, and add coverage for add_user.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 from __future__ import print_function
+import json
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


ssbarnea:
While the test will need more work before being valid, I will accept it and fix it. 

FYI: Test are executed on Travis on 4 Python version in parallel against the same JIRA instance... so it will fail. I will make changes like: using a username that is unique to the running instance . Also another change is to try to delete the user before creating it, just in case it did exist due to an failed previous execution.

Anyway, thank you and looking forward to see new merge requests ;)

sargentstudley:
Thanks @ssbarnea for the feedback, and for accepting the request. I am still learning the test environment, so I'll pull what changes you make to the test back into my repository so that I can follow that patterns from this point on. 

Thanks again!

ssbarnea:
There is a problem, you cannot run the tests if you do not have the credentials, send me an email and I will send them to you. 

For travis this is solved but I don't know how to solve it without putting them in the source code.

PR #21, opened 2015-01-30
Raise an AttributeError if a requested attribute doesn't exist within se...

bardal:
...lf or self.raw. This is a bug fix to ensure calls to hasattr return false when the attribute doesn't exist.


bardal:
See issue #28 for related information

PR #57, opened 2015-04-10
Restore issue transitions by fixing some flubs.

MerelyAPseudonym:

PR #59, opened 2015-04-15
Update client.py

jhosmer:
Fixed a typo (`this.transitions` should be `self.transitions`).


jhosmer:
Closing this as I just noticed that a previous pull request (https://github.com/pycontribs/jira/pull/57) already has this fix.

PR #64, opened 2015-04-23
Do not prevent users from adding remote links if we are not able to fetch applicationlinks.

coddingtonbear:


coddingtonbear:
Hrm, it looks like the tests originally failed because I didn't bump the patch number, so I went ahead and did that, but then it failed because the integration tests weren't able to create a project.

It *might* make sense to -- instead of relying upon specific project names existing or being available -- instead randomly generate projects for integration testing; I'd be glad to help out, of course, but can't do much without a JIRA account that's allowed to create projects ;-).

Let me know if you'd like me to do anything specific to help this through.

Cheers!

ssbarnea:
Can you please update the code and retry the pull request?

coddingtonbear:
Still looks like integration tests mashing on one another; anything else I can do to help, here, @ssbarnea?

coddingtonbear:
Hey there, @ssbarnea, let me know if there's anything I can do to help this one along; I'd be glad to cause another change in hopes that that the next travis build would be successful, but only if you ask me to do so.

awurster:
@coddingtonbear - i created something similar actually.. but more of a short circuit of this issue.  not sure why the build is failing - might be a paste issue (i just edited this on the fly in github) but it works locally here on another branch i've got.

maybe it helps you guys...?  https://github.com/pycontribs/jira/pull/79

coddingtonbear:
Yeah, the error we're encountering isn't a bug in my patch, it looks like it's a bug in the integration testing suite (it's complaining about a test project already existing), but thanks for the +1, @awurster; cheers!

PR #67, opened 2015-04-30
Fix issue#66: Don't always set oauth flag, to prevent oauth mode from always being enabled

fastcat:
... even if basic auth is requested


daryltucker:
Thanks for the fix!

PR #78, opened 2015-05-24
JWT authentication support for Atlassian Connect

fikander:
jirashell using jirashell.ini configured with authentication other than oauth didn't work


fikander:
Added JWT support that comes with couple of new optional dependencies (jwt, requests-jwt).
Added support for 'myself' API and JIRA construction parameter to disable server version check.
This would need update of documentation at http://jira.readthedocs.org/en/latest/

PR #92, opened 2015-06-19
Fix jirashell.py for non-oauth authentication

### LINKED ISSUE #91: jiraclient is broken for non-oauth logins by commit 292597c
See https://github.com/pycontribs/jira/commit/292597c573e976c7da42ce570a81086fae301166#diff-30139c45274d7869e2ee349578b0e30cR247

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22477615-jiraclient-is-broken-for-non-oauth-logins-by-commit-292597c?utm_campaign=plugin&utm_content=tracker%2F3650997&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3650997&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asqui:
This is a minor fix for commit 292597c

Fixes #91

### DIFF
--- a/jira/jirashell.py
+++ b/jira/jirashell.py
@@ -244,6 +244,8 @@ def main():
     if oauth['oauth_dance']:
         oauth = oauth_dance(
             options['server'], oauth['consumer_key'], oauth['key_cert'], oauth['print_tokens'], options['verify'])
+    else:
+        oauth = None
 
     jira = JIRA(options=options, basic_auth=basic_auth, oauth=oauth)
 


milo-minderbinder:
This pull request is duplicate with PR #67-"Fix issue#66: Don't always set oauth flag, to prevent oauth mode from always being enabled," and should not be merged, since it actually breaks the case where a user has already provided an access token & access token secret via command line options.

asqui:
@milo-minderbinder : Thanks for pointing this out -- I'm sorry I missed the duplicate issue/pull-request.

Don't panic -- looks like the merge commit (ec0516a) ended up being no-op so this broken change has not actually been merged! (I would have expected the pull request to have rejected in this case, rather than merged with a no-op commit!)

milo-minderbinder:
@asqui - no worries, for a small part of the package, this module has a number of issues, and a lot of the fixes that get merged end up stepping on or breaking something else. It probably wouldn't hurt to write some test cases for this!

And I ended up submitting a seperate pull request anyway, since, while this merge request was a no-op, someone else actually submitted the same line change as part of their own pull request, which *was* merged :-P

Really, the config file parsing needs to be fixed to account for the fact that some config options should be treated as booleans, otherwise the "if oath['oauth_dance']" test will always evaluate to True if that option is only specified in the jirashell.init config file, since options are treated as strings.

PR #98, opened 2015-07-15
Make `python setup.py test` work without first needing to install any…

dbaxa:
… dependencies.


ssbarnea:
Thank you!

PR #104, opened 2015-07-18
Skip character encoding detection when attempting to check for non-empty response.

coddingtonbear:
This PR changes response handling to use the response's `.content` attribute rather than `.text` when attempting to check for non-empty responses.  Using `.text` will cause the requests library to attempt to infer the character encoding of the returned document, and not only is that unnecessary when checking for non-empty responses, the character encoding detection process may take non-trivial amounts of time to return (sometimes as long as minutes) on certain kinds of binary files.

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -29,7 +29,7 @@ class ResilientSession(Session):
             if response.status_code in [502, 503, 504]:
                 return True
             elif not (response.status_code == 200 and
-                      len(response.text) == 0 and
+                      len(response.content) == 0 and
                       'X-Seraph-LoginReason' in response.headers and
                       'AUTHENTICATED_FAILED' in response.headers['X-Seraph-LoginReason']):
                 return False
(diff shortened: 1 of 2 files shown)


ssbarnea:
My impression is that this could break Py3 compatibility and it would be against W3C standards. If you think I am wrong and have a better approach let me know.

coddingtonbear:
1. Could you elaborate on how this could possibly break Py3 compatibility?  In Python 3, both `bytes` and `string` objects have a length, and there are no string encodings that I'm aware of that might be able to encode a non-zero string in zero bytes.
2. Could you elaborate on how this might be against any W3C standard?  This changes only situations in which the Jira plugin was checking that the response had **any** content -- not any situations in which this was **using** the content returned.

To be as absolutely clear as I can: this Jira library is 100% broken for downloading any large *binary* attachments from issues unless this PR (or a similar one) is merged.  Binary attachments, as I'm sure you're aware, do not have a character encoding, but the code in question (using `.text` instead of `.content`) will cause the process to attempt to detect the content's encoding, even if one doesn't exist, and that can take a [very, very long time](https://github.com/kennethreitz/requests/issues/2359).

Since those of us who aren't project members cannot re-open issues, I've opened #115 for you so you can keep track of this bug.

mrts:
@ssbarnea, I'm afraid @coddingtonbear is right - there is no downside in using `content` instead of `text`. `content` is just unencoded binary response content, see [requests docs](http://docs.python-requests.org/en/latest/user/quickstart/#binary-response-content). It has nothing to do with W3C standards and works well in Py3 as well.

ssbarnea:
This was already fixed on develop not need to merge.

Issue #105, opened 2015-07-22
Set check_update to False by default


metashock:
Currently the value is set to True by default which means it will issue an HTTPS request to pypi each time the client is initiated plus it polutes the output with the following misleading comment:

    You are running an outdated version of JIRA Python %s. Current version is %s. Do not file any bugs against older versions.

Misleading because the version is not outdated, it is the version I've tested my software against and the users MUST NOT upgrading it as long as they want support for my software. Issuing an HTTPS request on client instantiation does not makes sense IMHO - from the technical perspective as from the privacy perspective. I don't want that someone can track when I call a program which is using the library and I also don't want a program that performs an HTTPS request before it is starting to do it's millisecond work plus emiting a warning that it cannot reach the internet where it is meant to run as an intranet application.

To specify requirements with a specific version using requirements.txt and then running it in a virtualenv is one of the common ways to deploy python applications. IMHO the version check is contra productive because of this and I've never seen a library which is doing such a thing.

Please set it to `False` by default or remove it completely.

asqui:
Seconded

ssbarnea:
As long GitHub will not provide a way to to prevent people from raising bugs that were identified in an old version I am forced to use this this approach. So far 2 out of 3 issues were already solved or duplicate and I do not have the time to deal with all of these.

Please note that I am open to suggestions and contributions. Note that current behaviour does allow advanced users to disable this functionality.

PR #106, opened 2015-07-22
Deactivated version check by default

metashock:
Visit: https://github.com/pycontribs/jira/issues/105#issuecomment-123782843

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -143,7 +143,7 @@ class JIRA(object):
         "resilient": True,
         "async": False,
         "client_cert": None,
-        "check_update": True,
+        "check_update": False,
         "headers": {
             'X-Atlassian-Token': 'no-check',
             'Cache-Control': 'no-cache',


ssbarnea:
I do not want to disable this by default because we got too many false bug reports just because people were using outdated versions.

I do still understand that this may be undesired in some cases for performance issues or because of not having internet access. 

If you do have other proposals, I am still open to evaluate them. For example I was considering caching the response and not checking more often than once a day, week.

Please understand that nobody is paying me to support this so I need to find a way to minimize the support load by preventing false bugs from being raised. I do not want to end-up making the issue tracker readonly.

metashock:
@ssbarnea  I see the spare time issue. However, I still think that this check is too dirty. 

Btw, I'm offering (again) my help here. I'm not going about to change something in design, I would just have a look at obvious issues and merging pull requests regarding such issues plus helping to find a stable attempt for the release management. (This turned out to be a problem in other bug reports)

PR #113, opened 2015-07-30
Issue 112 - sendEmail kwarg doesn't work when calling add_user method

fender4645:
Removed the 'sendEmal' parameter from the add_user method as the REST call will just ignore this.


fender4645:
The correct parameter for this is 'notification' -- pull request to add this is at https://github.com/pycontribs/jira/pull/119

PR #121, opened 2015-09-11
Bugfixes for ResilientSession retry logic

asqui:
* max_reties should mean retries, not attempts
* If all attempts fail, re-raise the last exception (instead of
    returning None!?)
* Don't call raise_on_error()  with an Exception, it needs a response!


dbaxa:
@asqui hopefully we can drop most of the `ResilientSession` logic soon.

asqui:
@dbaxa Oh? In favour of what?

dbaxa:
@asqui hopefully not needing to use it (when JIRA is changed to do the right thing TM).

dbaxa:
@asqui it looks like the latest build for this pull request failed.

PR #122, opened 2015-09-15
Don't limit search_issues() results by default

asqui:
The default maxResults=50 is a hazard that we ran in to
because if you're not explicitly specifying maxResults
you don't expect to have to handle paging.

Since the implementation neatly handles batch retrieval
of all results, it seems appropriate to use this logic
to get all search results by default unless you explicitly
limit this by setting maxResults.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1550,7 +1550,7 @@ class JIRA(object):
 
     # Search
 
-    def search_issues(self, jql_str, startAt=0, maxResults=50, validate_query=True, fields=None, expand=None,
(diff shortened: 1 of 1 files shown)


ssbarnea:
I am not ok with this because if you get lots of results is usually a sign that you are doing something wrong and from my experience 9/10 users do hammer jira without knowing about the side effects of asking for 10k, 100k issues. 

For this reason the default will remain the same as the one in Atlassian API. 

Still, if you do not like this, I will accept a pull request that is changing the default value when you instantiate the JIRA() class. Still, the default will still be 50, but at least you will be able to specify a single value, without having to change each request.

PR #128, opened 2015-10-19
Fix issue ranking

joshfriend:
Current version throws an `AttributeError` when calling `rank()` because the `_rank` property does not exist for the `JIRA` class unless you monkey-patch it in. `GreenHopper` defines it correctly, but that class is deprecated.

Also, the `gh-global-rank` field is now obsolete, which caused errors for me when I tried to rank issues with that field. Using the newer `gh-lexo-rank` fixed this for me.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -227,6 +227,8 @@ class JIRA(object):
 
         self._options.update(options)
 
+        self._rank = None
+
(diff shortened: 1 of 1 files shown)


joshfriend:
Why is `setup.py prerelease` running on PR branches?

dbaxa:
@joshfriend good question.

PR #129, opened 2015-10-24
ResilientClient fix & exponential backoff

nbaggott:
@ssbarnea: First, thanks for your work on jira-python. We use it extensively at LinkedIn. 

A little too extensively it would seem. While debugging a recent JIRA outage (caused in part by my automation), I discovered that the ResilientClient isn't sleeping for the retryable 5xx status codes. While fixing this I added exponential backup with jitter. The jitter is helpful when you have a thundering herd effect. I tested the jitter with 50 threads making requests to a JIRA instance behind an nginx proxy that only allows 3 concurrent requests. W/out jitter 19 threads completed and with jitter 35 completed.

----------

Restore missing delay in ResilientClient and implement exponential backoff

This addresses two issues related to high concurrency/load.
1.) We weren't sleeping for the 502, 503, 504 statuses, resulting in immediate retry.
2.) Adds jitter to reduce thundering-herd effect that can happen when there is a sudden load at high concurrency.

The current delay logic results in a max of 60 (10 + 20 + 30) seconds of delay before final success/failure.

The jittered delay in this patch will result in a max of 120 seconds of delay (20 + 40 + 60) but each delay is multiplied by a random number between 0 and 1 so the average delay is still 60 seconds. This implementation is the one described as "full jitter" at http://www.awsarchitectureblog.com/2015/03/backoff.html.

Also a minor formatting change in the log message for 5xx case. "Err: <Response [503]>" is now "Err: 503 Service Temporarily Unavailable". Makes the logs easier to understand and might be important information if there is a customized error message (e.g. from a proxy).

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from requests import Session
 from requests.exceptions import ConnectionError
 import logging
+import random
 import time
 import json
 from .exceptions import JIRAError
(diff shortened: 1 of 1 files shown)


ssbarnea:
@nbaggott now you have full control over the repository, feel free to add your contributions. Sorry for the delay on granting you access.

PR #130, opened 2015-10-27
Fix for python 3.

Achimh3011:
Using six to make code work both for python2 and 3. Should be straight-forward.


dbaxa:
@Achimh3011 it looks like the build is failing with your change.

Achimh3011:
It fails the same way as any other build currently and this failing has nothing to to with the code being committed.

Maybe someone a little more intimate with the travis config can find out what is wrong with the `curl` invocation. Did the URL change?

PR #140, opened 2015-11-13
Added type hints to Issue class.

ak-ambi:
When learning jira library, I really missed some hints to learn the available fields. I came up with an idea, how to enable type hints functionality from PyCharm and other IDEs.

This is just an example implementation for Issue class to raise a discussion. I would be very thankful for comments and improvements proposals.

Note: private class _IssueFields is not used anywhere, it is only there to enable hints for issue.fields variable.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -349,15 +349,35 @@ class Filter(Resource):
         if raw:
             self._parse_raw(raw)
 
-
 class Issue(Resource):
-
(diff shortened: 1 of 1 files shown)


ak-ambi:
Ooops... I added the next commit to master branch after creating this pull request. I created separate PL for the 2nd commit.

ssbarnea:
Thank you! Really useful.

PR #142, opened 2015-11-13
Added JIRA._fetch_pages function and used it in all functions using p…

ak-ambi:
…agination and extended ResultList class.

Thanks to that, for all these functions items may be fetched in batches.
Fixed some issues found:
- JIRA.search_issues function was throwing an exception on python 3.x because of basestring type used.
- 'total' key is optional in paginated response, but was always expected.
- some parameters in few functions were not documented

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -96,15 +96,19 @@ def translate_resource_args(func):
 
     return wrapper
 
-
 class ResultList(list):
 
-    def __init__(self, iterable=None, _total=None):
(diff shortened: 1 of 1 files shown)


ak-ambi:
I would not bother about adding full download of all items to remaining functions, but I need _fetch_pages function because of support for new public JIRA Agile API. I am working on it now, and will create the next pull request when this one is merged.

ak-ambi:
Rebased

PR #149, opened 2015-11-21
fix_project_template_determination

Achimh3011:
This enables travis to actually run some tests (though a lot are failing). Culprit was a changed project template structure with JIRA that blocked the creation of the sample projects. 

Also one very obvious error introduced in the last weeks which prevented any test from running was fixed.


ak-ambi:
Yeah, Ialso spotted Resource._get_url problem. Look at my PRs, maybe pulling mycommits will save you some time.

jira/client.py
```diff
@@ -96,6 +96,17 @@ def wrapper(*args, **kwargs):
 
     return wrapper
 
+
+def _get_template_list(data):
```
dbaxa: Personally I would consider yielding templates instead of returning a list here. Users of the method could then call `list()` on the method result if we do not provide a `_get_template_list` method.

Achimh3011: Yes, it's a bit nicer with a generator. Will update PR later.

Achimh3011:
BUMP

Achimh3011:
BUMP^2

PR #152, opened 2015-11-24
Small fixes

filmackay:


Achimh3011:
Could not agree more with your fixes, they are the same in my PRs.

PR #169, opened 2016-01-25
fix non-ascii resource name error for python 2 by skipping if the res…

molog:
…ource name is non-ascii


ssbarnea:
Sorry but there is no way I would accept this patch which is just hiding an error. The right approach would be to encode it properly.

If you can provide me with an test case that replicates this it would be great as I would be able to fix it properly, and also to improve the unittest by adding this test.

PR #172, opened 2016-02-06
resolves issue where incompleted_issues() was failing for missing key…

casahome2000:
…; 'incompletedIssues' corrected to 'issuesNotCompletedInCurrentSprint'


arhip:
hi, when this fix is going to be released ?

PR #175, opened 2016-02-19
Adds support for Kerberos auth.

tuukkamustonen:
Couple of questions:

1. requests_kerberos package is marked compatible with python 2.6+ and 3.4+ which is a subset of the support of jira project. It might work on older versions... or not. Should I remove the package from requirements and make it optional (I see `requests-oauthlib` import happens within `_create_oauth_session()` but it's still declared in requirements file so that confuses me a little)?

2. Not directly possible to add tests for this, as there's no Kerberos server to authenticate against. Then again, it's simply `auth=HTTPKerberosAuth()` that this change introduces and `requests_kerberos` package is supposed to test that functionality anyway. Are we fine without tests?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -63,6 +63,10 @@ requests-oauthlib
 -----------------
 Used to implement OAuth. The latest version as of this writing is 0.3.3.
 
+requests-kerberos
+-----------------
+Used to implement Kerberos.
+
 IPython
 -------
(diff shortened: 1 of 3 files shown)


tuukkamustonen:
Well, not sure why travis jobs failed, on my local machine they pass jusk ok.

Care to take a look @ssbarnea ?

PR #181, opened 2016-03-17
added functionality for backing up from Cloud version

guymatz:
Please be merciless with your review


guymatz:
@ssbarnea  do you have any idea how I made the build fail?

PR #187, opened 2016-04-11
Split up documentation into multiple pages

jwodder:
Having all of the documentation on a single page is very unwieldy, so I split it up and gave every top-level section its own page.  I also promoted the API Documentation section up a level, as presumably it wasn't supposed to be under "Contributing," and added subsection headers for all of the classes.


ssbarnea:
Can you fix the conflicts so I can merge it?

jwodder:
Conflicts fixed.

PR #195, opened 2016-05-01
Use '.content' rather than '.text' when attempting to check for non-e…

ssbarnea:
…mpty response to avoid character encoding detection on binary files.


ssbarnea:
This was already fixed on develop

Issue #189, opened 2016-04-18
Get Watchers List


Simwar:
Hello,

How do I get the list of watchers for an issue?

I see there is a resource but I don't know what to do from there.

grundic:
```python
issue = jira_instance.issue('TST-123)
for watcher in jira_instance.watchers(issue).watchers:
    print(watcher)
```

ssbarnea:
If this would become a PR with documentation update it would be great :)

JSinks:
This issue should be closed now that the documentation has been updated?

hdost:
@ssbarnea I think we can close this issue

asqui:
Looks like it's already closed ;-)

Issue #184, opened 2016-03-29
Cannot get list of attachments


patrick-bark:
Hey, I am trying to get a list of attachments associated with an issue. From what I am reading in the docs that should be `issue.fields.attachment`, however that field is not present in the issue even though I know that the ticket has attachments. Any ideas here?
##

grundic:
```python
issue = jira_client.issue('TST-123')
for attachment in issue.fields.attachment:
    print(attachment)
```

you can verify that issue has an attachment with JQL
```sql
project = TST and attachments is not EMPTY
```

hdost:
@ssbarnea I think we can close this issue

hdost:
@patrick-bark This issue should be fixed. feel free to re-open.

PR #196, opened 2016-05-04
Documentation improvement

grundic:
Hi, I added some documentation :)
It should help with #189 and #184.

Also small note: `add_watcher` does not return anything, but `remove_watcher`. Should it be aligned?


ssbarnea:
Thank you!

ssbarnea:
That is to be expected because attachments are not returned by default when you instantiate a jira issues.

PR #201, opened 2016-05-15
Delete Issue Link Bug Fix

dilshanraja:
Fixed incorrect variable reference -> self._session instead of self.jira._session


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/201" target="_blank">ReviewNinja

PR #210, opened 2016-06-02
Added template_name parameter to create_project to be able to specify…

arthur-kovalenko:
… template directly, and fix issues where function cannot find suitable project name

This parameter allows to specify directly what project template should be used for the newly created project. Without it, create_project function does not work on latest versions of Jira (v7.17).


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/210" target="_blank">ReviewNinja

PR #227, opened 2016-06-11
Fixes to get 'python setup.py develop' working

lechat:
* Removed setup.py dependency on library itself
* Moved version number to setup.py
* Changed package version check to use pkg_resources
* Updated release script to increment version in setup.py


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/227" target="_blank">ReviewNinja

PR #232, opened 2016-06-26
add ability to use request session proxy functionality

avinson:
This is useful if you require a proxy to connect to Jira. The proxy must be set before any connection is initiated or the module will hang.


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/232" target="_blank">ReviewNinja

PR #239, opened 2016-07-14
Moved all tests into separate files and converted them to pytest

lechat:
Now, the problem of having singleton test manager class has disappeared as this class is initialized as session fixture.

Converted all unittest style test methods to pytest asserts.

It looks like Travis builds have better chance to pass when only one build is running at a time, so I set that option it Travis.


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/239" target="_blank">ReviewNinja

ssbarnea:
The trick is to write the test in such way that they WILL be able to run in parallel on Travis. 
We cannot wait for them to run sequentially for two reasons: feedback loop will be to long and 2nd that Travis has a timeout for job execution.

The Travis-after-all was added there in order to deal with this and getting rid of it will not sort the problem. We need to write the tests in such way that they will not fail even if 3-4 of them are running in parallel. That's why we do use project names that include each python version and the current username, so the operations executed by one test session will not affect another session.

lechat:
Observing the failures it doesn't look like the tests are bad, but it looks cloud can't react on rate of changes we're introducing by tests.

Travis was timing out job execution on test manager initialization purely because there was a loop with sleep() in it to wait till cloud jira wake up and create project. I removed it together with  ``travis_wait`` and it seems to be working fine. 

I'd rather not have ``travis_wait`` because it forces to wait for all tests to finish to see the result of the tests. In many cases it is faster to abort the test immediately when there is something failing.

ssbarnea:
Closing it for the moment  due to conflicts and stability issues regarding JIRA cloud instance used for testing. Once we are stable feel free to attempt it again but I would suggest making the transition gradual, like moving only some tests to another files instead of such a huge change which is too hard to review, test and merge.

PR #245, opened 2016-08-09
Add function to add user and group to roles

hillairet:
I want to create automatically a JIRA project with all the proper permissions so I need to add users and groups to specific roles and they will change everytime. So I added two separate functions to add separately a user or a group to a given role.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1667,6 +1667,51 @@ class JIRA(object):
 
     # Resolutions
 
+    def add_user_to_role(self, project, role, username):
+        """
+        Add the user ``username`` to the ``role`` in ``project``
+
+        :param project: ID or key of the project to get the role from
+        :param role: ID of the role where the user will be added
+        :param username: The user to add to the role.
+        """
+        data = {}
+        data['user'] = [username]
+        if isinstance(role, Number):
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/245" target="_blank">ReviewNinja

jira/client.py
```diff
@@ -1667,6 +1667,51 @@ def project_role(self, project, id):
...
+        url = self._get_url('project/' + project + '/role/' + role)
+
+        r = self._session.post(
+            url, data=json.dumps(data))
+        if r.status_code == 200:
+            return True
+        else:
+            logging.warning(
+                'Failed to add user %s to role. Got code: %s' % (username, r.status_code))
+            return r.status_code
```
lechat: This function returns `bool` or `int` as a result. I think that function result must be consistent - either all time returns `bool` or `int`.

lechat:
There is a `Role` class in `jira.resources` that has similar functionality and it is better to add missing functionality to that class.

hillairet:
Ok that makes sense. I didn't see this method in the Role class.

starryknight64:
Where is the existing functionality?

Issue #253, opened 2016-08-18
search_issues fails ~1/50 with KeyError in _fetch_pages


michaelk-stratoscale:
When calling search_issues, this code (in _fetch_pages), occasionally fails with resource returns as a dict with no 'issues' key.

```
resource = self._get_json(request_path, params=page_params, base=base)
next_items_page = [item_type(self._options, self._session, raw_issue_json) for raw_issue_json in (resource[items_key] if items_key else resource)]
```

```
/usr/lib/python2.7/site-packages/jira/client.pyc in search_issues(self, jql_str, startAt, maxResults, validate_query, fields, expand, json_result)
   1728             return self._get_json('search', params=search_params)
   1729 
-> 1730         issues = self._fetch_pages(Issue, 'issues', 'search', startAt, maxResults, search_params)
   1731 
   1732         if untranslate:

/usr/lib/python2.7/site-packages/jira/client.pyc in _fetch_pages(self, item_type, items_key, request_path, startAt, maxResults, params, base)
    382         resource = self._get_json(request_path, params=page_params, base=base)
    383         next_items_page = [item_type(self._options, self._session, raw_issue_json) for raw_issue_json in
--> 384                            (resource[items_key] if items_key else resource)]
    385         items = next_items_page
    386 

KeyError: u'issues'
```

Script that reproduces the issue quite consistently (~1/50):

```
import jira
i = 0
jiraClient = jira.JIRA('https://XXXXX.atlassian.net', basic_auth=('XXXXX', 'XXXXX'))
jql = "XXXXXXXXXXXXXXXXX"
while True:
    try:
        jiraClient.search_issues(jql)
        i += 1
    except Exception as e:
        print e
        print i
```

michaelk-stratoscale:
BTW, to bypass the problem i've downgraded to 1.0.3 which does not have the issue.

Fr3dBear:
1.0.7 seems to have a problem with the way it uses the search query in regards to the max returned results.

I had a similar problem: https://github.com/pycontribs/jira/issues/229

michaelk-stratoscale:
@Fr3dBear , i fail to see how these two are related (except for both being in search_issues method, and found in 1.0.7). You described a problem where a URL is malformed every time maxResults was set to False. The issue i'm reporting here is a sporadic failure to retrieve search results.

Fr3dBear:
[...]

Issue #229, opened 2016-06-16
404 error with "search_issues"


Fr3dBear:
I updated the jira version to: 1.0.7.dev20160607111203 for one of the scripts i was using to fix another bug issue, however i was unable to get the script to run do to a 404 issue when using the "search_issues" function.

Here is the line:
`issueList = authedJira.search_issues(jql_str="project = " + str(projectKey) + " AND issuetype = \"Feature Traceability Epic\" order by createdDate ASC",startAt=0,maxResults=False,validate_query=True,fields=None,expand=None,json_result=None)`

This line works fine in version 1.0.3 but not in 1.0.7

oechslein:
The problem is the "maxResults=False", it leads with 1.0.7 (also with the released one!) to an URL with "&maxResults=False" at the end. This results into a 404.

Example code:
from jira import JIRA

jira = JIRA('https://jira.atlassian.com')

for issue in jira.search_issues("resolution is not empty", maxResults=False):
    print issue

oechslein:
Problem is in _fetch_pages called from search_issues. parameter params has maxResults set to False, if this is cleared then code works again. Instead of 
        page_params = params.copy() if params else {}
        if startAt:
            page_params['startAt'] = startAt
        if maxResults:
            page_params['maxResults'] = maxResults
use the following (not the cleanest solution):
        page_params = params.copy() if params else {}
        if startAt:
            page_params['startAt'] = startAt
        if maxResults:
            page_params['maxResults'] = maxResults
        elif 'maxResults' in page_params:
            del page_params['maxResults']

georgyo:
@oechslein Fix is decent but partial. The code also queries Jira for a parameter called isLast which does not exist normally in Jira. This results in only one page of results being returned instead of all of them. 

See #275 for more.

hdost:
@ssbarnea Correct me if I'm wrong on this one, but as far as I can tell the change about setting `maxResults=false` was pretty intentional. 

I don't know that there will be a fix for this. There needs to be a cap on the lira side. Essentially what the API previously did was wrap up a loop which would retrieve all possible results. This is not really what an API should be designed to do. Therefore it looks like it was removed.

PR #262, opened 2016-09-25
fix for #229 and #253 by adding a fetch_all argument...

qeomash:
Added a fetch_all argument to search functions, and fixed behavior of maxResults=False in _fetch_pages(). This is a fix for #229 and #253.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -358,7 +358,8 @@ class JIRA(object):
             return False
         return True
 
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/262" target="_blank">ReviewNinja

ssbarnea:
This change just makes the API harder to use. We already have this functionality implemented: if maxResults is False it will return all results.

PR #248, opened 2016-08-12
params object for _fetch_pages should not contain maxResults

jeckhart:
This fixes issues if you pass False into jira.search_issues(maxResults=False).

PR #267, opened 2016-10-09
Partially fix falsy maxResults handling in search_issues

mmeisel:
This is the same fix as #248, but that PR (a) will no longer merge cleanly, and (b) missed the removal of the `startAt` parameter, which also should not be passed according to the comment in `_fetch_pages()`.

~~However, I don't think there's any bug here since the values passed to `_fetch_pages()` are the same as the ones provided in the `params`. So this is really just a bit of cleanup.~~

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1717,8 +1717,6 @@ class JIRA(object):
 
         search_params = {
             "jql": jql_str,
-            "startAt": startAt,
-            "maxResults": maxResults,
             "validateQuery": validate_query,
             "fields": fields,
             "expand": expand}


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/267" target="_blank">ReviewNinja

mmeisel:
Actually, I take it back -- there is a bug here. Without this change, if I try to use `maxResults=False`, the parameter is actually sent to the JIRA API, and I get an error:
```
JIRAError: JiraError HTTP 404 url: https://thousandeyes.atlassian.net/rest/api/2/search?jql=category%3DEngineering+and+estimate+is+not+empty&validateQuery=True&startAt=0&maxResults=False
```

After applying this change, there is no error, but it still only returns 50 results. So this change is necessary, but not sufficient to fix the bug.

ssbarnea:
I am closing this as being incomplete, also it does not contain any new test. Feel free to reopen it when is ready.

PR #268, opened 2016-10-12
Issue field dictionary lookup

fundthmcalculus:
Added support for Issue field indirect addressing for custom fields instead of fields.customfield_XXXXX property.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -320,6 +320,9 @@ class JIRA(object):
                 for name in f['clauseNames']:
                     self._fields[name] = f['id']
 
+        self.field_map = {}
+        self.generate_field_mapper()
+
     def _check_update_(self):
         # check if the current version of the library is outdated
         try:
@@ -3149,6 +3152,34 @@ class JIRA(object):
             raise NotImplementedError('No API for moving issues to backlog for agile_rest_path="%s"' %
                                       self._options['agile_rest_path'])
 
+    def map_custom_python_name(self, customfieldname=None, pythonfieldname=None):
+        python_to_custom = self.field_map
+        custom_to_python = [(v,k) for k,v in python_to_custom]
+
+        try:
+            if customfieldname is not None:
+                return list_lookup(custom_to_python, customfieldname)
+            elif pythonfieldname is not None:
+                return list_lookup(python_to_custom, pythonfieldname)
+            else:
+                raise KeyError("Need to provide a custom field name, or a python field name.")
+        except:
+            if customfieldname is not None:
+                return customfieldname
+            else:
+                return pythonfieldname
+
+    def generate_field_mapper(self):
+        # Process the custom fields
+        customfields = self._fields
+
+        iscustomfield = lambda k,v: not re.match('cf\[\d*\]', k) and re.match('customfield_[\d*]', v)
+
+        # Strip out the invalid identifier characters and the customfield indexing.
+        pythonidents = [(python_identifier(k),v) for k, v in customfields.items()
+                        if iscustomfield(k,v)]
+        self.field_map = pythonidents
+
 
 class GreenHopper(JIRA):
 
@@ -3157,3 +3188,13 @@ class GreenHopper(JIRA):
             "GreenHopper() class is deprecated, just use JIRA() instead.", DeprecationWarning)
         JIRA.__init__(
             self, options=options, basic_auth=basic_auth, oauth=oauth, async=async)
+
+
+def python_identifier(fieldstring):
+    return re.sub('[^0-9a-zA-Z_]', '', fieldstring.replace(' ', '_')).lower()
+
+def list_lookup(lst, key):
+    for k,v in lst:
+        if k == key:
+            return v
+    raise KeyError("key=`{0}` not found".format(key))
--- a/jira/client.py
+++ b/jira/client.py
@@ -284,7 +284,7 @@ class JIRA(object):
             self._create_kerberos_session()
         else:
             verify = self._options['verify']
-            self._session = ResilientSession()
+            self._session = ResilientSession(self)
             self._session.verify = verify
         self._session.headers.update(self._options['headers'])
 
@@ -2098,7 +2098,7 @@ class JIRA(object):
     # Utilities
     def _create_http_basic_session(self, username, password):
         verify = self._options['verify']
-        self._session = ResilientSession()
+        self._session = ResilientSession(self)
         self._session.verify = verify
         self._session.auth = (username, password)
         self._session.cert = self._options['client_cert']
@@ -2115,7 +2115,7 @@ class JIRA(object):
             signature_method=SIGNATURE_RSA,
             resource_owner_key=oauth['access_token'],
             resource_owner_secret=oauth['access_token_secret'])
-        self._session = ResilientSession()
+        self._session = ResilientSession(self)
         self._session.verify = verify
         self._session.auth = oauth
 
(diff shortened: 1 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/268" target="_blank">ReviewNinja

jira/resources.py
```diff
@@ -482,6 +482,13 @@ def permalink(self):
     def __eq__(self, other):
         return self.id == other.id
 
+    def __getitem__(self, item):
+        # Grab the JIRA server
+        jira_server = self._session.jira_server
+
+        # Look for the appropriate fields list on the server
+        return getattr(self.fields, jira_server._fields[item])
```
asqui: One thought: Would it be worth expanding out this list on construction and populating all the custom fields as attributes of `self.fields` on construction?

This way you don't have to know if a given field is a custom field or not -- they will be treated equally.

Also means that you would able to see in the debugger what custom fields are available in an `Issue`, which is a little bit nicer for interactive development.

(Also would mean not having to reach into internals of the `JIRA` instance potentially a long time after `Issue` construction.)

Could use an intersection of `_fields.keys()` with `dir(self.fields)` to get the list of attributes to iterate through and set.

jira/resilientsession.py
```diff
@@ -72,10 +72,13 @@ class ResilientSession(Session):
     At this moment it supports: 502, 503, 504
     """
 
-    def __init__(self):
+    def __init__(self, jira_server=None):
```
fundthmcalculus: Excellent - I will make this change.

jira/resources.py
fundthmcalculus: Excellent! I will do this too.

fundthmcalculus:
Oops, didn't think about the issue creation - will look into it.

jira/resources.py
fundthmcalculus: One other thing to think about is a field named "Release Note". How to handle the use of spaces and other punctuation? Just replace with dashes and/or underscores? Or just remove spaces?

asqui: Hmm, very good point. Maybe there's some standard code that we can use to transform arbitrary text into a valid_python_identifier? (Because there's not just spaces but also punctuation and other cruft that could potentially be in custom field names.)

fundthmcalculus: How about this: http://stackoverflow.com/questions/5843518/remove-all-special-characters-punctuation-and-spaces-from-string

Basically remove everything that isn't alphanumeric. I could add some code to be intelligent about avoiding duplicate names (i.e. appending "_X" on the end if necessary).

asqui: Not sure duplicates are too much of a concern (this would only arise if two custom fields differ in name only by punctuation or spacing, which seems pretty edge-case).

It would be good to turn the custom field names into nice python identifiers though. How about this:
* Replace spaces with underscores
* Strip invalid characters
* Lowercase the rest

"Requestor's Manager" becomes "requesters_manager".

I figured this was a solved problem, but after some googling didn't turn up anything that quite fits this, but it shouldn't be too much trouble to hand-roll something appropriate.

fundthmcalculus: Can do. Rolling my own for that is trivial.

jira/exceptions.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, status_code=None, text=None, url=None, request=None, response
         self.response = response
         self.headers = kwargs.get('headers', None)
         self.log_to_tempfile = False
+        self.travis = False
```
asqui: Not sure what this is doing here -- seems unrelated to the rest of this commit?

jira/client.py
```diff
@@ -320,6 +320,9 @@ def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=N
                 for name in f['clauseNames']:
                     self._fields[name] = f['id']
 
+        self.field_map = {}
```
asqui: Minor: How about `self.field_map = self.generate_field_map()` for better readability?

jira/client.py
```diff
@@ -3157,3 +3188,13 @@ def __init__(self, options=None, basic_auth=None, oauth=None, async=None):
             "GreenHopper() class is deprecated, just use JIRA() instead.", DeprecationWarning)
         JIRA.__init__(
             self, options=options, basic_auth=basic_auth, oauth=oauth, async=async)
+
+
+def python_identifier(fieldstring):
+    return re.sub('[^0-9a-zA-Z_]', '', fieldstring.replace(' ', '_')).lower()
+
+def list_lookup(lst, key):
```
asqui: Why not use dictionaries rather than lists and avoid having to do this iterative lookup?

jira/client.py
```diff
@@ -3157,3 +3188,13 @@ def __init__(self, options=None, basic_auth=None, oauth=None, async=None):
             "GreenHopper() class is deprecated, just use JIRA() instead.", DeprecationWarning)
         JIRA.__init__(
             self, options=options, basic_auth=basic_auth, oauth=oauth, async=async)
+
+
+def python_identifier(fieldstring):
+    return re.sub('[^0-9a-zA-Z_]', '', fieldstring.replace(' ', '_')).lower()
```
asqui: Nice!

jira/client.py
```diff
@@ -3149,6 +3152,34 @@ def move_to_backlog(self, issue_keys):
...
+        custom_to_python = [(v,k) for k,v in python_to_custom]
+
+        try:
+            if customfieldname is not None:
+                return list_lookup(custom_to_python, customfieldname)
+            elif pythonfieldname is not None:
+                return list_lookup(python_to_custom, pythonfieldname)
+            else:
+                raise KeyError("Need to provide a custom field name, or a python field name.")
+        except:
```
asqui: Naked except blocks are rarely a good idea... this is going to catch and hide any unexpected exception that may arise within the `try` block... (Including the one you're raising on the line above, which is probably not what you intended?)  👎

jira/exceptions.py
fundthmcalculus: This is a separate bug-fix due to the `.travis` field being initialized outside the constructor (violation PEP 8).

jira/client.py
fundthmcalculus: Thanks! :)

jira/client.py
fundthmcalculus: Good question. Here's why:
customfield_1234 = cf[1234] = "Release Note"

This is a multi-to-one mapping, which a dictionary can't handle properly (no duplicate keys). So a set of lists with equivalence sub-tuples seemed the best way to handle this. Any better solution ideas?

jira/client.py
fundthmcalculus: Can do, just seemed dissimilar to the way the other fields are generated in the rest of the constructor.

jira/client.py
fundthmcalculus: I raise exception on line 3165 for forcing a jump into the except block to return the default key (the  key to look up), since anything that is not a custom field (and thus in the mapping lists) should remain untouched, since those are normal fields. Thoughts on how to do this more cleanly?

ssbarnea:
Closing it due to the lack of updates on it. Feel free to reopen it when you update it.

jira/client.py
fundthmcalculus: Adjusted to only catch the `KeyError`, which is the error raised when the mapping table lookup fails.

PR #271, opened 2016-10-19
Use context processor for opening files

radowit:
Without it, jira is causing "too many open files" error when there JIRA instance doesn't work/


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/271" target="_blank">ReviewNinja

tomislater:
LGTM

Issue #269, opened 2016-10-13
Search_issues is creating a bad url. 


cfortuner:
Trying to use the search_issues function is returning:

JiraError HTTP 404 url: https://jira.riotgames.com/rest/api/2/search?maxResults=False&validateQuery=True&jql=project+%3D+SCM&startAt=0 details: /var/folders/_m/wnb0sxqs67l6b3gn2pylxwlj_lknc3/T/jiraerror-b9ybpoc3.tmp

The search_params: validateQuery and maxResults dont seem to be working correctly.

Fr3dBear:
This is probably related to #229 a source fix is at the bottom. You should try using the modified client.py file. If this does not work, I have been able to circumvent the error in my own tests by defining maxResults to anything but False, such as 500. [https://jira.riotgames.com/rest/api/2/search?maxResults=500&validateQuery=True&jql=project+%3D+SCM&startAt=0](https://jira.riotgames.com/rest/api/2/search?maxResults=500&validateQuery=True&jql=project+%3D+SCM&startAt=0)

If all this fails go back to jira version 1.0.3, which suits most needs with the exception of the Outward/Inward issueLinks functionality.

hdost:
Closing for same reason as #229

PR #275, opened 2016-11-01
Fix fetching all issues when maxResults is Falsy

georgyo:
There are two problems that are currently affecting the ability to
fetch all results.

1. In older versions of this code when maxResults was falsy, it would
then change it to be 50, and then iterate over the code. This was
changed to a much more ideal method, however the old code blindly put
the value of maxResults into the dict of search params. It didn't matter
as it get reassigned pretty fast. The new code still put's it blindly
 into the map, but then never reassigns it, and continues to do a search
with it to determin how many results there are. Which creates an invalid
search query. Solution: Remove it from being added to the map. It
actually never intentionally used in that map without having it
reassigned first.

2. There is this isLast code, which likely could be removed entirely.
It, according to the comment only works with the Jira agile module. If
jira doesn't know what isLast is, the code would return that is_last is
True. Obviously this is a bad failure case, as if we don't have the
agile module we can only ever return one page. Solution: Change the
default to False.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -390,7 +390,7 @@ class JIRA(object):
             if isinstance(resource, dict):
                 total = resource.get('total', 1)
                 # 'isLast' is the optional key added to responses in JIRA Agile 6.7.6. So far not used in basic JIRA API.
-                is_last = resource.get('isLast', True)
+                is_last = resource.get('isLast', False)
                 start_at_from_response = resource.get('startAt', 0)
                 max_results_from_response = resource.get('maxResults', 1)
             else:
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/275" target="_blank">ReviewNinja

georgyo:
This fixes #229 and #253 by actually fixing the problems and not breaking compatibility with older releases. The other solutions to this problem didn't even seem to look at the code.

georgyo:
Also fixes #269

atulatri:
+1

ssbarnea:
Closing because this seems to make maxResults parameter not functional. Still, is this in not yet solved I would recommend writing a small unittest for testing this bug.

PR #276, opened 2016-11-03
Probably want this or in long running processes it can run out of fil…

UberMeatShield:
…e handles?

We have a long running process that can use this in a web service, when you have enough errors the process can actually run out of file handles and tank your web server.    This is mostly alleviated as an issue since file logging is no longer the default, but since the file log still exists in the released pypi version it gets a little awkward.

### DIFF
--- a/jira/exceptions.py
+++ b/jira/exceptions.py
@@ -51,6 +51,7 @@ class JIRAError(Exception):
             f = open(file_name, "w")
             t += " details: %s" % file_name
             f.write(details)
+            f.close()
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/276" target="_blank">ReviewNinja

asqui:
I think this is already covered by pull request #271...

lechat:
Yes, it is. I just merged #271

PR #277, opened 2016-11-10
Make subprocess error compatible with python < 3.3

kevinfrommelt:
`subprocess.SubprocessError` was added in python 3.3 as is not compatible with earlier versions.


review-ninja:
<a href="https://app.review.ninja/pycontribs/jira/pull/277" target="_blank">ReviewNinja

ssbarnea:
I guess this is not needed anymore.

PR #279, opened 2016-11-12
[wip] Job id prefix (jid) is now using travis numbers, shoudl work for PRs …

ssbarnea:
…too.

Change-Id: I17bb2e4ab0651edb6aad7b1cad0c9d65c7ebfb63
Signed-off-by: Sorin Sbarnea <ssbarnea@redhat.com>


jira/client.py
```diff
@@ -17,42 +17,46 @@
...
-from six import iteritems
 from requests.utils import get_netrc_auth
+from six import iteritems
+from six.moves.urllib.parse import urlparse
 # JIRA specific resources
-from .resources import Resource, Issue, Comment, Project, Attachment, Component, Dashboard, Filter, Votes, Watchers, \
-    Worklog, IssueLink, IssueLinkType, IssueType, Priority, Version, Role, Resolution, SecurityLevel, Status, User, \
-    CustomFieldOption, RemoteLink
+from jira.resources import *  # NOQA
+import xml.etree.ElementTree as etree
```
dbaxa: Please use [`defusedxml`](https://pypi.python.org/pypi/defusedxml) to access `ElementTree` instead of directly using it.

tests/tests.py
```diff
@@ -215,19 +215,42 @@ def __init__(self):
                 # [7-8] python version A=0, B=1,..
                 # [9] A,B -- we may need more than one project
 
-                prefix = 'Z' + (re.sub("[^A-Z]", "",
-                                       getpass.getuser().upper()))[0:6] + \
-                         str(sys.version_info[0]) + \
-                         str(sys.version_info[1])
+                """ `jid` is important for avoiding concurency problems when
```
asqui: Just trying to grok these changes. Looks like the comments above no longer reflect what the code is doing, which is not helpful :-(

tests/tests.py
```diff
@@ -215,19 +215,42 @@ def __init__(self):
...
+
+                jid lenght must be less than 9 characters because we may append
+                another one and the JIRA Project key length limit is 10.
+
+                Tests run in parallle:
+                * git branches master or developer, git pr or developers running
+                  tests outside Travis
+                * Travis is using "Travis" username
+
+                https://docs.travis-ci.com/user/environment-variables/
```
asqui: I'm confused. That page says not to depend on the USER being travis...

tests/tests.py
```diff
@@ -215,19 +215,42 @@ def __init__(self):
...
+                user = re.sub("[^A-Z_]", "", getpass.getuser().upper())
 
-                self.project_a = prefix + 'A'  # old XSS
+                def hashify(some_string, max_len=8):
+                    return hashlib.md5(some_string.encode('utf-8')).hexdigest()[:8].upper()
+
+                if user == 'TRAVIS' and 'TRAVIS_JOB_NUMBER' in os.environ:
+                    # please note that user underline (_) is not suppored by
+                    # jira even if is documented as supported.
+                    self.jid = 'T' + hashify(user + os.environ['TRAVIS_JOB_NUMBER'])
```
asqui: I guess we don't need to bring travis user into it if we're namespacing the jid with T for travis and Z for others?

Also, TRAVIS_JOB_NUMBER is going to be something like 531.3 - why do we even need to hash this to 8 digits? Can't we just replace the '.' and have a much more human-readable project key?

tests/tests.py
```diff
@@ -215,19 +215,42 @@ def __init__(self):
...
+                else:
+                    identifier = user + \
+                        chr(ord('A') + sys.version_info[0]) + \
+                        chr(ord('A') + sys.version_info[1])
+                    self.jid = 'Z' + hashify(identifier)
```
asqui: Similarly, do we really need to hash this? With the old ZUSERNAME35 the only way we'll clash is if two users that start with the same 8 letters are running tests simultaneously (on the same version of python). Do we have any evidence of this being a problem?

And even if it was, it would never affect Travis because of the Z prefix, which is the important part...

PR #286, opened 2016-11-16
Small typo fix in incompleted_issues method

sevenfourk:


asqui:
Should that entire method be removed? The deprecation warning suggests it would go away in February, 9 months ago!

sevenfourk:
@asqui you're right. I'm seeing these occurrences of warnings in 4 places.

```
$ grep -Ir 'on the 1st February 2016' *
jira/client.py:        warnings.warn('JIRA.completed_issues uses deprecated API, that is going to be removed on the 1st February 2016',
jira/client.py:                      ' on the 1st February 2016',
jira/client.py:                      ' on the 1st February 2016',
jira/resources.py:    """ Old, private API. Deprecated and will be removed from JIRA on the 1st February 2016. """
```

I would remove three methods in jira/client.py, where the warnings occur. And keep jira/resources.py unchanged for now. As it might lead to more changes, which will need to be reviewed more carefully.

Let me know, please, if you're okay with proposed changes, and I'll proceed (create a new PR for that change).  Thanks.

PR #287, opened 2016-11-16
Addressed flaky Travis parallel test execution by adding a retry

### LINKED ISSUE #143: CI failing
What is the issue with all the CI jobs failing - is it config/auth at the JIRA server end, or ..?
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28243638-ci-failing?utm_campaign=plugin&utm_content=tracker%2F3650997&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3650997&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ssbarnea:
Fixes #143

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -19,6 +19,8 @@ import py
 import pytest
 import requests
 from six import integer_types
+from tenacity import retry
+from tenacity import stop_after_attempt
 
 if platform.python_version() < '3':
     try:
@@ -80,6 +82,26 @@ def rndpassword():
     return ''.join(random.sample(s, len(s)))
 
 
+def hashify(some_string, max_len=8):
+    return hashlib.md5(some_string.encode('utf-8')).hexdigest()[:8].upper()
+
+
+def get_unique_project_name():
+    jid = ""
+    user = re.sub("[^A-Z_]", "", getpass.getuser().upper())
+
+    if user == 'TRAVIS' and 'TRAVIS_JOB_NUMBER' in os.environ:
+        # please note that user underline (_) is not suppored by
+        # jira even if is documented as supported.
+        jid = 'T' + hashify(user + os.environ['TRAVIS_JOB_NUMBER'])
+    else:
+        identifier = user + \
+            chr(ord('A') + sys.version_info[0]) + \
+            chr(ord('A') + sys.version_info[1])
+        jid = 'Z' + hashify(identifier)
+    return jid
+
+
 class Singleton(type):
 
     def __init__(cls, name, bases, dict):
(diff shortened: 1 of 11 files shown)


tests/tests.py
```diff
@@ -1692,9 +1694,9 @@ def test_create_version(self):
 
     def test_create_version_with_project_obj(self):
         project = self.jira.project(self.project_b)
-        version = self.jira.create_version('new version 1', project,
+        version = self.jira.create_version('new version 2', project,
```
asqui: Are we running tests in parallel?! (If so, where is this configured?)

Specifically, I know we run multiple Travis builds concurrently, but _within a single Travis build_ (or local test run, for that matter) do multiple tests run in parallel?

This change, and the setup of the `JiraTestManager` imply that we do, but I can't see any evidence of this in the configuration or Travis test run logs, so I'm a bit confused.

ssbarnea: @asqui Yes, Travis is running 3 test executions in parallel, one for each python version 27,34,37 but tests are run sequentially for each of them. That's a good thing, when it works and instead of going the dirty way of running them sequentially we should do the right thing and fix them. 

Also running tests in parallel will allow us to discover some real world scenarios that may not happen on a empty jira instance that is used exclusively for the purpose of one test. Is not quite a load test but if JIRA does not have atomic operations we should know it and adapt the library to make it more resilient to the server misbehaviours.

Based on the execution history I am inclined to believe that Jira is misbehaving and that is caching some results: example you may get a 5xx or 4xx error trying to create a resource (like a group or user) that was recently removed. By putting a sleep() or 1-2 seconds we may workaround this but I am not 100% sure yet as is very hard to get confirmation due to the flaky nature of the tests.

Still, I observed that these errors do not happen randomly over all the tests, when they happen they usually happen inside specific ones. 

@asqui It would be great if you can have a look as this is the most important issue with the library, one that is preventing us from starting to release often and to provide bug fixes. Feel free to use the `feature/travis` branch to test other changes.

If you want to simulate what happens with Travis but on your own machine you could try to run `detox` which is mainly running tox in parallel.

The entire idea is that we should support running tests in parallel and that's why we generate the `TestManager.jid` unique id, which is supposed to be unique per execution: we should never have two test executions that run in parallel that are using the same `jid`.

Again, I am talking about parallelism regarding the whole execution of the test suites, not tests themselves, these are sequential and even the order is important.

vkrizan:
I would suggest using random keys/names for test projects and issues. That should ensure that the next or parallel test will not fail. WDYT?

PR #291, opened 2016-11-18
Removing deprecated Agile-related methods

sevenfourk:
Hello. As was discussed here: https://github.com/pycontribs/jira/pull/286#issuecomment-261102036.

Created a PR for removing deprecated Agile-related methods.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2757,52 +2757,6 @@ class JIRA(object):
 
         return json_loads(r)
 
-    def completed_issues(self, board_id, sprint_id):
-        """Return the completed issues for ``board_id`` and ``sprint_id``.
-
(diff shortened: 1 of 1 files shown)


asqui:
@ssbarnea: What's the process for versioning and documenting breaking changes? What versioning scheme are we using for this project? (The fully automated version incrementing and release note generation doesn't offer much flexibility!)

ssbarnea:
We have not tests for these and they were already removed so I will just merge the change. The only API contract that we currently have are the tests, if we break them we will increase the version

If someone wants more stability for Python JIRA, they are welcome to make PR with new tests.

We already have 10+ API calls that are broken on the cloud by Atlassian and it took two weeks to get an engineer from them to have a look.

Issue #298, opened 2016-11-25
Fail on AttributeError when can not authenticate


lukas-bednar:
```
In [1]: from jira.client import JIRA

In [2]: j = JIRA(options=dict(server="https://issues.jboss.org", verify=True), basic_auth=("invalid_user", "invalid_password"), validate=True, max_retries=1)
WARNING:root:Bug https://jira.atlassian.com/browse/JRA-59676 trying again...
WARNING:root:Bug https://jira.atlassian.com/browse/JRA-59676 trying again...
WARNING:root:Bug https://jira.atlassian.com/browse/JRA-59676 trying again...
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-27f57a346db5> in <module>()
----> 1 j = JIRA(options=dict(server="https://issues.jboss.org", verify=True), basic_auth=("invalid_user", "invalid_password"), validate=True, max_retries=1)

/home/lbednar/work/jira/jira/client.pyc in __init__(self, server, options, basic_auth, oauth, jwt, kerberos, validate, get_server_info, async, logging, max_retries, proxies)
    305                 self._version = tuple(si['versionNumbers'])
    306             except Exception as e:
--> 307                 logging.error("invalid server_info: %s", si)
    308                 raise e
    309         else:

AttributeError: 'bool' object has no attribute 'error'
```

This issue was introduced by following commit:

https://github.com/pycontribs/jira/commit/97b36dfd1665e1693425aa264b08724afa4a9076#diff-12827fc5aab7f1c957381e453bdfe85aR301

The problem is that logging module is hidden by parameter passed to ``__init__``.

eanxgeek:
I hit this bug yesterday, it seemed to just appear out of nowhere.  My program had been working and since I wrote it, over a week ago, but then yesterday this error started to surface.

I have python version 3.6.4 and the latest version of the Jira library:
$ python --version
Python 3.6.4 :: Anaconda custom (64-bit)
jira (1.0.10)                         - Python library for interacting with JIRA via REST
                                        APIs.
  INSTALLED: 1.0.10 (latest)

When I went into client.py I noticed this fix had not been applied, apparently a failed merge somewhere?  I applied the change and my program began working again.

adehad:
fixed in `pip install jira==3.1.0rc1`

PR #299, opened 2016-11-25
Fix AttributeError issue #298

lukas-bednar:


asqui:
Actually I'd question what the value of the logging argument is, and whether we should remove that instead of working around it.

Looks like this argument was added to `__init__` in commit ef82109 which is described as "v0.31 manual merge with github fork (used to revive unittests)" and it, to this day, is only used to guard one single log line within `create_project()` so this smells like some debugging code that was accidentally left in and should be removed.

@ssbarnea : Any objection if we solve this issue by removing the `logging` argument to `JIRA.__init__()`

lukas-bednar:
> Actually I'd question what the value of the logging argument is, and whether we should remove that instead of working around it.

It is boolean

asqui:
By "what's the value" I meant "what is the point of having it" not what type of object is assigned to the variable :-)

vkrizan:
I've created a separate pull request #308 to fix logger issues (that I've observed), and to deprecate `JIRA.logging` boolean.

ssbarnea:
I am closing this in favour of PR#308

Issue #300, opened 2016-11-25
JIRA(validate=True) doesn't raise exception when user is not authenticated


lukas-bednar:
This part of code promises that I should get exception when user is not authenticated, but it passes this section without problem ...
https://github.com/pycontribs/jira/blob/develop/jira/client.py#L296
```
            # This will raise an Exception if you are not allowed to login.
            # It's better to fail faster than later.
            self.session()
```
it fails further on issue #298

hdost:
This is fixed as of 1.0.10

PR #301, opened 2016-11-25
Fix JIRA(validate=True) option, issue #300

lukas-bednar:


jira/client.py
```diff
@@ -296,7 +296,12 @@ def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=N
         if validate:
             # This will raise an Exception if you are not allowed to login.
             # It's better to fail faster than later.
-            self.session()
+            user = self.session()
+            if user.raw is None:
+                auth_method = (
+                    oauth or basic_auth or jwt or kerberos or "anononymous"
```
ssbarnea: `anononymous` sounds like a typo to me :)

lukas-bednar: oh, yes you are right, I will fix it, thanks.

PR #302, opened 2016-12-05
chg: Added support for dev_status endpoint

andrew-epstein:
Adding a way to access JIRA's internal `dev-status` endpoint. This endpoint is used to access information concerning commits, pull requests, repositories, and branches that relate to a given issue.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -651,6 +651,32 @@ class JIRA(object):
         """
         return self._find_for_resource(Dashboard, id)
 
+    # Dev-status
+
+    def dev_status(self, issue, dataType='repository'):
+        """Access JIRA's internal "dev-status" endpoint
+
+        :param issue: either an actual issue, or an issue id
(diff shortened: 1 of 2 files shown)


ssbarnea:
Can you also add an unittest for it? Without it is hard to know if the code would work or not.

andrew-epstein:
@ssbarnea: I would be happy to write a unit test. However, looking at the existing testing code, I'm not sure how to go about that. It looks like there is a lot of code to set up objects that get used in testing later, and I don't know how to modify that to include the dev-status info. Can you provide any help? FYI, [here](https://answers.atlassian.com/questions/32535478/get-commits-info-of-a-jira-issue-using-rest-api) is a link with a little info about this endpoint. There's not much info out there since it's internal-only.

andrew-epstein:
This is no longer something that I need, so I'm not going to write unit tests for it. I'll close the pull request.

PR #304, opened 2016-12-06
Add support for delete_attachment.

### LINKED ISSUE #282: Deleting Attachments
Is there a way to delete attachments from issues?

k0ste:
Also include docs.
Resolve #282.

Test on 7.2.5, works fine.


asqui:
Nice work with updating the docs!

Would be great if you could also include a unit test to cover this functionality?

PR #308, opened 2016-12-16
Use own logger and remove JIRA.logging

### LINKED ISSUE #298: Fail on AttributeError when can not authenticate

vkrizan:
jira package used direct `logging` functions (warning, error, debug, ...),  which uses root logger for logging. If an application, that uses jira package, does not register any handlers to root logger, then a logging function automatically creates and adds a `StreamHandler` to it. That can have an unexpected results for application developer. Jira package already registered its own logger (named "jira") and attached a NullHandler to it, but it was never used.

Changelog:
* logging to "jira" logger (fixes #298)
* removed `Jira.logging` boolean
* refactored usage of logging msg formatting trough arguments

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -460,8 +460,8 @@ class JIRA(object):
         :param size: number of threads to run on.
         """
         if hasattr(self._session, '_async_jobs'):
-            logger.info("Executing async %s jobs found in queue by using %s threads..." % (
-                len(self._session._async_jobs), size))
+            logger.info("Executing async %s jobs found in queue by using %s threads...",
+                        len(self._session._async_jobs), size)
             threaded_requests.map(self._session._async_jobs, size=size)
 
             # Application properties
@@ -540,8 +540,8 @@ class JIRA(object):
         if isinstance(attachment, string_types):
             attachment = open(attachment, "rb")
         if hasattr(attachment, 'read') and hasattr(attachment, 'mode') and attachment.mode != 'rb':
-            logger.warning(
-                "%s was not opened in 'rb' mode, attaching file may fail." % attachment.name)
+            logger.warning("%s was not opened in 'rb' mode, attaching file may fail.",
+                           attachment.name)
 
         url = self._get_url('issue/' + str(issue) + '/attachments')
 
@@ -1194,7 +1194,7 @@ class JIRA(object):
         try:
             r_json = json_loads(r)
         except ValueError as e:
-            logger.error("%s\n%s" % (e, r.text))
+            logger.error("%s\n%s", e, r.text)
             raise e
         return r_json
 
@@ -2107,7 +2107,7 @@ class JIRA(object):
         try:
             r_json = json_loads(r)
         except ValueError as e:
-            logger.error("%s\n%s" % (e, r.text))
+            logger.error("%s\n%s", e, r.text)
             raise e
         return r_json
 
@@ -2148,7 +2148,7 @@ class JIRA(object):
                 return mimetypes.guess_type("f." + imghdr.what(0, buff))[0]
             except (IOError, TypeError):
                 logger.warning("Couldn't detect content type of avatar image"
-                                ". Specify the 'contentType' parameter explicitly.")
+                               ". Specify the 'contentType' parameter explicitly.")
                 return None
 
     def email_user(self, user, body, title="JIRA Notification"):
@@ -2190,7 +2190,7 @@ class JIRA(object):
                 'username': old_user}
 
             # raw displayName
-            logger.debug("renaming %s" % self.user(old_user).emailAddress)
+            logger.debug("renaming %s", self.user(old_user).emailAddress)
 
             r = self._session.put(url, params=params,
                                   data=json.dumps(payload))
@@ -2213,7 +2213,7 @@ class JIRA(object):
                 "RunCanned": "Run"}
 
             # raw displayName
-            logger.debug("renaming %s" % self.user(old_user).emailAddress)
+            logger.debug("renaming %s", self.user(old_user).emailAddress)
 
             r = self._session.post(
                 url, headers=self._options['headers'], data=payload)
@@ -2227,7 +2227,8 @@ class JIRA(object):
 
             if re.compile("XSRF Security Token Missing").search(r.content):
                 logger.fatal(
-                    "Reconfigure JIRA and disable XSRF in order to be able call this. See https://developer.atlassian.com/display/JIRADEV/Form+Token+Handling")
+                    "Reconfigure JIRA and disable XSRF in order to be able call this. "
+                    "See https://developer.atlassian.com/display/JIRADEV/Form+Token+Handling")
                 return False
 
             with open("/tmp/jira_rename_user_%s_to%s.html" % (old_user, new_user), "w") as f:
@@ -2255,7 +2256,8 @@ class JIRA(object):
                 logger.error("User %s does not exists. %s", old_user, e)
                 return msg
 
-            logger.error(msg + '\n' + "User %s does still exists after rename, that's clearly a problem." % old_user)
+            logger.error(msg + "\nUser %s does still exists after rename, that's clearly a problem.",
+                         old_user)
             return False
 
     def delete_user(self, username):
@@ -2293,7 +2295,7 @@ class JIRA(object):
                 return True
             else:
                 logger.warning(
-                    'Got response from deactivating %s: %s' % (username, r.status_code))
+                    'Got response from deactivating %s: %s', username, r.status_code)
                 return r.status_code
         except Exception as e:
             print("Error Deactivating %s: %s" % (username, e))
@@ -2350,8 +2352,7 @@ class JIRA(object):
             if r.status_code == 200:
                 return True
             else:
-                logger.warning(
-                    'Got %s response from calling backup.' % r.status_code)
+                logger.warning('Got %s response from calling backup.', r.status_code)
                 return r.status_code
         except Exception as e:
             logger.error("I see %s", e)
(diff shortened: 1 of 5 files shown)


vkrizan:
Tests in Travis seems to be failing because of testing jira server.

asqui:
Build passed on re-run.

Please remove the logging argument from JIRA() constructor and I'll merge this change.

vkrizan:
Removed.

However, tests are failing again. It looks like a previous test created a project but did not deleted it, then the next test is failing to create an already existing project.
```
JIRAError: JiraError HTTP 400 url: https://pycontribs.atlassian.net/rest/project-templates/latest/templates
A project with that name already exists., Project 'Test user=travis key=TTRAVIS27T A' uses this project key.
```
Could be fixed by checking for project existence or with `try-except`.

ssbarnea:
Until we solve the test execution and merge the fixes (travis branch), I will not merge any other change to the code base.  See https://github.com/pycontribs/jira/pull/287 -- feel free to play with it and to add fixes there.

It is insane to try to make any changes when the tests are failing, we should focus on getting these tests stable first. I will try to do this over the weekend but help would be highly appreciated -- my time is quite limited.

The problem is with the test fixtures: because our tests are integration tests run against a cloud product it means that we do not start with a clean instance, it is likely to have leftovers from previous test executions. The trick is to improve the test fixtures to be able to cope with these left overs. Also we need to be sure that running them in parallel for each supported python version would not fail randomly tests.

I know that's much harder than classic unittests, but when we will have them working they will provide a much better QA than unittests because JIRA API is always changing and we do want to have a library that works with current versions of JIRA.

jira/client.py
```diff
@@ -92,7 +92,8 @@ def emit(self, record):
 #    "which means that there is a big change of having problems. " \
 #    "Possible workaround http://stackoverflow.com/a/17628350/99834" % encoding)
 
-logging.getLogger('jira').addHandler(NullHandler())
+logger = logging.getLogger('jira')
```
ssbarnea: I think that the pattern is to use `logging.getLogger(__name__)`  like in  https://github.com/kennethreitz/requests/blob/5aa66615870beeb7b0ed3af27457f689db262ab3/requests/packages/urllib3/contrib/appengine.py#L68

jira/client.py
```diff
@@ -2376,14 +2377,14 @@ def reindex(self, force=False, background=True):
 
         r = self._session.get(url, headers=self._options['headers'])
         if r.status_code == 503:
-            # logging.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
+            # log.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
```
ssbarnea: logger or log, be sure we use only one name.

jira/client.py
```diff
@@ -196,7 +197,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False,
-                 validate=False, get_server_info=True, async=False, logging=True, max_retries=3, proxies=None):
+                 validate=False, get_server_info=True, async=False, max_retries=3, proxies=None):
         """Construct a JIRA client instance.
```
ssbarnea: This will break the old code that using this parameter. We need to refactor this so the old cold will not break, probably via `**kwargs` but one thing that I don't like about `**kwargs` is that they make much harder to read the documentation / API.

jira/client.py
asqui: +1

jira/client.py
asqui: +1
I'd vote for `log` because it reads better as `log.warning()`

jira/client.py
asqui: -1
I don't think this flag actually does anything useful; it only protects one single logging call in `create_project()`, and it's not documented, so I really doubt anyone is using it. I really don't think it's worth preserving for the sake of maintaining backwards compatibility.

PS: @ssbarnea I [asked for your input](https://github.com/pycontribs/jira/pull/308#discussion_r92927997) on this back in December! Don't you monitor mentions?

ssbarnea: I am investigating now a possible migration to `kwargs` as I would prefer not to break the existing code that is using the logging parameter. If we would use kwargs we could avoid breaking the code, old code will continue to work.

I am quite busy and it is easy to miss a notification, also I don't use this library at work which makes is less likely to get attention.

asqui: -1 on introducing `**kwargs` for the reasons you mentioned above.
What is your goal?
Take `logging` without breaking existing calls, but ignore it?
Take `logging` and keep the current behaviour? (controls one log line in `create_project()`)
Take `logging` and change it to control all logging?

I don't understand the point of having the argument. If we're using the python `logging` framework correctly then calling code can customising what logging they do and don't get using features of the `logging` library...

ssbarnea: @asqui it seems that you missed to observe that `JIRA.__init__()` constructor accepts the logger parameter. If we are to remove it we need to be careful in order not to break code that is already using it. When I mean preventing breaking, I mean to keep it running instead of getting a compilation error because we removed the parameter.

jira/client.py
vkrizan: `'jira'` logger name was the original one (was there before this PR). Using `__name__` for a logger name is not correct here, because `__name__ = 'client'`, which does not describe the library logger properly. However, naming it after `__package__` would be a more correct way.

FYI: Werkzeug uses harcoded logger name: https://github.com/pallets/werkzeug/blob/423a356d4a49a50179a2be7ba131570dbc607a34/werkzeug/_internal.py#L80

jira/client.py
vkrizan: I can revert commit 3bd9e69 back, which kept the argument but deprecated it.

jira/client.py
vkrizan: If you guys prefer `log` over `logging` then I can change it.

jira/client.py
ssbarnea: @vkrizan pointed about `__package__` which seems like the winner even if almost none is using. Out of curiosity I added [this question on SO](http://stackoverflow.com/questions/41694787/how-should-i-name-the-python-logger-inside-a-python-library).

vkrizan: > Out of curiosity I added this question on SO.

I kinda have answer to that: Loggers are composed in hierarchy, separated by dot syntax. `None` or empty string is a root logger. `'jira'` is a logger which has a root logger as parent. `'jira.utils'` would be another logger (namespace) which will have "jira" logger as a parent. By default all sub loggers propagate their messages to their parents, unless configured not to do so.

`__package__` is a good candidate, but it will create another logger namespace when used in sub packages. (Might not be always desirable.)

These two documents should have more info:  
https://docs.python.org/3/howto/logging-cookbook.html
https://docs.python.org/3/howto/logging.html

jira/client.py
ssbarnea: @vkrizan Ok, rename it to `log` and keep the constructor signature but document that the parameter is deprecated and ignored. Once we have this we will merge this PR and move to fix others. Thanks.

jira/client.py
asqui: `__name__` is `'jira.client'` here, not `'client'`.
If we're expecting people to reference the name of this logger then we need to be conscious that this logger name becomes part of our public API and a change to its name is a breaking change that will impact users of the library (silently), so perhaps having it explicitly hard-coded is not so bad?

vkrizan: Interesting. You are right with the `__name__` (I've confused it with func/class name).

> we need to be conscious that this logger name becomes part of our public API and a change to its name is a breaking change

If mapped to `__name__`, then unless top package name changes, it will be stable. `jira.client` is fine for a logger, messages are propagated. However, I guess that its not needed to divide logger namespaces for each module in the package.

Here is a test of propagation:
```
import logging

parent = logging.getLogger('jira')
child = logging.getLogger('jira.client')
parent.addHandler(logging.StreamHandler())
assert child.propagate
assert child.parent is parent
child.error('Test')
```

ssbarnea: In this case I vote for using `__name__`, because
* most popular option among serious packages
* backwards compatible with "jira" (if someone it change jira logger, the change would also apply to jira.client because is a child one).
* allows separation of logging if someone is renaming the main package or if it embeds jira package in their own module.

The logging cookbook could use a pull request, just to clarify the recommended way of using logging in libraries. I bet we are not the only ones confused by it.

ssbarnea:
Probably you observed that I merged this into a new branch. This is in order to fix the current problems with the logging before going into main line.

vkrizan:
Thank you for letting me know. Feel free to change logger name or the variable.

PR #309, opened 2016-12-18
Support for bulk issue creation

yoyoyopcp:
Since Atlassian has added a bulk create to their REST API, this commit
enables the use of making multiple issues in one batch. It will log
errors for each unsuccessful issue and return the list of successfully
created issues.

Testing: py.test tests/tests.py

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -151,11 +151,16 @@ You can even bulk create multiple issues::
         'description': 'Final issue of batch.',
         'issuetype': {'name': 'Bug'},
     }]
-    issues = jira.create_issues(field_list=issue_dict)
+    issues = jira.create_issues(field_list=issue_list)
 
 .. note::
     Project, summary, description and issue type are always required when creating issues. Your JIRA may require
     additional fields for creating issues; see the ``jira.createmeta`` method for getting access to that information.
+
+.. note::
+    Using bulk create will not throw an exception for a failed issue creation. It will return a list of dicts that
+    each contain a possible error signature if that issue had invalid fields. Successfully created issues will contain
+    the issue object as a value of the ``issue`` key.
 
 You can also update an issue's fields with keyword arguments::
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -130,6 +130,29 @@ Or you can use a dict::
     }
     new_issue = jira.create_issue(fields=issue_dict)
 
+You can even bulk create multiple issues::
+
+    issue_list = [
+    {
+        'project': {'id': 123},
+        'summary': 'First issue of many',
+        'description': 'Look into this one',
+        'issuetype': {'name': 'Bug'},
+    },
+    {
+        'project': {'key': 'FOO'},
+        'summary': 'Second issue',
+        'description': 'Another one',
+        'issuetype': {'name': 'Bug'},
+    },
+    {
+        'project': {'name': 'Bar'},
+        'summary': 'Last issue',
+        'description': 'Final issue of batch.',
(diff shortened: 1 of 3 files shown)


docs/examples.rst
```diff
@@ -130,6 +130,29 @@ Or you can use a dict::
...
+        'description': 'Another one',
+        'issuetype': {'name': 'Bug'},
+    },
+    {
+        'project': {'name': 'Bar'},
+        'summary': 'Last issue',
+        'description': 'Final issue of batch.',
+        'issuetype': {'name': 'Bug'},
+    }]
+    issues = jira.create_issues(field_list=issue_dict)
```
asqui: This doesn't make sense. You've constructed `issues_list` but are not using it in this call.

Also, is it worth mentioning that the return value contains only the ones that were successfully created? (How do you programmatically find out which ones failed, e.g. if you want to retry them? Or mark them as failed somewhere upstream?)

jira/client.py
```diff
@@ -888,6 +893,46 @@ def create_issue(self, fields=None, prefetch=True, **fieldargs):
...
             return Issue(self._options, self._session, raw=raw_issue_json)
 
+    def create_issues(self, field_list, prefetch=True):
+        """Bulk create new issues and return an issue Resource for each successfully created issue.
+
+        See `create_issue` documentation for field information.
+
+        :param field_list: a list of dicts each containing field names and the values to use. Each dict
+            is an individual issue to create and is subject to its minimum requirements.
+        """
```
asqui: Missing documentation for `prefetch` argument.

jira/client.py
```diff
@@ -121,6 +121,18 @@ def _get_template_list(data):
...
 
+def _field_worker(fields=None, **fieldargs):
+    data = {}
+    if fields is not None:
+        data['fields'] = fields
+    else:
+        fields_dict = {}
+        for field in fieldargs:
+            fields_dict[field] = fieldargs[field]
+        data['fields'] = fields_dict
```
asqui: Isn't this just equivalent to `data['fields'] = fieldargs.clone()`? Which itself is probably equivalent to `data['fields'] = fieldargs` (unless the fields dict is being mutated, which I'd consider very unexpected.)

What am I missing...?

jira/client.py
```diff
@@ -888,6 +893,46 @@ def create_issue(self, fields=None, prefetch=True, **fieldargs):
...
+        for error in raw_issue_json['errors']:
+            failed_fields = field_list[error['failedElementNumber']]
+            logging.error('Error creating issue with fields: %s\n%s', failed_fields,
+                          error['elementErrors']['errors'])
+        for result in raw_issue_json['issues']:
+            if prefetch:
+                issue_list.append(self.issue(result['key']))
+            else:
+                issue_list.append(Issue(self._options, self._session, raw=result))
+        return issue_list
```
asqui: This design is problematic, because the caller has no easy way to correlate the inputs to the outputs.

i.e. if I input a list of 3 issues and get back 2 results, how do I know which two inputs correspond to those two results?

docs/examples.rst
yoyoyopcp: Good catch on the documentation error!

The return value has been redesigned. Please view the follow up commit to see the new return value that addresses this valid concern.

jira/client.py
yoyoyopcp: I extracted this code from the original create_issue method, but you are right that a simpler design was available. It has been refactored in the follow up commit.

jira/client.py
yoyoyopcp: Good catch, again!

PR #314, opened 2016-12-27
Allow to define internal comments in service desk

### LINKED ISSUE #313: Allow to create internal comments in Jira Service Desk
So far it is not possible to create an internal comment by using JIRA client. 

In order to achieve this, the body of request has to have next structure:
```json
{"body":"comment message", "properties":[{"key":"sd.public.comment","value":{"internal":True}},]}
```

PS. I will add a MR later

tarasmatsyk:
Closes #313 

I would be happy to get an understanding of your plans regarding Jira Service Desk Support.

Maybe current client has to be left as is and we can come up with a separate one?

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -959,7 +959,7 @@ class JIRA(object):
         return self._find_for_resource(Comment, (issue, comment))
 
     @translate_resource_args
-    def add_comment(self, issue, body, visibility=None):
+    def add_comment(self, issue, body, visibility=None, is_internal=False):
         """Add a comment from the current authenticated user on the specified issue and return a Resource for it.
 
         The issue identifier and comment body are required.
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -970,15 +970,20 @@ def add_comment(self, issue, body, visibility=None):
             "type" is 'role' (or 'group' if the JIRA server has configured
             comment visibility for groups) and 'value' is the name of the role
             (or group) to which viewing of this comment will be restricted.
+        :param is_internal: defines whether a comment has to be marked as 'Internal' in Jira Service Desk
         """
         data = {
-            'body': body}
+            'body': body,
+            'properties':[{'key':'sd.public.comment','value':{'internal':is_internal}},]
```
ssbarnea: This parameter should not be added if is not defined. If we are to implement this we need to keep the old behaviour the same: not adding the parameter. This is important because adding it may break the call on other versions of JIRA.

tarasmatsyk: @ssbarnea, understood, did not know it will break other versions of JIRA as I tested only two of them (ServiceDesk) + Base JIRA (I forgot how it is called).

I will close the PR.

Issue #316, opened 2017-01-05
input function is improperly imported in jirashell.py


xiaket:
If we have configparser pre-installed in a python2 environment, the input function is not imported from six, but as provided as a built-in function, which is equivalent to eval(raw_input(prompt)), and that eval would ruin the user input:

```
Have you authorized this program to connect on your behalf to https://xxxx.atlassian.net?(y/n)y
Traceback (most recent call last):
  File "/usr/local/bin/jirashell", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/jira/jirashell.py", line 248, in main
    options['server'], oauth['consumer_key'], oauth['key_cert'], oauth['print_tokens'], options['verify'])
  File "/usr/local/lib/python2.7/dist-packages/jira/jirashell.py", line 63, in oauth_dance
    'Have you authorized this program to connect on your behalf to {}?'.format(server))
  File "<string>", line 1, in <module>
NameError: name 'y' is not defined
```

We should put the import of input from six into a separate line.

PR #317, opened 2017-01-05
properly import input function from six. see #316

xiaket:
Signed-off-by: Xia Kai(夏恺) <xiaket@gmail.com>


xiaket:
I cannot really see why this commit is breaking the tests.   :(

asqui:
Build was failing due to issues with test stability. I've re-run the failed builds and they are green now.

PR #329, opened 2017-01-20
Removed logging parameter and switched back to standard module logging

ssbarnea:
Sem-Ver: deprecation

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -197,7 +197,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False,
-                 validate=False, get_server_info=True, async=False, max_retries=3, proxies=None):
+                 validate=False, get_server_info=True, async=False, max_retries=3, proxies=None, **kwargs):
         """Construct a JIRA client instance.
 
         Without any arguments, this client will connect anonymously to the JIRA instance
@@ -2377,7 +2377,7 @@ class JIRA(object):
 
         r = self._session.get(url, headers=self._options['headers'])
         if r.status_code == 503:
-            # log.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
+            # logger.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
             return 503
 
         if not r.text.find("To perform the re-index now, please go to the") and force is False:
--- a/jira/client.py
+++ b/jira/client.py
@@ -304,6 +304,7 @@ class JIRA(object):
             # It's better to fail faster than later.
             self.session()
 
+        self.deploymentType = None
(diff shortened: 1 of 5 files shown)


jira/client.py
```diff
@@ -92,7 +92,8 @@ def emit(self, record):
 #    "which means that there is a big change of having problems. " \
 #    "Possible workaround http://stackoverflow.com/a/17628350/99834" % encoding)
 
-logging.getLogger('jira').addHandler(NullHandler())
+logger = logging.getLogger('jira')
```
asqui: Suggest calling this variable `log` for more readable log statements throughout.

jira/client.py
```diff
@@ -196,7 +197,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False,
-                 validate=False, get_server_info=True, async=False, logging=True, max_retries=3, proxies=None):
+                 validate=False, get_server_info=True, async=False, max_retries=3, proxies=None, **kwargs):
```
asqui: Serious objection to this approach.

This allows callers to continue supplying a `logging` argument and have it silently ignored (which is itself a breaking change of the API). If you are arguing for no breaking change in this revision then you have to continue to support the logging argument and its current behaviour. (I have previously tried to argue that the current behaviour of this undocumented argument is of little value and likely not of use to anyone.)

Additional objection to adding `**kwargs` is that this will allow any other argument to be accepted and silently ignored by the constructor. This may include arguments that don't exist, or that are typos of arguments that do exist. e.g. `get_serverinfo=False`. This is terrible for usability of this API.

If you don't care about preserving the behaviour associated with the `logging` argument, but just want to make sure that no code breaks due to syntax errors, then leave logging as an explicit argument, and document it as unsupported in the docstring. (and hopefully remove it in the next major revision?)

jira/client.py
```diff
@@ -2376,14 +2377,14 @@ def reindex(self, force=False, background=True):
 
         r = self._session.get(url, headers=self._options['headers'])
         if r.status_code == 503:
-            # logging.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
+            # logger.warning("JIRA returned 503, this could mean that a full reindex is in progress.")
```
asqui: Delete commented out code rather than revising it?

jira/client.py
```diff
@@ -92,7 +92,8 @@ def emit(self, record):
 #    "which means that there is a big change of having problems. " \
 #    "Possible workaround http://stackoverflow.com/a/17628350/99834" % encoding)
 
-logging.getLogger('jira').addHandler(NullHandler())
+logger = logging.getLogger('jira')
+logger.addHandler(NullHandler())
```
asqui: Probably more correct for this handler to be added in the `__init__.py` of the `jira` module.

PR #340, opened 2017-02-10
Fixing missing jirashell entry point

### LINKED ISSUE #334: jirashell got removed from entry_points
I've recently upgraded the requirements of my application and noted that the `jirashell` command has been removed from the setup.py `entry_points['console_scripts']` list.

Version 1.0.7 still had it and with 1.08 it looks to be gone:
https://github.com/pycontribs/jira/blob/1.0.7/setup.py#L183

I understand that the script is still available in the installation in site-packages, but it was really handy to have it right there on the command line. I referred to it from the documentation so beginners can easily setup the JIRA OAuth.

Is there any particular reason for this? Can we bring it back?

Thanks

ssbarnea:
Fixes #334

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2097,6 +2097,13 @@ class UserAdministrationTests(unittest.TestCase):
         self.jira.delete_user(self.test_username)
 
 
+class JiraShellTests(unittest.TestCase):
+
+    def test_jirashell_command_exists(self):
+        result = os.system('jirashell --help')
+        self.assertEqual(result, 0)
+
+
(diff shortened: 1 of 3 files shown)


tests/tests.py
```diff
@@ -2097,6 +2097,13 @@ def test_remove_user_from_group(self):
         self.jira.delete_user(self.test_username)
 
 
+class JiraShellTests(unittest.TestCase):
```
asqui: Nice. I like the unit test to prevent regression :-)

PR #347, opened 2017-02-22
Commented out sleep that shouldn't be in Resource.update()

qeomash:
Updated to the last version today, and noticed that updates were taking significantly longer than before...because of this sleep. Can we remove it for the next release, please?

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -291,7 +291,7 @@ class Resource(object):
 
         # TODO(ssbarnea): compare loaded data in order to verify if resource was updated indeed
         # we had random test failures (probably) due to caching
-        time.sleep(4)
+        #time.sleep(4)
         self._load(self.self)
 
     def delete(self, params=None):


ssbarnea:
The delay was added there because JIRA api was unreliable and the load could have returned the old data due to caching.

Still, I do understand that a 4 seconds delay on each update may be a serious issue in some use-cases. 

If we just remove this delay we will randomly fail to pass the tests so the solution is to transform it in an optional feature, like `async` or `skip_reload`. 

As we probably want this option in more places it would worth to implement it per session (JIRA class) and use it from there instead of adding a new param to the update() method.

See how the timeout or retry mechanism is implemented, i think this one should follow the same pattern.

qeomash:
Is it only the tests? Have users encountered random failures in regular use? (I haven't.)

ssbarnea:
I did while running the unittesg with Travis so this is a real issue. If it happens on CI it may also happen when used in production. This is why I had to add it! 

-- sorin

sp-1234:
This delay was complete nonsense in the first place, it should have never been added, instead the underlying issue should have been solved. What a shame!
At least there should be an option to disable it, right now it's impossible, so I hope this PR gets merged. It's been 9 months already, what are you waiting for?

ssbarnea:
The sleep was already removed. BTW, commenting code is not recommended, better to remove.

Issue #281, opened 2016-11-12
User don't have KEY field which breaks integrity


President:
jira.group_members("jira-users") returns list of users where KEY field is actually NOT a user's key but a user's name!
But this field - user's name can be changed in JIRA by administrator and this causes big problem if we want to extract the data and keep the references between the entities.

Example:
1.you exported users to your database
2.you exported issues and linked them to the users (you should use issue.fields.assignee.name, as you cannot use fields.assignee.key because you don't know the user's keys)
3.you changed user's name
4.you exported new users (new user is created which is actually a duplicate)
5. you trying to sync issues and link them to users and new issues are linked to new user.
Now you have duplicate users both with issues!

Suggestion:
jira.group_members() should use user's key for keys and user's name field should be among other user's fields together with email and fullname.

JIRA's API already supports this - 
"/rest/api/2/group?groupname=#{groupname}&expand=users"  returns JSON with users' keys.

Please fix this.

President:
this requires actually a very simple change:
_def group_members()_ should be updated
...
result[user['**key**']] = {**'name': user['name'],** 'fullname': user['displayName'], 'email': user.get('emailAddress', 'hidden'), 'active': user['active']}
...

asqui:
Seems like a sensible suggestion. Please submit a pull request for this and it will be incorporated into the library.

@ssbarnea: This would be a breaking change. Any code that is already using `group_members()` will begin to receive different results. What is our approach for handling breaking changes to the library?

ssbarnea:
Usually we should not break the API. Still, if it solves a broken API and it adds tests, I would not mind merging the change into a newer minor or major version (but not in a patch one).

Just make a PR and I will take care to be merged at the right moment.

NfNitLoop:
I think we can make a change here that doesn't break backward compatibility:

Keep the same (dictionary) key in the return results, but add a new "key" property to the dictionary that gets returned for each user. That entry will contain the "key" value from the Jira API.

I was reading the source to figure this out, so I might have enough of this in my head to make a PR this weekend.

One thing to note, this is using the [deprecated] /rest/api/2/group endpoint, so the return values/schema aren't even documented there. 🙁 

[deprecated]: https://docs.atlassian.com/jira/REST/server/#api/2/group-getGroup

NfNitLoop:
*poke*   My team is still using the branch which I've got open in PR #281 so that we can access the Jira primary key.  Is there anything in particular I need to do to help get the fix merged into mainline?

PR #365, opened 2017-03-10
Fix for Issue #281: Include user key in group_members() results

NfNitLoop:
Hello!  First-time contributor, so let me know if I'm doing something wrong in the PR.  😄 

This fixes Issue #281. I based this change on top of the last release tag, 1.0.10, but it looks like some of those changes hadn't yet been merged back into your develop branch. Do I need to rebase, or is merging those back OK?

Also: I tried running `make info clean flake8 test docs` locally, but some of the tests are failing even on the base 1.0.10 tag. This change didn't cause any *more* tests to fail, though. 😛

I've got a separate [branch] that also includes a fix to the `make package` target if you want that. I needed to build the package locally to test out my change.

[branch]: https://github.com/pycontribs/jira/compare/develop...NfNitLoop:add-user-key-more?expand=1

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -803,7 +803,7 @@ class JIRA(object):
         result = {}
         for user in r['users']['items']:
             result[user['name']] = {'fullname': user['displayName'], 'email': user.get('emailAddress', 'hidden'),
-                                    'active': user['active']}
+                                    'active': user['active'], 'key': user['key']}
         return result
 
     def add_group(self, groupname):


NfNitLoop:
Hello!  This PR seems to be languishing. Is there anything I should be doing to help it move along or get rejected? Halp, I'm in limbo!  :)

jira/client.py
```diff
@@ -803,7 +803,7 @@ def group_members(self, group):
         result = {}
         for user in r['users']['items']:
             result[user['name']] = {'fullname': user['displayName'], 'email': user.get('emailAddress', 'hidden'),
-                                    'active': user['active']}
+                                    'active': user['active'], 'key': user['key']}
```
hdost: As per some of the comments in #281, we may want to maintain backwards compatibility. 

@ssbarnea What do you think about adding a block like.

```
if user.get('key'):
    result[user['name']].append('key',user['key'])

NfNitLoop: Yep, I'm the one that brought up the backwards compatibility issue. :)

This change maintains backward compatibility for users of this library by keeping the map keyed by the same value (`user['name']`), so that anyone who's depending on that behavior isn't broken when this fix is merged and deployed.   (Others had proposed changing it to be `user['key']`, which would definitely be a breaking change for anyone who was using `user['name']`, since it was only available as the dict key.)

NfNitLoop: Oh!  Are you saying we may need to *also* be backward compatible w/ older versions of the Jira API, where this value wasn't present?  I suppose we could do that. I don't know for a fact whether that field was ever added or removed. The [documentation](https://docs.atlassian.com/software/jira/docs/api/REST/7.6.1/#api/2/group-getGroup) doesn't even list the fields that that endpoint returns, so I suppose it's best to be on the safe side.

What about just doing:

~~~
result[user['name']] = {'fullname': ...
    ...
    'key': user.get('key')
}
~~~

It simplifies the code, and also makes it clear to users "Hey, there's this key value that you're not getting from your API endpoint." when they have a `user['key']` that `is None`.

ssbarnea:
Made obsolete by https://github.com/pycontribs/jira/pull/611

PR #367, opened 2017-03-16
[POC] support the plugin REST API

t-8ch:
The plugin API is behind websudo, so other authentication mechanism than
basic will probably need extra logic.

Usage example, check for plugin license expiry:

```py
for_ plugin in jira.plugins():
    if plugin.usesLicensing:
        license = plugin.license()
        for attr in ['expiryDate', 'maintenanceExpiryDate']:
            date = getattr(license, attr, None)
            if date is not None:
                print('{}: {}'.format(plugin.name, date))
                break
```

This is currently in a POC stage.
Some classes and functions should probably be moved somewhere else or better intregrated into the existing code. Pointers are welcome.
It is also missing docs which will follow after the code is done.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+max_line_length = 160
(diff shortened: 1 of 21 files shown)


hdost:
Rebased off master and created new PR.

PR #368, opened 2017-03-21
Add support for the agile API in update_sprint

mmeisel:
The new API does support this call, but it must be a POST to do a partial update.


hdost:
Functionality exists.

PR #371, opened 2017-03-23
Added Jira ServiceDesk support

igor-pavlenko:
Added following API methods support (it doesn't break previously added methods):

- create_customer() Creates a customer that is not associated with a service desk project.
- service_desk_info() Returns runtime information about JIRA Service Desk.
- create_organization() Creates an organization
- delete_organization() Delete organization
- organization() Get an organization for a given organization ID
- organizations() Returns a list of organizations in the JIRA instance.
- add_users_to_organization() Add users to organization
- remove_users_from_organization() Remove users from organization
- get_users_from_organization() Returns a list of users in organization.
- service_desks() Get a list of ServiceDesk Resources from the server
- service_desk() Returns the service desk for a given service desk Id.
- create_request() Create a new customer request and return an issue Resource for it.
- request() Get an issue Resource from the server.
- my_customer_requests() Returns all customer requests for the user that is executing the query.
- request_comments() Returns all comments on a customer request, for a given request Id/key.
- servicedesk_attachment() Add attachment (from RequestTemporaryAttachment) to request
- attach_temporary_file() Create temporary attachment file
- request_types() Returns all request types from a service desk, for a given service desk Id.
- request_type() Returns a request type for a given request type Id.
- request_type_by_name() Return request type id by it name

All method covered with tests.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+max_line_length = 160
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ tests/test-reports-*/*
 /.python-version
 /CHANGELOG
 /AUTHORS
+/tests/build
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,6 @@ tests/test-reports-*/*
 **/*.log
 /.python-version
 /CHANGELOG
+/ChangeLog
 /AUTHORS
 /tests/build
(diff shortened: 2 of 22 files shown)


asqui:
Definitely agree with @ssbarnea that it would be a good idea to segregate this code off to prevent the JIRA class growing too monolithic. This seems like a separate area of functionality and a nice boundary on which to segregate.

ssbarnea:
Closing this work needs to be rebased on master. Please feel free to create new PR, preferably splitting the work in multiple ones. It would be much easier to get small reviews in than bigger ones.

Issue #324, opened 2017-01-14
Cookies based authentication support


MaximBB:
It would be useful to add cookies based authentication support into this module
It was raised before in: https://github.com/pycontribs/jira/issues/72
It is also described in
https://jira.atlassian.com/browse/JRA-44654

I didn't assess all consequences but the result could be achieved with a small workaround in the existing code and calling constructor JIRA with "Validate" parameter. Extra line of the code is shown in bold below. 
I suppose it could be done in more reliable and elaborate way (maybe just adding one extra parameter for conditional launch of this part of the code) in one of future releases

client.py

```
def session(self):
        """Get a dict of the current authenticated user's session information."""
        url = '{server}/rest/auth/1/session'.format(**self._options)

        if isinstance(self._session.auth, tuple):
            authentication_data = {
                'username': self._session.auth[0], 'password': self._session.auth[1]}
```
            **## self._session.auth=None #Custom - remove authorization at every call**
```
            r = self._session.post(url, data=json.dumps(authentication_data))
        else:
            r = self._session.get(url)

        user = User(self._options, self._session, json_loads(r))
        return user
```

jettero:
I'm trying to make a command-line tool to automate stuff. The only login method available to me on this corp-jira is basic auth... I would really need this cookie-based session persistence to use this module. I'm surprised it's not already sorted since this module is so otherwise kickass.

Hrm.

jettero:
The above PR is how I would do it. In fact, I'm installing this at my house currently.

MaximBB:
inject one line of code into client.py as suggested above and you get needed result. Cookie will be used automatically instead of sending login/password everytime
self._session.auth=None #Custom - remove authorization at every call

It works if you initiation with "validat" 
e.g.:
jira = JIRA(server,basic_auth=(user, passw),validate=True)

jettero:
I wanted to be able to save the cookie externally and pass it back in ... possibly saving multiple cookies I can switch between; but especially the requirement for me is having the program exit, save the cookie, and use the cookie on the next run.

Just commenting out the one line doesn't achieve that result afaik.

PR #380, opened 2017-04-07
add cookie based auth support …

jettero:
This is a reply to issue #324 …

allowing us to save jira.session().session to a file and re-use it with

JIRA(server='blah', session=(cookie_name, cookie_value))

or perhaps a pickled session that has .name and .value attributes

sess = working_jira.session().session
another_working_jira = JIRA(server='blah', session=sess)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -317,7 +317,7 @@ class JIRA(object):
             self._create_cookie_based_session(session, timeout)
         elif basic_auth:
             self._create_http_basic_session(*basic_auth, timeout=timeout)
-            self._session.headers.update(self._options['headers']) ## XXX: dup?
+            self._session.headers.update(self._options['headers'])  # XXX: dup?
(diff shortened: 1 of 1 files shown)


jettero:
I'm not sure what to do with the test failures in travis-ci. Mostly they seem to not concern this patch, but I did add a raise Exception() and noticed there were no other raises anywhere else in the module... Perhaps I should go another way. 

My thinking was: if the user supplies a value for session and the session doesn't work or isn't understood, the module should let them know, rather than plodding ahead and failing in an inexplicable way later on.

hdost:
@jettero does  #484  solve your base issue?

hdost:
@jettero If you want to help with other work on this jira project, it is certainly welcome 👍 I'm somewhat new to this project, but I am attempting to triage a cleanup the project a decent amount.

PR #391, opened 2017-05-09
add verify option to JIRA object and to basic_auth

jasimmonsv:
Add verify option to Jira object and to the basic_auth process to allow option to ignore SSL errors during connection.

I do not know enough about the jwt, oauth or kerberos implementations to add the verify option there.


dgerod:
I had same problem but after looking for information I realized that now this is allowed by using the 'verify' key in the options dictionary.

```
options = {
    'server': "https://jira.cmc.local/",
    'verify': "False"
}
```

So, I think that this issue should be closed.

Issue #392, opened 2017-05-12
Cookie based authentication in addition to current available ones


juangermano-okta:
Implement cookie based authentication as described in https://developer.atlassian.com/jiradev/jira-apis/jira-rest-apis/jira-rest-api-tutorials/jira-rest-api-example-cookie-based-authentication

The idea is to take the user credentials (username and password), authenticate using the credentials in the body of a POST only, as described in the link above, and then let the cookies in subsequent requests be used for authorization.

Further, upon cookie expiration, re-authentication should work transparently.

As a side note, we've been requested by Atlassian to use this method instead of Basic Authentication due to heavy additional resource utilization when using Basic Authentication, because each request requires the authorization to be processed. Cookie based authorization is thus better from a performance point of view (as perceived by the server).

Also, please note this is different from the use case described in #72 "Support login by cookies. Don't need to take username password every time" or #324 "Cookies based authentication support", because our use case doesn't have to do with persisting the cookie. We still want to accept username and password and have everything working as similar as before as possible.

Finally, the idea of removing authorization from every request, as described in https://github.com/pycontribs/jira/issues/324#issuecomment-292647132, would probably achieve similar performance improvements, but wouldn't lead to automatic re-authentication when the cookie expires, as far as I can tell.

I've implemented this authentication and will follow with a PR shortly.

PR #393, opened 2017-05-12
Cookie based authentication in addition to current available ones #392

juangermano-okta:
Proposed implementation for cookie based authentication as described in #392

### DIFF
--- a/examples/cookie_auth.py
+++ b/examples/cookie_auth.py
@@ -1,0 +1,26 @@
+# This script shows how to connect to a JIRA instance with a
+# username and password over HTTP BASIC authentication.
+
+from collections import Counter
+from jira import JIRA
+
+# By default, the client will connect to a JIRA instance started from the Atlassian Plugin SDK.
+# See
+# https://developer.atlassian.com/display/DOCS/Installing+the+Atlassian+Plugin+SDK
+# for details.
+jira = JIRA(auth=('admin', 'admin'))    # a username/password tuple for cookie auth
+
+# Get the mutable application properties for this server (requires
+# jira-system-administrators permission)
+props = jira.application_properties()
+
+# Find all issues reported by the admin
+issues = jira.search_issues('assignee=admin')
+
+# Find the top three projects containing issues reported by admin
+top_three = Counter(
+    [issue.fields.project.key for issue in issues]).most_common(3)
+
+# import time; time.sleep(65)  # Fake cookie expiration
+
+issues = jira.search_issues('assignee=admin')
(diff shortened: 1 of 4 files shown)


DanielOverdevest:
How can I help to review this and bring this in the pycontrib:develop github?

tests/tests.py
```diff
@@ -184,6 +184,7 @@ def __init__(self):
                 else:
                     self.CI_JIRA_USER_PASSWORD = 'sd4s3dgec5fhg4tfsds3434'
 
+                self.CI_JIRA_ADMIN_EMAIL_DOMAINE_NAME = os.environ.get('CI_JIRA_ADMIN_EMAIL_DOMAINE_NAME', 'ssbarnea.33mail.com')
```
hdost: Is this related to the cookie fix?

examples/cookie_auth.py
```diff
@@ -0,0 +1,26 @@
...
+props = jira.application_properties()
+
+# Find all issues reported by the admin
+issues = jira.search_issues('assignee=admin')
+
+# Find the top three projects containing issues reported by admin
+top_three = Counter(
+    [issue.fields.project.key for issue in issues]).most_common(3)
+
+# import time; time.sleep(65)  # Fake cookie expiration
```
hdost: Please remove the commented code.

hdost:
@juangermano-okta can you rebase off dev or segment those changes to things not pertinent to the cookie auth into separate PRs?

juangermano-okta:
@hdost Sure thing.

juangermano-okta:
There's a bug in the implementation: it logins OK in response of a 401, and re-authenticates in case the cookie has expired, but it isn't handling correctly getting a 401 when accessing a resource the user is not _authorized_ to access. The response (401) is interpreted as cookie timeout, so login is attempted again. I'll need to come up with something to handle the case.

hdost:
@juangermano-okta It looks like when a resource is not authorized JIRA spits back https://en.wikipedia.org/wiki/HTTP_403 not a 401, but this is server version 7.2.7

juangermano-okta:
@hdost That's interesting, I'll check it out. I noticed the issue when using the cloud hosted version, but I've got access to a JIRA v7.3.6.

juangermano-okta:
@hdost I wasn't able to reproduce the issue with either JIRA v7.3.6 on-prem or hosted in cloud. As you pointed out, I've been receiving 403 for unauthorized resources, which is correctly handled. Since you opened #484 and resolved pending conflicts, we can continue work there.

hdost:
Sounds like a plan

PR #394, opened 2017-05-16
Accept custom authentication URL

pbabinca:
If JIRA instance doesn't accept default URL for authentication
(/rest/auth/1/session) one can set custom one.

All following works:

- key 'auth_url' in options keyword of JIRA() class
- option name 'auth_url' in 'options' section of
  ~/.jira-python/jirashell.ini
- --auth-url option of jirashell

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -195,6 +195,7 @@ class JIRA(object):
 
     DEFAULT_OPTIONS = {
         "server": "http://localhost:2990/jira",
+        "auth_url": '/rest/auth/1/session',
         "context_path": "/",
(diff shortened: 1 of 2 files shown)


hdost:
@ssbarnea I don't know how big of a use case this is, but the change is very low impact.

pbabinca:
Ah, thanks for reminding me this PR. Only later I have found that I haven't covered this PR with tests. Frankly I don't know how I can test it though.

asqui:
Also worth adding documentation for this new configuration variable...

PR #398, opened 2017-05-23
[wip] Move requirements extras to setup.cfg

ssbarnea:
Now you should be able to install only the library
using `pip install jira[library]`
`cli` adds cli command and `testing` includes all.


tox.ini
```diff
@@ -6,30 +6,25 @@ skip_missing_interpreters = true
...
-    -rrequirements-dev.txt
-    -rrequirements-opt.txt
-
 commands=
     python -m pip check
     python -m flake8
     python -m pytest
-setenv =
-    PYTHONPATH =
+; setenv =
```
hdost: Should these lines just removed?

hdost:
Wasn't sure if this was still a work in progress.

PR #403, opened 2017-06-19
Update client.py

xnuiem:
removed a " self._session.headers.update(self._options['headers'])" under basic_auth as it does it no matter what  little later after the else.


ssbarnea:
Please recreate after fixing conflicts.

PR #407, opened 2017-07-12
Set initial cookies

sausame:


ssbarnea:
Can you please explain the purpose of this change, preferably referencing a ticket.

mitsuo0114:
I guess he want to similar thing as https://github.com/pycontribs/jira/pull/413

I'm in same boat. I want to save/load cookies from previous sessions. (OAuth is disabled in our JIRA at work)

jira/client.py
```diff
@@ -323,6 +323,12 @@ def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=N
 
         self._session.max_retries = max_retries
 
+        # Set initial cookies for the session
+        if cookies is not None and type(cookies) is list:
```
dbaxa: You could use `isinstance` instead of `type`.

tubaman:
I like this idea better than my #413 because it doesn't assume where the cookies are coming from.

ericfrederich:
No need for a for loop.  Just do:
```python
if cookies is not None:
    self._session.cookies.update(cookies)
```

hdost:
@sausame Does  #484  solve the root issue?

hdost:
Closing as develop is no longer used. 
Please feel free to rebase off master and submit a new PR.

PR #413, opened 2017-07-25
add the ability to use a cookies.txt file for auth

tubaman:
At work we have an instance of JIRA that is authenticated via some SSO solution.  Also oauth is disabled.  As a workaround, I log in via a web browser and export the cookies.  I can then authenticate using those cookies.

Here I've added code to accept a cookies.txt path via the JIRA classes init and also via a jirashell option.


hdost:
@tubaman There was recently a merge of  #484  does this handle your desired usecase?

PR #414, opened 2017-07-25
Fix build: remove ChangeLog from setup.cfg

tubaman:
There's a circular dependency.  pbr builds the ChangeLog but it also
reads setup.cfg which references the ChangeLog.  Let's remove ChangeLog
from setup.cfg for now to get the build going again.


hdost:
This was fixed in a different PR.

PR #422, opened 2017-07-31
Update examples.rst

Rossmairm:
No mention of search limiting results to 50


Rossmairm:
not sure this check applies to my change

PR #425, opened 2017-08-11
Cookie Authentication for JIRA

pmfrlyn:
- Add a new cookie authentication method to avoid the problems described in
  https://jira.atlassian.com/browse/JRASERVER-26397
- Added a new module called cookieauth and requests authentication backend
  cookieauth.JIRACookieAuth
- Added a new option to the jira client called cookie_auth which works
  exactly like the existing basic_auth option

### DIFF
--- a/jira/cookieauth.py
+++ b/jira/cookieauth.py
@@ -1,0 +1,58 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from __future__ import print_function
+
+"""
+This module implements cookie based authentication, which can be used to avoid
+a problem where JIRA can become unresponsive under great volume of HTTP basic
+auth requests (https://jira.atlassian.com/browse/JRASERVER-26397)
+"""
+
+import json
+import datetime
+
+import requests
+from requests.auth import AuthBase
+
+
+
+class JIRACookieAuth(AuthBase):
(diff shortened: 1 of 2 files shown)


jira/cookieauth.py
```diff
@@ -0,0 +1,58 @@
...
+        login_url = "{}/rest/auth/1/session".format(server)
+        login_data = {'username': self.username, 'password': self.password}
+
+        session = requests.Session()
+        r = session.post(login_url,
+                         data=json.dumps(login_data),
+                         headers={'Content-Type': 'application/json'})
+
+        self.cookies = r.cookies
+        self.last_refresh = datetime.datetime.now()
```
dbaxa: IMHO you should be using the session cookie expiry time to know when to refresh the cookie.

jira/cookieauth.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class JIRACookieAuth(AuthBase):
+    """ JIRA specific cookie authentication """
+
+    def __init__(self, username, password):
+        self.username = username
+        self.password = password
+        self.last_refresh = None
+        self.check_interval = datetime.timedelta(minutes=5)
```
dbaxa: You shouldn't need to check if a session is still active every 5 minutes.

hdost:
This has been implemented slightly differently in  #484

PR #434, opened 2017-09-04
Use cookies to authenticate user deactivations in cloud instances

zoeimogen:
The /admin/rest/um/1/user/deactivate endpoint no longer works with any form of authentication except cookies - see, for example, https://community.atlassian.com/t5/JIRA-Core-questions/How-can-I-deactivate-a-user-using-curl/qaq-p/428769#M30276 and https://jira.atlassian.com/browse/ID-6230.

The fix is to send a request outside of the usual session system, as the endpoint is quite fragile and simply 401s if any other form of authorization is hinted at.


jira/client.py
```diff
@@ -2486,9 +2486,29 @@ def delete_user(self, username):
...
+                import pprint
+                pprint.pprint(r.request.__dict__)
+                if r.status_code == 200:
+                    return True
+                else:
+                    logging.warning(
+                        'Got response from deactivating %s: %s' % (username, r.status_code))
+                    return r.status_code
+            except urllib2.HTTPError as e:
+                print("Error Deactivating %s: %s" % (username, e))
```
ssbarnea: Do not use print, use logging.error instead.

zoeimogen:
Updated - also changed the existing non-Cloud error printing for consistency

hdost:
@zoeimogen There was recently a merge of  #484  does this handle your scenario?

PR #440, opened 2017-09-14
made Makefile a little more Linux-friendly, added more requirements t…

guymatz:
…o dev requirements file


Makefile
```diff
@@ -29,9 +38,23 @@ req:
...
+ifeq ($(PLATFORM), darwin)
 	which atlas-run-standalone || brew tap atlassian/tap && brew install atlassian/tap/atlassian-plugin-sdk
+else ifneq ($(wildcard /etc/debian_version),)
+	sudo sh -c 'echo "deb https://sdkrepo.atlassian.com/debian/ stable contrib" >/etc/apt/sources.list.d/atlassian_development.list'
+	sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys B07804338C015B73
+	sudo apt-get install apt-transport-https
+	sudo apt-get update
+	sudo apt-get install atlassian-plugin-sdk
+else ifneq ($(wildcard /etc/redhat-release),)
+	curl https://marketplace.atlassian.com/download/plugins/atlassian-plugin-sdk-rpm/version/42380 -o /tmp/atlassian-plugin-sdk.noarch.rpm
```
dbaxa: The specification destination location is not secure. Please create a temporary directory to save it under using `mktemp -d`

Makefile
```diff
@@ -29,9 +38,23 @@ req:
 install: prepare
 	$(PREFIX)pip install .
 
+# https://developer.atlassian.com/docs/getting-started/set-up-the-atlassian-plugin-sdk-and-build-a-project/install-the-atlassian-sdk-on-a-linux-or-mac-system#InstalltheAtlassianSDKonaLinuxorMacsystem-Homebrew
 install-sdk:
-	# https://developer.atlassian.com/docs/getting-started/set-up-the-atlassian-plugin-sdk-and-build-a-project/install-the-atlassian-sdk-on-a-linux-or-mac-system#InstalltheAtlassianSDKonaLinuxorMacsystem-Homebrew
+ifeq ($(PLATFORM), darwin)
```
dbaxa: IMHO we shouldn't assume that people are running as a user with sudo or root access - i they don't have root / sudo access we can fail out and point them at a location to find out how to install the sdk.

Makefile
```diff
@@ -29,9 +38,23 @@ req:
...
 	which atlas-run-standalone || brew tap atlassian/tap && brew install atlassian/tap/atlassian-plugin-sdk
+else ifneq ($(wildcard /etc/debian_version),)
+	sudo sh -c 'echo "deb https://sdkrepo.atlassian.com/debian/ stable contrib" >/etc/apt/sources.list.d/atlassian_development.list'
```
dbaxa: If this fails do we call back to informing the user that the sdk cannot be installed ? (for example in the case of a user not having sudo access)

Makefile
guymatz: Thanks for the comments!!  Yes, will do this  . . .

Makefile
guymatz: I'm doing a check to see if the user is root (sudo will also show them as root)

guymatz:
@dbaxa I'm getting this error from travis:
error in setup command: Error parsing /home/travis/build/pycontribs/jira/setup.cfg: IOError: [Errno 2] No such file or directory: 'ChangeLog'

I don't think this has anything to do with my code . . .  any ideas?   Thanks!

hdost:
@guymatz If you want to rebase of of develop the changelog issue should be fixed for Travis for the time being.

PR #445, opened 2017-09-25
Fix move to backlog

peterst28:
Without this change, move_to_backlog() throws a NotImplementedError.  The change is modeled on add_issues_to_sprint().


peterst28:
Pulled in extra commits.

PR #450, opened 2017-09-28
Add cookies to JIRA constructor

### LINKED ISSUE #449: Allow cookies to be passed to JIRA session
Other things like proxies and validate are already passed... should allow cookies as well.

This is required for me in our single-sign-on environment

ericfrederich:
This fixes #449


hdost:
This may be unnecessary now that #484 has been merged.

PR #452, opened 2017-09-29
fix python 2.7 unicode decode error when jira returns unicode chars

nikepan:


dbaxa:
@ssbarnea can we merge this?

dbaxa:
A test demonstrating the fix would be nice.

nikepan:
Test case - JIRA with russian error messages. This module used in other files (for example client.py). Only in this file python try concatenate unicode and not unicode strings if error from JIRA is unicode. Problem actual only for python 2

ssbarnea:
Closing because it needs full rebase in order be make it mergeable.

PR #460, opened 2017-10-18
Add support for custom session cookies to JIRA

fetzerch:
This patch allows to connect to a JIRA server behind a single-sign-on
system that requires the client to pass a cookie for enabling HTTP
basic authentication.

The implementation has been done in a generic way by extending the
`options` parameter of the JIRA constructor to support a `cookies`
dict (similar to `headers`) which is passed to request's session
cookies.

Unfortunately just passing the cookie with a "Cookie" header is not
enough because the cookie has to remain also in all further HTTP
requests of the session.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -268,6 +268,7 @@ class JIRA(object):
         * verify -- Verify SSL certs. Defaults to ``True``.
         * client_cert -- a tuple of (cert,key) for the requests library for client side SSL
         * check_update -- Check whether using the newest python-jira library version.
+        * cookies -- A dict of custom cookies that are sent in all requests to the server.
 
     :param basic_auth: A tuple of username and password to use when establishing a session via HTTP BASIC
         authentication.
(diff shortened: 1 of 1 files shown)


hdost:
I don't know if this is still necessary now that  #484  has been merged.
@ssbarnea what do you think?

fetzerch:
I'm not sure if #484 really covers our use case where the SSO-System requires a cookie to be set.
From what I can see this is unrelated to JIRA's cookie authentication.

I'll give it a try and report back.

fetzerch:
Ok as I thought, the JIRA cookie authentication is unrelated. Our SSO-System will only allow HTTP basic authentication if a special cookie is set in all requests. Other authentication methods are unfortunately not allowed.

Since the change is rather unintrusive, I hope it can be accepted.

fetzerch:
https://github.com/pycontribs/jira/issues/509 describes the same problem.
@anamariaradut got in touch with me via E-Mail and confirmed as well that this patch solves the issue.

Stealthii:
If we're merging this, the only suggestion I would have is to add a line to the documentation to describe this option (so people know it's there to use it)

fetzerch:
I've updated the documentation accordingly.

fetzerch:
I've rebased the change. Are there still any blocking issues with it and is there a chance to get it it?

PR #466, opened 2017-11-09
Fix install by removing use of ChangeLog

hdost:
- ChangeLog is currently in .gitignore


hdost:
@ssbarnea Will do 👍  just trying to get the build all working and then I'll separate it into different

PR #467, opened 2017-11-09
Remove References to Changelog

hdost:


hdost:
Removed this as it is currently in the root level .gitignore and seems to be problematic as part of the build.

hdost:
@ssbarnea the problem is that Travis fails to start the build successfully similarly to as mentioned in https://github.com/pycontribs/jira/pull/440#issuecomment-331463261

hdost:
This PR is obsolete.

PR #469, opened 2017-11-13
Clean code styling which was previously causing PEP8 failures

hdost:


tests/tests.py
```diff
@@ -97,10 +97,9 @@ def hashify(some_string, max_len=8):
 def get_unique_project_name():
     jid = ""
     user = re.sub("[^A-Z_]", "", getpass.getuser().upper())
-
     if user == 'TRAVIS' and 'TRAVIS_JOB_NUMBER' in os.environ:
-        # please note that user underline (_) is not suppored by
-        # jira even if is documented as supported.
+        # please note that user underline (_) is not supported by
+        # JIRA even if is documented as supported.
```
zenpizza: should read, 'even if it is documented'

hdost:
@ssbarnea wasn't sure if someone else approves whether I could merge.

tests/tests.py
```diff
@@ -97,10 +97,9 @@ def hashify(some_string, max_len=8):
 def get_unique_project_name():
     jid = ""
     user = re.sub("[^A-Z_]", "", getpass.getuser().upper())
-
     if user == 'TRAVIS' and 'TRAVIS_JOB_NUMBER' in os.environ:
-        # please note that user underline (_) is not suppored by
-        # jira even if is documented as supported.
+        # please note that user underline (_) is not supported by
+        # JIRA even if it is documented as supported.
```
dbaxa: This should be `Jira` now.

PR #471, opened 2017-11-14
Passthrough states as a csv string

### LINKED ISSUE #382: jira 1.0.10 implements the state parameter of sprints() incorrectly
When retrieving sprints for a board using .../board/{id}/sprint, it is possible to filter the sprints by state ('closed', 'active', 'future') JIRA 7.x expects the state parameter to be a single string with the state values inside it (e.g. 'active,future'), but the JIRA.sprints() method splits this parameter into a list ('active,future' => ['active', 'future']) As a result, only sprints matching the first state value supplied are returned.

The following should be removed from the sprints() method(unless they're needed for backward compatibililty - my knowledge doesn't go back that far)

    params = {}
    if state:
        if isinstance(state, string_types):
            state = state.split(",")
        params['state'] = state

should be

    params = {}
    if state:
        params['state'] = state

(Sorry about lack of diff, pull request, etc, but my code is behind a locked down environment that won't let me pull the diffs out)

hdost:
Fixes #382


ssbarnea:
Please rebase your fork and reopen the PR as a simple fix, now you can make it directly against master branch.

Issue #424, opened 2017-08-10
jira: Can't authenticate with kerberos


KKoukiou:
I am running the following code to establish Jira connection with kerberos authentication.

``` python
from jira import JIRA
import logging

JIRA_SERVER = 'https://myserver.example.com'

logging.basicConfig(level=logging.DEBUG)

jira = JIRA(
    options={
        'server': JIRA_SERVER,
        'verify': False
    },
    kerberos=True
)
```
I don't see some Error but when I try to get the user that I am connected with I get anonymous user

``` python
In [1]: jira.current_user()
/home/kkoukiou/sandbox/Jira/lib/python2.7/site-packages/urllib3/connectionpool.py:858: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
DEBUG:urllib3.connectionpool:https://projects.engineering.redhat.com:443 "GET /rest/api/2/serverInfo HTTP/1.1" 200 None
DEBUG:requests_kerberos.kerberos_:handle_other(): Handling: 200
DEBUG:requests_kerberos.kerberos_:handle_response(): returning <Response [200]>
Out[1]: 'anonymous'
```

Apparently something is wrong here.
I tried to check the sources of python jira a bit, and I saw  that if I remove the OPTIONAL for the mutual_authentication I get error to authenticate.
https://github.com/pycontribs/jira/blob/develop/jira/client.py#L2284

Am I doing something wrong or is kerberos authentication broken?

Thanks in advance

hdost:
@KKoukiou we recently merged https://github.com/pycontribs/jira/pull/472 into develop can you test with that and see if it fixes your issue?

KKoukiou:
@hdost, still can't authenticate from kerberos. Nothing changed in the logs, from the post above.

Also I took a look in the PR you sent me and looks problematic from a first sight.
https://github.com/pycontribs/jira/blob/69fac2d1325589a37b5ab40542d1b6b2ca5ff418/jira/client.py#L2310-L2314
Is it correct that the only available option would be to disable mutual authentication?
Not setting at all kerberos_options will fail for example, or setting mutual_authentication to OPTIONAL will fail as well.

hdost:
@pbabinca has volunteered to take up this issue

pbabinca:
@KKoukiou can you test if authentication works with referenced PR?

[...]

PR #472, opened 2017-11-14
Make Kerberos option mutual_authentication configurable

pbabinca:
Before this change `mutual_authentication` was set to
`requests_kerberos.OPTIONAL`. This option value doesn't change.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -221,7 +221,7 @@ class JIRA(object):
     JIRA_BASE_URL = Resource.JIRA_BASE_URL
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
-    def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False,
+    def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False, kerberos_options=None,
                  validate=False, get_server_info=True, async=False, logging=True, max_retries=3, proxies=None,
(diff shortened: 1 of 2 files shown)


hdost:
@ssbarnea This looks fine by me

ssbarnea:
@hdost from now on please add yourself as reviewer and use the review option as you can. Review button is on the "Files changed" tab.

jira/client.py
```diff
@@ -2294,15 +2297,24 @@ def _create_oauth_session(self, oauth, timeout):
...
         from requests_kerberos import HTTPKerberosAuth
-        from requests_kerberos import OPTIONAL
+        from requests_kerberos import OPTIONAL, DISABLED
+
+        mutual_authentication = OPTIONAL
+        if kerberos_options.get('mutual_authentication') == 'DISABLED':
+            mutual_authentication = DISABLED
+        else:
+            raise ValueError("Unknown value for mutual_authentication: %s" %
+                             kerberos_options['mutual_authentication'])
```
KKoukiou: Hey, we really want this behavior here?
If not set to DISABLED it will fail.

pbabinca: Thanks for spotting this! You are very right.

hdost: @pbabinca don't know if you want to work on a fix, but you could relate it to https://github.com/pycontribs/jira/issues/424. If so I can assign it to you.

pbabinca: Yes, I will look into that one.

PR #474, opened 2017-11-14
Fix Continuous Integration Testing

hdost:
Added a Project Template to the JiraTestManager which will allow the slug fixture to run properly


hdost:
So while testing doesn't completely work now, this does now go through the full gambit of testing. I think this is a good step towards then fixing a lot of the existing failures that show up in the the suite

PR #477, opened 2017-11-15
Fix Tests

hdost:
This is a another batch of changes to fix the test suite. The 8 remaining ones seem that they may be a combination of problems with the library and the test suite. They will require a little more work.


hdost:
@asqui Could you take a look at this?

hdost:
Thank you 👍

PR #478, opened 2017-11-15
Add contribution guidelines

hdost:


asqui:
Very nice and comprehensive.

Have you worked out how to run the tests locally? If so, it would be great to capture that detail here too.

hdost:
That's a very good point, I'll put that together before we merge I'll try
to get that done by tomorrow.

*Harold Dost* | @hdost

asqui:
Hi @hdost,

Sorry for the extremely late reply on this. I don't have as much time to spend on open source projects as I'd like, and I think @ssbarnea is in the same unfortunate position, so your efforts on this project are greatly appreciated!

I believe anyone who is in [this group](https://github.com/orgs/pycontribs/teams/pycontribs/members) has the ability to merge to this project, which includes you!

Great work on sorting through the backlog of open issues. I think stable tests is one of the main challenges for this project, because of he complexities and edge-cases around interacting with a JIRA Cloud instance, so any work on improving this would be extremely beneficial.

Keep up the good work!

PR #479, opened 2017-11-15
Make OPTIONAL default and allowed value of mutual_authentication

pbabinca:
Before this change Kerberos authentication worked only if
Kerberos option mutual_authentication was explicitly set to DISABLED.
OPTIONAL was not default value anymore but it wasn't among allowed
values of mutual_authentication. Therefore these authentications failed
with raised exception:

```
ValueError: Unknown value for mutual_authentication
```

Fixes: #424
Relates: #472

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2306,8 +2306,9 @@ class JIRA(object):
         from requests_kerberos import HTTPKerberosAuth
         from requests_kerberos import OPTIONAL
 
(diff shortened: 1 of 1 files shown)


hdost:
@ssbarnea this looks good to merge.

ashcrow:
@ssbarnea 👍 PTAL

PR #483, opened 2017-11-20
Fix blanket resource creation for unknown self

astralblue:
The signature of `Resource.__init__()` differs from those of concrete
resource types; previous code erroneously mapped `(options, session, raw)`
onto `(path, options, session)`; the resulting resource did not avail the
raw contents.


astralblue:
The encouraging use case for this was to examine `i.fields.status.statusCategory.key` (such as `new`) for an issue `i`.  Previously, statusCategory was created from a `None` raw value, and fields of the category was not available.

PR #484, opened 2017-11-21
Cookie Based Authentication #392

### LINKED ISSUE #392: Cookie based authentication in addition to current available ones

hdost:
Supercedes #393 
Resolves #392 

Cherry Picked Commits Done by @juangermano-okta

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -44,6 +44,16 @@ authentication. These sessions will apply to all subsequent calls to the JIRA ob
 
 The library is able to load the credentials from inside the ~/.netrc file, so put them there instead of keeping them in your source code.
 
+Cookie Based Authentication
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Pass a tuple of (username, password) to the ``auth`` constructor argument::
+
+    authed_jira = JIRA(auth=('username', 'password'))
+
+Using this method, authentication happens during then initialization of the object. If the authentication is successful,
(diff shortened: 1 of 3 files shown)


hdost:
Looks like I will need to fix a conflict , but after that, should hopefully
be good.

Harold Dost | @hdost

hdost:
No new failed tests.

PR #488, opened 2017-12-22
doc/examples: fix jql example which had wrong syntax.

AnneTheAgile:


AnneTheAgile:
Hmm. re travis fail.

asqui:
I don't understand. What was wrong with the syntax? Your pull request changes it to output the issue key and summary (which now disagrees with the comment on the line above) but seems to use the same syntax.

What am I missing?

hdost:
The actual issue with the syntax is `print` vs `print()` it's not actually an issue with the list comprehension. Python2.7 uses `print` no paren Python3.x uses `print()`

Could add the parentheses, for the example and then make a note saying to remove parentheses if using 2.7.

hdost:
I don't know that I would have merged that.  `print()` only works if you import futures

asqui:
Hmmm, okay, how about this:
```
    # Summaries of my last 3 reported issues
    for issue in jira.search_issues('reporter = currentUser() order by created desc', maxResults=3):
        print('{issue.key}: {issue.fields.summary}'.format(issue=issue))
```

This should be 2/3 compatible, no?

Issue #465, opened 2017-11-01
Documentation: The JIRA.__init__ docstring is not rendered in the API docs


asqui:
Looks like the lengthy and detailed docstring for JIRA.__init__() is not being rendered in the API docs at http://jira.readthedocs.io/en/latest/api.html#jira

My attention was drawn to this by issue #455.

hdost:
Fixed in #494

PR #494, opened 2018-01-11
Update client.py to fix the __init__ docstring

hdost:
It appears that sphinx (as would make sense) does not attempt to
document dunder / _private_ methods. As a result i have moved the
information into the class documentation which I believe had yielded the
desired results. We can now see the parameters for the __init__ in the
documentation which should hopefully alleviate some misunderstandings

Ref #465


hdost:
@asqui Thank you

PR #502, opened 2018-01-30
Fix pbr dependency

Achimh3011:
Fixes 501.


Achimh3011:
Sorry, clicked too fast.

PR #518, opened 2018-03-01
Merge master onto develop

dbaxa:
The commit that we need to fix pbr usage during test execution seems to be aba65f5. That is, pbr is required because of it being imported in jira/__init__.py.


Stealthii:
@dbaxa it looks like the `master` branch was recently rebased, for the 1.0.14 tag release with recent fixes. Not sure why (or who) but this PR won't work with those changes unfortunately. I've tried redo-ing it [here](https://github.com/pycontribs/jira/compare/develop...Stealthii:release/master-to-develop?expand=1), let me know what you think!

dbaxa:
@Stealthii looks good.

PR #522, opened 2018-03-01
Changes by dbaxa which skip user group tests when contacting pycontribs instance.

hdost:


dbaxa:
@hdost I'll open a pr with the user/group test fixes. This pull request is missing a number of required commits/changes.

PR #525, opened 2018-03-01
Add support for providing a location_type and location_id when creating a board in cloud.

dbaxa:
Sem-Ver: feature


hdost:
Is this new parameter on the Greenhopper API? I did not realize that was still being updated.

dbaxa:
@hdost  yes.

PR #527, opened 2018-03-02
Remove some old and obsolete methods.

dbaxa:
Sem-Ver: api-break


ssbarnea:
Please recreate after fixing conflicts, and try to make it a small as possible.

PR #532, opened 2018-03-15
Update the location of the test pypi uri.

dbaxa:


dbaxa:
@ssbarnea the location of TestPyPi has moved to https://test.pypi.org/legacy/ , see https://packaging.python.org/guides/using-testpypi/ for more details.

PR #538, opened 2018-03-18
implement async issues search

ssbarnea:


tests/tests.py
```diff
@@ -30,7 +30,7 @@
 _non_parallel = True
 if platform.python_version() < '3':
     _non_parallel = False
-
+    import mock
```
dbaxa: You cannot import mock without adding it as a dependency for python versions < 3.3.

tests/tests.py
```diff
@@ -1939,6 +1956,63 @@ def test_session_server_offline(self):
         self.assertTrue(False, "Instantiation of invalid JIRA instance succeeded.")
 
 
+class AsyncTests(unittest.TestCase):
+
+    def setUp(self):
+        self.jira = JIRA('https://jira.atlassian.com', logging=False,
+                         async=True, validate=False, get_server_info=False)
+
+    def test_fetch_pages(self):
```
dbaxa: This test will not pass without 086afbbc96c1f981b997cf9e1351bda8ba81ed6e and 05e8b09456eaa9b3a2742033cc145e841bf1822e

dbaxa:
@ssbarnea let's continue this in #540.

PR #540, opened 2018-03-18
Add async support to fetch pages

dbaxa:
Sem-Ver: feature

### DIFF
--- a/jira/utils/__init__.py
+++ b/jira/utils/__init__.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 """JIRA utils used internally."""
 from __future__ import unicode_literals
-import json
 import threading
 
 from jira.resilientsession import raise_on_error
@@ -77,8 +76,10 @@ def threaded_requests(requests):
 
 def json_loads(r):
     raise_on_error(r)
-    if len(r.text):  # r.status_code != 204:
-        return json.loads(r.text)
-    else:
+    try:
+        return r.json()
+    except ValueError:
         # json.loads() fails with empty bodies
-        return {}
+        if not r.text:
+            return {}
+        raise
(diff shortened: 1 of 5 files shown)


jira/resilientsession.py
```diff
@@ -57,12 +57,6 @@ def raise_on_error(r, verb='???', **kwargs):
     # for debugging weird errors on CI
     if r.status_code not in [200, 201, 202, 204]:
         raise JIRAError(r.status_code, request=request, response=r, **kwargs)
-    # testing for the WTH bug exposed on
```
ssbarnea: Why are you removing this? This is a workaround for https://jira.atlassian.com/browse/JRASERVER-41559 which was never fixed.

dbaxa: @ssbarnea perhaps I misunderstood the code that I removed. It looked to me like it wasn't actually doing anything.

ssbarnea: Remove any change that is unrelated to this feature and I will merge it as it seems to have its own tests which are passing.

dbaxa: @ssbarnea will do. I'll break the other changes out into individual prs.

jira/utils/__init__.py
```diff
@@ -77,8 +76,10 @@ def threaded_requests(requests):
 
 def json_loads(r):
     raise_on_error(r)
-    if len(r.text):  # r.status_code != 204:
-        return json.loads(r.text)
-    else:
+    try:
+        return r.json()
```
dbaxa: @ssbarnea I'd very much like to keep this change / make this change as well. If it needs to go in another PR please let me know :-).

dbaxa:
@ssbarnea can we merge this ?

PR #544, opened 2018-04-04
Add hashable functionality to Issue.

pworters:
In Python 3, the hashable is not longer inherited by default.
Adding __hash__ and __eq__ to Issue class.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -432,6 +432,14 @@ class Issue(Resource):
         if raw:
             self._parse_raw(raw)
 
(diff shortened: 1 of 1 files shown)


jira/resources.py
```diff
@@ -432,6 +432,14 @@ def __init__(self, options, session, raw=None):
         if raw:
             self._parse_raw(raw)
 
+    def __hash__(self):
```
dbaxa: @pworters please add a test for this.

PR #548, opened 2018-04-06
use post when jql in params

### LINKED ISSUE #336: Error in the JQL Query: The character '%' is a reserved JQL character.
Hi.

I trying to do the following:
`result = jira.search_issues('filter=12345')`
But get the following:
`Traceback (most recent call last):
  File "playlist-lamps.py", line 84, in <module>
    main()
  File "playlist-lamps.py", line 46, in main
    be_issues = jira.search_issues('filter=23233', validate_query=False)
  File "c:\Python27\lib\site-packages\jira\client.py", line 1740, in search_issues
    issues = self._fetch_pages(Issue, 'issues', 'search', startAt, maxResults, search_params)
  File "c:\Python27\lib\site-packages\jira\client.py", line 390, in _fetch_pages
    resource = self._get_json(request_path, params=page_params, base=base)
  File "c:\Python27\lib\site-packages\jira\client.py", line 2165, in _get_json
    r = self._session.get(url, params=params)
  File "c:\Python27\lib\site-packages\jira\resilientsession.py", line 149, in get
    return self.__verb('GET', url, **kwargs)
  File "c:\Python27\lib\site-packages\jira\resilientsession.py", line 145, in __verb
    raise_on_error(response, verb=verb, **kwargs)
  File "c:\Python27\lib\site-packages\jira\resilientsession.py", line 56, in raise_on_error
    r.status_code, error, r.url, request=request, response=r, **kwargs)
jira.exceptions.JIRAError: JiraError HTTP 400 url: https://jira.host.net/rest/api/2/search?jql=filter%253D12345&validateQuery=False&startAt=0&maxResults=50
        text: Error in the JQL Query: The character '%' is a reserved JQL character. You must enclose it in a string or use the escape '\u0025' instead. (line 1, character 7)`

Some double encoding is happening?

Jira module 1.0.9
Jira version: 7.1.9

264768502:
Sometimes, use a filter.jql to search issue, http get couldn't work well.
Switch to use post according formal recommended
it can  also fix #336


dbaxa:
Please add a test for this change.

264768502:
original test already cover it

jira/client.py
```diff
@@ -2483,7 +2483,10 @@ def _get_url(self, path, base=JIRA_BASE_URL):
 
     def _get_json(self, path, params=None, base=JIRA_BASE_URL):
         url = self._get_url(path, base)
-        r = self._session.get(url, params=params)
```
dbaxa: I don't think we should be conditionally switching from get to post here like this.

asqui: @264768502 : Is there any reason not to just POST all the time?

264768502: I get jql from filter first, them try to use jql to do issue search
JIRA server feedback input parameter with invalid string.
After switch to use post, search success.
But it has a little strange, after I switch another os, issue go. I have no idea why, maybe use different requests.

My commit maybe not best resolution.
As you know,  http url has length limitation
and 'get' need encode  parameter to url, this will involve some unknown issue.

how about add parameter to let user control use post or get?
In my opinion, check url will give a little hard to maintain

@asqui I don't want change original design too much
Only switch to post when search issue.

PR #549, opened 2018-04-09
Added applicationKeys parameter to add_user

jrosen:
The jira api has a parameter to allow you to specify which applications the user has access to.  By default, a user has access to all applications.  If you want to specify which applications a user can use, specify them with a list (e.g. `applicationKeys=['jira-software']`) . If a user is only being created to be a reporter for Jira Service Desk purposes, use `applicationKeys=[]`.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2949,7 +2949,7 @@ class JIRA(object):
         return False
 
     def add_user(self, username, email, directoryId=1, password=None,
-                 fullname=None, notify=False, active=True, ignore_existing=False,applicationKeys=None):
+                 fullname=None, notify=False, active=True, ignore_existing=False, applicationKeys=None):
         """Create a new JIRA user.
 
         :param username: the username of the new user
--- a/jira/client.py
+++ b/jira/client.py
@@ -2949,7 +2949,7 @@ class JIRA(object):
         return False
 
     def add_user(self, username, email, directoryId=1, password=None,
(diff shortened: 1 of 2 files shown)


dbaxa:
Please add a test for this change.

jrosen:
@dbaxa I added a test including my change.  It's kinda lame, as the keys are not returned in the "get user" api call.  I can't find a way to actually find that information for an existing user via api.

jrosen:
Hey @dbaxa, can you see why my test isn't being noticed by CodeCov?  It looks like it should be running it.

dbaxa:
@jrosen don't be overly worried about the coverage check.

jrosen:
@dbaxa are we good here?

jira/client.py
```diff
@@ -2949,7 +2949,7 @@ def create_project(self, key, name=None, assignee=None, type="Software", templat
         return False
 
     def add_user(self, username, email, directoryId=1, password=None,
-                 fullname=None, notify=False, active=True, ignore_existing=False):
+                 fullname=None, notify=False, active=True, ignore_existing=False, applicationKeys=None):
```
dbaxa: I'd argue that applicationKeys should use snake case and be `application_keys`.

jrosen: Good call. I've updated the argument.  I had it consistent with the api call, but it makes more sense to be consistent with the rest of the package.

PR #550, opened 2018-04-10
Merge Develop onto master

dbaxa:

PR #551, opened 2018-04-10
Install tox-travis before using tox in the build.

dbaxa:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,30 +16,30 @@ notifications:
 jobs:
   include:
   - stage: docs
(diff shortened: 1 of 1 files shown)


asqui:
Why is this required? What benefits does it bring?

I'm not objecting, just trying to understand :-)

dbaxa:
@asqui @ssbarnea broke the travis build in https://github.com/pycontribs/jira/pull/550/commits/d7c766d81e3003789b2f07163d2aa187211e2040 by using tox before it was available. This change fixes that.

PR #555, opened 2018-04-24
add_worklog user assignee field updated, now you can specify user to …

Chetan07j:
There was issue in adding worklog, user field is not working in original code, so I have updated add_worklog jira rest request input parameter.
With this author and updateAuthor fields are in working state


Chetan07j:
Sorry not working it seems issue from jira only.
If I found any solution will update back

huashanms:
@Chetan07j Hi, Is there any way to solve this issue now?

PR #559, opened 2018-04-29
enabled cspell check

ssbarnea:


.gitchangelog.rc
```diff
@@ -1,61 +1,3 @@
-##
```
hdost: Is this relevant to the spell check?

hdost:
I believe you've made these changes in a different PR already.

PR #441, opened 2017-09-19
Feature/webhooks

guymatz:
Resolves Issue #272

Issue #272, opened 2016-10-27
webhooks


guymatz:
Please add the ability to CRUD webhooks, per https://developer.atlassian.com/jiradev/jira-apis/webhooks#Webhooks-rest

guymatz:
PR 441

PR #566, opened 2018-05-11
#272 Add Webhooks

hdost:
This is a rebase of #441 from @guymatz

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -81,6 +81,7 @@ from jira.resources import User
 from jira.resources import Version
 from jira.resources import Votes
 from jira.resources import Watchers
+from jira.resources import Webhook
 from jira.resources import Worklog
 
 from jira import __version__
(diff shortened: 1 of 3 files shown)


hdost:
@guymatz

dbaxa:
I'd prefer to see some tests for this change.

hdost:
Yea after I saw the codecov report I was thinking the same.

Harold Dost | @hdost

hdost:
@dbaxa Added.

jira/client.py
```diff
@@ -2422,6 +2424,95 @@ def kill_session(self):
...
+        else:
+            fields_dict = {}
+            for field in fieldargs:
+                fields_dict[field] = fieldargs[field]
+            data['fields'] = fields_dict
+
+        url = self._get_url('webhook', base=self.WEBHOOK_BASE_URL)
+        json_data = json.dumps(fieldargs)
+        try:
+            r = self._session.post(url, data=json_data)
```
dbaxa: You should be able to specify `json=fieldargs` instead of needing to explicitly `json.dump` it first..

jira/client.py
```diff
@@ -2422,6 +2424,95 @@ def kill_session(self):
...
+        except Exception as e:
+            print("Error Creating webhook with json_data:\n%s\n\n%s" % (json_data, e))
```
dbaxa: I think we should be raising an exception here or at least logging what has occurred.

ssbarnea: Never ever use print() inside a library, use logging instead.

jira/client.py
hdost: @dbaxa can you provide an example of what you mean I think I may be missing something.

hdost: @dbaxa Are you saying just forego the `json_data` and just replace with `fieldargs`

Issue #565, opened 2018-05-08
proposal: dropping develop branch and use only master


ssbarnea:
I think that in the last years the use of two branches, master and develop proved to be more of a confusion factor and caused more maintenance work.

I would like to propose dropping use of develop branch and using directly master.

Please use github up/down voting buttons on this as I don't want to make a decision without getting enough feedback from a signifiant number of contributors (probably 5 at least).

hdost:
I'm good with it, we should just make one last merge, and then drop it.

hdost:
This is done.

asqui:
Develop branch still exists. Should we delete this to avoid future confusion?

hdost:
There's one breaking change that needs to be merged from it at this point
and it can be deleted after that. Otherwise if someone wants to cherry pick
the one commit from #577 and put it on it's own PR then we can delete
develop.

________________________________________

Harold Dost | @hdost

asqui:
Ah, I didn't appreciate that #577 is still based off `develop`.

What is blocking merging of #577; is it just that the build is red, or something else?

(Just trying to get a feel for how long it might be keeping the `develop` branch alive for.)

hdost:
Like I said, I can just cherry pick the change and open a new PR, then we
could delete develop, but that won't happen until Monday.

________________________________________

Harold Dost | @hdost

asqui:
Now that master is green, any objections if I just go ahead and merge #577 and #605 and delete `develop` branch?

hdost:
Only if our next release should have breaking changes, @ssbarnea thoughts?

________________________________________

Harold Dost | @hdost

asqui:
Ah, I was under the impression that we already have breaking changes on master but I appear to be mistaken.

So is the model that breaking PRs get queued up and wait for the next breaking release, at which point we merge all PRs tagged with the "SemVer: breaking-change" label?

Only problem with this model is that long-lived branches may develop more merge conflicts over time, but I can't see any way to avoid this.

Issue #397, opened 2017-05-22
DeprecationWarning: 'async' and 'await' will become reserved keywords in Python 3.7


stebunovd:
It turns out that 'async' parameter which used in [a couple of methods](https://github.com/pycontribs/jira/search?utf8=✓&q=async&type=) prevents this library to be compatible with the future of Python. These warnings can be seen when running tests under Python 3.6 with `-Wd` option.

I guess this is not urgent since Python 3.7 is not released yet, but at the same time it affects public library API, so probably it makes sense to deprecate this parameter in advance and let developers fix their code.

urosjarc:
Soon python 3.7 will be default version on ubuntu. When will be this issue resolved?

ssbarnea:
I am curious about what it would be the best way to approach this. How did others solve the issue?

hdost:
Currently there is an open PR on pylama https://github.com/klen/pylama/pull/120

hdost:
OpenStack solution: https://github.com/openstack/mistral/commit/4d8b30270ab5edf201d76b09322c374b446d3ec9

stebunovd:
Python 3.7 is out! 

Do you need any help with developing a fix for this issue, or maybe with testing?

hdost:
Superceded by #603

PR #577, opened 2018-05-17
Remove email_user

hdost:
Per the proposal: #565

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2579,30 +2579,6 @@ class JIRA(object):
                                 ". Specify the 'contentType' parameter explicitly.")
                 return None
 
-    def email_user(self, user, body, title="JIRA Notification"):
-        """(Obsolete) Send an email to an user via CannedScriptRunner."""
-        url = self._options['server'] + \
-            '/secure/admin/groovy/CannedScriptRunner.jspa'
-        payload = {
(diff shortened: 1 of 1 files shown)


hdost:
This seems to have made the difference, I'm not sure if we want to move to a 2.0, though with #397 we will likely need to anyhow.
```
commit 389ec969cf5324175d1773dc464ec33645d096eb               
Author: David Black <dblack@atlassian.com>                    
Date:   Thu Mar 1 14:30:59 2018 +1100                         

    Remove support for renaming users in Jira versions less than 6.0.0.                                                      
                               
    Sem-Ver: api-break         
                               
```

dbaxa:
@hdost we can always revert that commit if we don't want to release a 2.0.0 version just yet.

hdost:
I think we would have to cherry pick the other commits since I believe it
relies on git history. Not sure how advanced it is.

Harold Dost | @hdost

asqui:
Is this PR currently waiting on the next major release to get merged?

ssbarnea:
Indeed the history prevent a merge so we would have to cherry pick already merged changes and close it after. 

You have me ok to do the work on that as my current load is high. Shortly develop is here only until we merge all useful changes from it to master. Please avoid opening new PR to it.

hdost:
Ok so I did, some parlor tricks and then rebased off of master and it looks like the only thing not already merged was another SemVer breaking change...

PR #580, opened 2018-05-18
Fix blanket resource creation for unknown self

hdost:
The signature of Resource.__init__() differs from those of concrete
resource types; previous code erroneously mapped (options, session, raw)
onto (path, options, session); the resulting resource did not avail the
raw contents.

Rebase of:  #483 from @astralblue


dbaxa:
Can we add a test for this change?

PR #583, opened 2018-05-25
Change default template

ReedOnly:
Original template for create new project is removed from Jira and not working anymore

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2958,7 +2958,7 @@ class JIRA(object):
             possible_templates = [template_name]
 
         # https://confluence.atlassian.com/jirakb/creating-a-project-via-rest-based-on-jira-default-schemes-744325852.html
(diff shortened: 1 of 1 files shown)


hdost:
As of which version is this true?

ReedOnly:
For current Jira Cloud, `1001.0.0-SNAPSHOT#100085`.

hdost:
Ideally we want to support backwards compatiblility given that this API is
used with more than just a the cloud offering.

Harold Dost | @hdost

ssbarnea:
I am closing this PR as rejected because is not complete.

1. our testing is passing in JIRA Cloud, so the template was removed on *your* instance, not in general
2. proposed template is agile specific and 100% sure it will break on jira classic
3. there is no test assigned
4. i would have preferred a better logic for detecting the default template instead of using a hardcoded value

Please don't be upset about having this rejected, is just that we want to avoid breaking the library for other hundreds of users. Feed free to make another one that is a bit safer.

PR #586, opened 2018-05-31
Enable Attachment Test

dbaxa:
- The way the JIRA returns upon completion of a DELETE is a 204


dbaxa:
Potentially we could merge this instead of #577.

hdost:
There are a couple others as well, but this approach should work

PR #473, opened 2017-11-14
Kerberos mutual authentication in jirashell

pbabinca:
This builds up on #395 and makes #472 available in jirashell.

PR #589, opened 2018-06-14
Make Kerberos option mutual_authentication configurable

tuukkamustonen:
Cherry-picked from #473 (by @pbabinca):

    Before this change mutual_authentication was set to requests_kerberos.OPTIONAL.
    This option value doesn't change.

(the original PR is closed for being out-of-date and the author seems somewhat inactive.)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2483,6 +2483,13 @@ class JIRA(object):
         from requests_kerberos import HTTPKerberosAuth
         from requests_kerberos import OPTIONAL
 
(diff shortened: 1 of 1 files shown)


hdost:
This code is not necessary, please take a look at: https://github.com/tuukkamustonen/jira/blob/9e712de386cb37106fb66afc79c7c3d27ce79a41/jira/client.py#L2493-L2499

tuukkamustonen:
Right, good to see it's already there!

PR #593, opened 2018-06-18
Add user group resource

### LINKED ISSUE #591: GroupPicker fields cannot be read out like other fields.
I have a project where I have added a GroupPicker field to an issue. When I try to read the issue, I cannot (easily) access he name of the group listed:

```python
issue = jira.issue('SIS-11', fields = ['customfield_10800'])
print('issue', issue)

print('grouplist', issue.fields.customfield_10800)
print('group', issue.fields.customfield_10800[0])
print('group-name', issue.fields.customfield_10800[0]._session['name'])  # Weirdly works...

print('expected', issue.fields.customfield_10800[0].name)  # Boom!
```

yields (lines reordered):

```
issue SIS-11
grouplist [<JIRA Resource at 4402162936>]
group <JIRA Resource at 4402162936>
group-name jira-software-users
  File "jira-python/jira/resources.py", line 160, in __getattr__
    return self[item]
TypeError: 'Resource' object is not subscriptable

During handling of the above exception, another exception occurred:

  File "jira-python/sample.py", line __, in <module>
    print('expected', issue.fields.customfield_10800[0].name)  # Boom!
  File "jira-python/jira/resources.py", line 173, in __getattr__
    if hasattr(self, 'raw') and item in self.raw:
TypeError: argument of type 'NoneType' is not iterable
```

The response of the REST call looks quite sensible:

```
GET http://jira:8080/rest/api/2/issue/SIS-11?os_authType=basic&fields=customfield_10800

{
  "expand": "renderedFields,names,schema,operations,editmeta,changelog,versionedRepresentations",
  "id": "10600",
  "self": "http://jira:8080/rest/api/2/issue/10600",
  "key": "SIS-11",
  "fields": {
    "customfield_10800": [
      {
        "name": "jira-software-users",
        "self": "http://jira:8080/rest/api/2/group?groupname=jira-software-users"
      }
    ]
  }
}
```

As far as I can see from debugging, the problem lies in creating a bland Resource in line 921. However, the constructor signature of Resource is different from those of the subclasses. Therefore, information is misplaced in the created object.

https://github.com/pycontribs/jira/blob/0243eac8570366170af98789e38c1a55c1c6eda3/jira/resources.py#L919-L928

https://github.com/pycontribs/jira/blob/0243eac8570366170af98789e38c1a55c1c6eda3/jira/resources.py#L963-L969

Achimh3011:
Fixes #591 by implementing UserGroupResource and introducing UnknownResource.


Achimh3011:
Is `develop`the right base branch? If so, why are there that many commits that are on `master`but not on `develop`?

hdost:
Master is going to be the new working branch and develop is going to go
away.

Harold Dost | @hdost

hdost:
I would suggest checking out the latest master and cherrypicking 6740c51

Achimh3011:
OK, will do that. If the documentation regarding branching model is inaccurate, I'll raise an issue for that as well.

Achimh3011:
Ready to merge if the Travis fixes are ok (for now).

Achimh3011:
OK, reset branch. If you want my (preliminary) travis fixes, you can find them at 2482db9e28409d7ec7e2ed49b4cab6a69007d1d6.

Achimh3011:
anything missing to merge the PR?

Achimh3011:
Hmm, so adding the case to the existing test cases (and thus improving test coverage by 0.2% points according to codecov) is not sufficient? I'll see what I can do to do better.

Maybe mentioning what you expect on http://jira.readthedocs.io/en/latest/contributing.html would be helpful.

hdost:
@Achimh3011 my appologies, I was looking at an email of the CodeCov I received previously.

Let me take a second look

jira/resources.py
```diff
@@ -953,16 +970,26 @@ def dict2resource(raw, top=None, options=None, session=None):
...
     r'user\?username.+$': User,
+    r'group\?groupname.+$': UserGroup,
     r'version/[^/]+$': Version,
     # GreenHopper specific resources
     r'sprints/[^/]+$': Sprint,
     r'views/[^/]+$': Board}
 
 
+class UnknownResource(Resource):
+    """A JIRA project."""
```
hdost: This does not seem to be an accurate description of the unknown resource maybe a little explanation why this is here.

jira/resources.py
```diff
@@ -764,6 +764,23 @@ def __eq__(self, other):
         return str(self.name) == str(other.name)
 
 
+class UserGroup(Resource):
```
ssbarnea: Any reason for not naming it Group instead of UserGroup? I would prefer simple unless we have a good reason.

jira/resources.py
Achimh3011: OK

Issue #411, opened 2017-07-18
adopt pre-commit.com


ssbarnea:
It would be nice to adopt use of pre-commit hooks as documented at http://pre-commit.com/hooks.html

hdost:
@ssbarnea Were you thinking to use all of them or just a subset?

ssbarnea:
Merged, from now on we will be able to add extra pre-commit checks on one-by-one change, making easier to review. Feel free to raise PRs, no need to add extra tickets, can link to this one.

PR #595, opened 2018-06-20
Adopt pre-commit hooks

ssbarnea:


asqui:
Nice idea to include these hooks!

The build is red and at least some of the failures look related to your config changes...

(Would this PR resolve #411 when merged?)

PR #597, opened 2018-06-20
Fix cspell problems

hdost:


asqui:
Nice! Beat me to it...

(I wonder why @ssbarnea didn't spot these errors when he added cspell in 777a325?)

hdost:
@asqui same :confused:

PR #598, opened 2018-06-21
Update README to reflect branch model

hdost:


README.rst
```diff
@@ -86,9 +86,9 @@ Development
 
 Development takes place on GitHub_, where the git-flow_ branch structure is used:
 
-* ``master`` - contains the latest released code.
-* ``develop`` - (default branch) is used for development of the next release.
```
ssbarnea: Can remove the develop and feature branches, no need to document them.
Correct the description: master does not contain only released code, only the tags are release.
Also remove the reference to git-flow, it may confuse users as we are not really using it.

PR #599, opened 2018-06-21
travis fixes

ssbarnea:


hdost:
@ssbarnea Looks good, I also know that when a merge is done it still seems to fail. I think that all we would need to do is not perform the deploy step unless there has been a tag. https://docs.travis-ci.com/user/environment-variables#Default-Environment-Variables `TRAVIS_TAG` What do you think?

PR #600, opened 2018-06-21
Travis fixes

ssbarnea:


.travis.yml
```diff
@@ -5,7 +5,7 @@ cache:
   - directories:
       - "node_modules"
       - $HOME/.cache
-sudo: false
+sudo: true
```
dbaxa: Is there an easy way we can avoid needing sudo in travis (perhaps using virtualenv(s)) ?

ssbarnea: Yes, there is but I was trying to avoid the SSL warnings we got from pip. I am still playing with this change. Since the builds are running in containers I am no longer sure that there is any cost on using sudo.

You will see more changes on this PR so no real need to review it until we get it passing on Travis (at least the non unittest part should pass without issues in the end).

Achimh3011: One does not need `sudo` if `pip` and `setuptools`are installed within the virtualenv.

ssbarnea:
Obsoleted

Issue #603, opened 2018-06-28
`async` is a reserved keyword on Python 3.7


gordonshieh94:
**Describe the bug**
Keyword arguments that are named `async` now conflicts with the reserved keyword of the same name, introduced in Python 3.7.

**To Reproduce**
In a Py3.7 interpreter with jira installed:
```
from jira.client import Comment  
```

**Stack Trace**
Error I got from tox while running unit tests in our CI
```
    from jira.client import Comment  # noqa: E402
E     File "/tmp/.tox/py37-linux/lib/python3.7/site-packages/jira/client.py", line 225
E       validate=False, get_server_info=True, async=False, logging=True, max_retries=3, proxies=None,
E                                                 ^
E   SyntaxError: invalid syntax
```

**Version Information**
Python Interpreter: 3.7.0
jira-python: 1.0.14
OS: any

hdost:
Duplicate of #397

PR #604, opened 2018-06-29
api: py37 support and rename async to async_

ssbarnea:
Breaking change as is renames all async parameters to async_
in order to assure pyhon 3.7 compatibility.

Fixes: #603

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ cache:
   - directories:
       - "node_modules"
       - $HOME/.cache
-sudo: false
 os:
   - linux
 stages:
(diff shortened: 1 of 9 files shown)


asqui:
No `Sem-Ver: api-break` tag on the commit? :-)

ssbarnea:
I am not used to Sem-Ver but now I am working on adopting pre-commit hooks and once they are in, if we have a plugin for Sem-Ve we can adop it.

I am fond of having well written commit messages but only if we have a way to lint them, otherwise it worth nothing because sooner or later someone will put bad things in. I spend effectively days trying to rewrite history and get rid of that commit message in CJK which was breaking so many tools.

ssbarnea:
Regarding async/await, we have to *await* until travis will add py37 support (not dev, official release). Many dependencies are not building on 3.7 due to special SSL requirements in python.

Another thing that I don't like is that Python didn't give any official recommendation regarding how to deal with renames, like which names to use. As a result everyone is messing his API in a different way. 

If someone finds some references or stats regarding what others did, it would be great to see them.

.travis.yml
```diff
@@ -59,6 +72,12 @@ jobs:
       env: TOXENV=py36 PYTHON='3.6' PYENV_VERSION='system'
       after_success:
         - bash <(curl -s https://codecov.io/bash) -e TOX_ENV
+    - stage: test
```
hdost: Since we are adding 3.7 does it make sense to remove 3.4?

ssbarnea: Nope, py34 is still in production for till April 2019.

jira/client.py
```diff
@@ -325,7 +325,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False, kerberos_options=None,
```
dbaxa: One option would have been to accept kwargs and then check if `async` had been set.

ssbarnea:
@dbaxa I am aware about kwargs and I considered it but I decided to go the easy way due:
- harder effort to implement
- would mess the auto-complete in editors due to lose of signature information

I struggled with the kwargs dilemma for years and I did not find a good solution yet but if someone knows one, feel free to shoot. I will be more than happy to learn about it and use it.

As we are going to break few apis with the next release, it would be a good idea to do more refactoring now as this should not happen too often.

BTW, I plan to make few pre-releases to pypi to allow people to test them using `pip install --pre jira`. This makes me wonder if we should use this occasion to adopt a new versioning. like "7.0" to follow jira compatibility, maybe "18.0" to follow the year, or just 2.0. Again, open to suggestions.

hdost:
For versioning I would be down with either the JIRA or year/month releasing.
I think more towards the year month only because it helps prevent us from
breaking semver too much without actually needing to fully follow semver.
We tend to get a bunch of complaints about breaking semver.

Harold Dost | @hdost

asqui:
I'd vote for pure semver since we're not that much on top of supporting latest features aligned to JIRA releases. i.e. do we envisage ever signing off on a release and saying "okay, now we have full support for all the features in JIRA version x"?

+1 on not adding **kwargs. I really object to **kwargs because it really harms usability and compatibility.

pieterbergmans:
Hi, I'm receiving this error with jira==1.0.15 and python 3.7.  I know you're working hard on getting this fixed.  I was just wondering if you had a date for when this will be released?  Thanks so much.

chevell:
@pieterbergmans for what it's worth, you can install the latest `master` branch now by pointing `pip` directly to this repo:

```
pip install --upgrade git+https://github.com/pycontribs/jira.git
```

PR #605, opened 2018-06-29
Remove support for renaming users in Jira versions less than 6.0.0.

hdost:
Sem-Ver: api-break


hdost:
Splinter of #577

PR #609, opened 2018-06-30
travis: remove test pypi server

ssbarnea:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,17 +75,6 @@ deploy:
       tags: true
       branch: master
       repo: pycontribs/jira
-  - provider: pypi
-    server: https://test.pypi.org/legacy/
-    user: ssbarnea
-    password:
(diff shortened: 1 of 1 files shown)


asqui:
Holy crap, the master build is green!

Great work @ssbarnea!!!  🥇:clap:

hdost:
For real, that's great!

ssbarnea:
The big challenge is to make all these PRs green too! 

-- sorin

asqui:
I guess we should rebase all open PRs from the newly green master?

Seems like the sort of thing we should be able to do from the GitHub UI but the only option I can see is "Rebase and Merge", which is not what we want...

Alternatively (and more generally useful, especially for long-lived PRs) would be an option to have Travis run a build on the result of merging the PR branch to master. Do you know if this is possible?

I would then expect this build to be green for any PR branches that have no merge conflicts and don't introduce any regressions...

PR #611, opened 2018-06-30
 group_members should return sorted user keys

ssbarnea:
Updates group_mebers() to return a sorted dictionary
of users where user key is used instead of name attribute
which can be a duplicate.

Fixes: #281

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -991,9 +991,11 @@ class JIRA(object):
 
         result = {}
         for user in r['users']['items']:
-            result[user['name']] = {'fullname': user['displayName'], 'email': user.get('emailAddress', 'hidden'),
-                                    'active': user['active']}
-        return result
+            result[user['key']] = {'name': user['name'],
+                                   'fullname': user['displayName'],
+                                   'email': user.get('emailAddress', 'hidden'),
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -991,9 +991,11 @@ def group_members(self, group):
 
         result = {}
         for user in r['users']['items']:
-            result[user['name']] = {'fullname': user['displayName'], 'email': user.get('emailAddress', 'hidden'),
-                                    'active': user['active']}
-        return result
+            result[user['key']] = {'key': user['key'],
+                                   'fullname': user['displayName'],
```
asqui: Do we want to also include `user['name']` in the dict, since this is no longer returned as the key?

jira/client.py
```diff
@@ -991,9 +991,11 @@ def group_members(self, group):
...
+                                   'email': user.get('emailAddress', 'hidden'),
+                                   'active': user['active']}
+        return OrderedDict(sorted(result.items(), key=lambda t: t[0]))
```
asqui: Minor: Could use `itemgetter(0)` in place of the `lambda`.

PR #614, opened 2018-07-02
[POC] support the plugin REST API

hdost:
The plugin API is behind websudo, so other authentication mechanism than
basic will probably need extra logic.

Usage example, check for plugin license expiry:

```py
for plugin in jira.plugins():
    if plugin.usesLicensing:
        license = plugin.license()
        for attr in ['expiryDate', 'maintenanceExpiryDate']:
            date = getattr(license, attr, None)
            if date is not None:
                print('{}: {}'.format(plugin.name, date))
                break
```

Supercedes: #367

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -62,6 +62,7 @@ from jira.resources import Issue
 from jira.resources import IssueLink
 from jira.resources import IssueLinkType
 from jira.resources import IssueType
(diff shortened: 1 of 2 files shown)


t-8ch:
@ssbarnea I think this would still be nice to have

PR #616, opened 2018-07-02
Use cookies to authenticate user deactivations in cloud instances

hdost:
Supersedes: #434 from @zoeimogen

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2657,8 +2657,6 @@ class JIRA(object):
             # We can't use our existing session here - this endpoint is fragile and objects to extra headers
             try:
                 r = requests.post(url, headers={'Cookie': self.authCookie, 'Content-Type': 'application/json'}, data={})
-                import pprint
-                pprint.pprint(r.request.__dict__)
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -2631,9 +2631,28 @@ def delete_user(self, username):
...
+                r = requests.post(url, headers={'Cookie': self.authCookie, 'Content-Type': 'application/json'}, data={})
+                if r.status_code == 200:
+                    return True
+                else:
+                    logging.warning(
+                        'Got response from deactivating %s: %s' % (username, r.status_code))
+                    return r.status_code
+            except urllib2.HTTPError as e:
+                logging.error(
+                        "Error Deactivating %s: %s" % (username, e))
```
ssbarnea: I am not sure we want to catch exceptions and hide them. I think that we should raise a JiraError instead, or at least return False. Otherwise consumers of this function would not be able to know if call succeeded or not.

Same applies for the code below.

Also I wonder if we can avoid duplicating so much code between cloud mode and self-hosted one.

PR #624, opened 2018-07-08
Async improvement: Allow setting the number of worker threads for JIRA._fetch_pages

dchevell:
**Summary**
This PR allows users to configure the number of worker threads used when `JIRA._fetch_pages` is run in async mode.

**Background**
Async mode for `JIRA._fetch_pages` is implemented using `requests_futures` module, which itself is a wrapper for using `requests` with `concurrent.futures.ThreadPoolExecutor`. `requests_futures` uses a default of 2 worker threads instead of  `ThreadPoolExecutor`'s default of 5 × # of CPU's. Since `JIRA._fetch_pages` provides no interface for overriding the `requests_futures` default, it results in async mode only using up to 2 threads.

**Performance tests**
Here's the result of some basic tests, running against Apache's public facing Jira instance at https://issues.apache.org/jira:

```
from time import time

from jira import JIRA

JIRA_USER = 'myusername'
JIRA_PASS = 'mypassword'

auth = (JIRA_USER, JIRA_PASS)
jql = 'created>"-5d"'

print('## No async')
jira = JIRA(server='https://issues.apache.org/jira', basic_auth=auth, async_=False)
start = time(); len(jira.search_issues(jql, maxResults=False)); end = time(); print('Time to complete:', end-start)

print('## Async with default of 2 worker threads')
jira = JIRA(server='https://issues.apache.org/jira', basic_auth=auth, async_=True)
start = time(); len(jira.search_issues(jql, maxResults=False)); end = time(); print('Time to complete:', end-start)

print('## Async with 5 worker threads')
jira = JIRA(server='https://issues.apache.org/jira', basic_auth=auth, async_=True, async_workers=5)
start = time(); len(jira.search_issues(jql, maxResults=False)); end = time(); print('Time to complete:', end-start)
```

```
## No async
Time to complete: 52.44412875175476
## Async with default of 2 worker threads
Time to complete: 24.775141954421997
## Async with 5 worker threads
Time to complete: 11.562492847442627
```

**Other considerations**
This library currently has two very different methods of executing async behaviour. `JIRA._fetch_pages`is dependent on `requests_futures` use of `concurrent.futures` library, using 2 worker threads, and other operations such as updating issues depend on `JIRA.async_do` which provides an interface to a manually managed threadpool of up to 10 worker threads. At some point these two approaches should be unified by a common interface, so that the defined number of worker threads apply to all async operations available in this library, and so all async operations share a common model for executing tasks concurrently. That's beyond the scope of the proposed change here however. One step at a time!

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -292,6 +292,7 @@ class JIRA(object):
     :param get_server_info: If true it will fetch server version info first to determine if some API calls
         are available.
     :param async_: To enable asynchronous requests for those actions where we implemented it, like issue update() or delete().
+    :param async_workers: Set the number of worker threads for async operations.
     :param timeout: Set a read/connect timeout for the underlying calls to JIRA (default: None)
         Obviously this means that you cannot rely on the return code when this is enabled.
     """
@@ -307,6 +308,7 @@ class JIRA(object):
         "verify": True,
         "resilient": True,
         "async": False,
+        "async_workers": 2,
         "client_cert": None,
         "check_update": False,
         "headers": {
@@ -325,7 +327,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False, kerberos_options=None,
-                 validate=False, get_server_info=True, async_=False, logging=True, max_retries=3, proxies=None,
+                 validate=False, get_server_info=True, async_=False, async_workers=2, logging=True, max_retries=3, proxies=None,
                  timeout=None, auth=None):
         """Construct a JIRA client instance.
 
@@ -374,6 +376,7 @@ class JIRA(object):
         :param get_server_info: If true it will fetch server version info first to determine if some API calls
             are available.
         :param async_: To enable async requests for those actions where we implemented it, like issue update() or delete().
+        :param async_workers: Set the number of worker threads for async operations.
         :param timeout: Set a read/connect timeout for the underlying calls to JIRA (default: None)
         Obviously this means that you cannot rely on the return code when this is enabled.
         :param auth: Set a cookie auth token if this is required.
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -325,7 +327,7 @@ class JIRA(object):
     AGILE_BASE_URL = GreenHopperResource.AGILE_BASE_URL
 
     def __init__(self, server=None, options=None, basic_auth=None, oauth=None, jwt=None, kerberos=False, kerberos_options=None,
-                 validate=False, get_server_info=True, async_=False, logging=True, max_retries=3, proxies=None,
+                 validate=False, get_server_info=True, async_=False, async_workers=2, logging=True, max_retries=3, proxies=None,
```
dbaxa: Why default to `2`?

dchevell: @dbaxa because the current library *already* defaults to 2, I'm just being conservative about changing a library's expected behaviour. I always set this to 5 and I think it's a reasonable setting, but I'm erring on preserving the current behaviour and allowing users to override it.

dbaxa: @dchevell sure but just because the library default to 2 isn't a good reason to keep it at 2. Changing the worker count to 5 is probably going to be more useful / closer to what people who use the async code want.

hdost:
I believe you may need to rebase your branch.

dchevell:
@hdost I've rebased, but master is still failing?

hdost:
@dchevell The master is in a stable state now, my recommendation would be to blow away your local master and re-check it out. 
Branch off of the correct master and then cherry pick your commits, if you'd prefer I can do it. Either way I would suggest at a minimum resetting your master back a number of commits and then pulling.

ssbarnea:
I do agree that the async implementation in the library needs extra work but my knowledge of async approaches is very limited. One thing is for sure: there is no guarantee regarding API and compatibility for any async features in the library. Mainly we consider it experimental and each new release could break it.

PR #625, opened 2018-07-08
Prevent mutation of original fields list JIRA.search_issues

### LINKED ISSUE #418: search_issues modifies fields parameter if it is a list
I encountered a side effect of ```search_issue``` method. When I'm passing list of fields as list (not as comma separated string), then that method is doing "translation" and replaced 'key' with 'issuekey' in my list.
Later on when I'm trying to get field (key) as I requested, I'm getting error since issuekey field does not exist in Issue object. **As workaround:** I'm cloning my list of fields before passing it to search_issue method.

**Recommendation**
Do not alter fields parameter inside search_issue method if list is passed (or raise an exception if fields is not a string).

Version I'm using is ```jira==1.0.10```

dchevell:
This is to fix #418 , where JIRA.search_issues can be seen to mutate the original list of fields passed to it.

To avoid this behaviour, this change simply checks if `fields` is a list and creates a copy of it.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2064,6 +2064,9 @@ class JIRA(object):
         if fields is None:
             fields = []
 
+        if isinstance(fields, list):
+            fields = fields.copy()
+
         if isinstance(fields, string_types):
             fields = fields.split(",")
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -2063,11 +2063,9 @@ class JIRA(object):
         """
         if fields is None:
             fields = []
-
-        if isinstance(fields, list):
+        elif isinstance(fields, list):
             fields = fields.copy()
-
-        if isinstance(fields, string_types):
+        elif isinstance(fields, string_types):
             fields = fields.split(",")
 
         # this will translate JQL field names to REST API Name


jira/client.py
```diff
@@ -2064,6 +2064,9 @@ def search_issues(self, jql_str, startAt=0, maxResults=50, validate_query=True,
         if fields is None:
             fields = []
 
+        if isinstance(fields, list):
```
asqui: Minor observation: Might be better to use `if`/`elif`/`elif` construct here to avoid multiple branches executing for the same input.

i.e. for `fields==None` this will reset fields to an empty list in the first branch, and then immediately create an (unnecessary) copy of said list in the second branch.

Thanks for your contribution to the project!

chevell: Fair point - perhaps I was trying to keep it *too* simple :) PR updated.

asqui:
Looks good but I just don't want to merge changes while the master build is red...

chevell:
@asqui no concerns from me, if I were you I wouldn't be merging to master until my builds were green again too! :) 

Once you've fixed that I'll rebase my PR's to ensure their builds are green too.

hdost:
@dchevell Master is green you should be able to rebase.

dchevell:
@hdost master still isn't green as far as I can see, so holding off until build fixes are merged.

dchevell:
Updated PR based off fixed master

dchevell:
@ssbarnea happy to send a PR with different behaviour (e.g. @Achimh3011 's suggestion of just coercing input into a list) but reverting this seems like a poor idea - the previous behaviour mutates input unexpectedly, which is certainly a bug. The API docs define the valid input for `fields` as a string, so ensuring support for data types beyond strings and lists seems more like an undocumented feature.

hdost:
I think it would also be prudent to create some tests around this

rajsahae:
I know others have mentioned this but I would like to reiterate that this causes a bug when you pass a list of strings.

`
00:51:54    File "/usr/local/lib/python2.7/site-packages/jira/client.py", line 2084, in search_issues
00:51:54      fields = fields.copy()
00:51:54   AttributeError: 'list' object has no attribute 'copy'
`

This was a breaking change, I'm surprised it was merged?

dchevell:
@rajsahae as mentioned above, [according to the package documentation](
https://jira.readthedocs.io/en/master/api.html#jira.JIRA.search_issue) the only supported input type for fields is `str`. `list` was never a supported type for fields, it only worked because of the internal implementation. It's important to distinguish between breaking the package's API (which this change does not) vs breaking people's usage of undocumented behaviour.

Having said I can see I made an error in my implementation. My mistake was using `list.copy()` which only works in python 3.3+. I've [opened a PR](https://github.com/pycontribs/jira/pull/677) that uses `fields = list(fields)` and also updates the documentation to explicitly acknowledge `list` as a supported type for fields.

rajsahae:
@dchevell fair enough. Thanks for the info and the new PR.

Issue #619, opened 2018-07-05
Replace pycrypto with pycryptodome


Rovanion:
When using oauth this python package uses the pycrypto library. The problem is that pycrypto  stopped being developed four years ago and has not been touched since. This entails a number of issues, probably including security issues, but what got me started down this path was the dependency hell it landed me in. 

Since pycrypto is four years old it depends on an even older version of cryptography, and long story short this led me into pinning old versions of two other libraries than pycrypto that pycrypto depended on. 

But the story doesn't end there, which probably is a good thing considering for how long pycrypto has gone unmaintained. The software I'm building is kept in a docker container during development, a container running Alpine Linux. Alpine has switched from OpenSSL to LibreSSL and in the end it was impossible to both satisfy the old requirements of pycrypto and those of postgresql.

The solution to this issue is fortunately quite straight forward: Switching to a maintained fork named pycryptodome. 

In my limited testing this library works as a drop in replacement for pycrypto without any changes to the jira python library.

For further reference here is an issue against PyCrypto which has been linked by many others in the same situation as you: https://github.com/dlitz/pycrypto/issues/238

ssbarnea:
@Rovanion I don't mind doing a switch to a maintained library but I see few problems:
- new library is hosted under individual github account, which is single-point of failure kind of issue.
- i do not see build status icons on the main page
- apparently this fork is not published on pypi - or if it is, is not documented in the readme
- concerns regarding availability of precompiled versions on major linux distros: deb and rpm based ones.

Switching a dependency can be tricky if you are not switching it with the right thing. Lets gather some extra info.

Rovanion:
Yes absolutely. I'm starting to think that we might have read different pages since we're seeing such different information. But here is the information you requested:

1.  new library is hosted under individual github account, which is single-point of failure kind of issue.

I see your point. But on the other hand a single point of failure is better than something which has already failed.

2.  i do not see build status icons on the main page

They're linked in the readme: https://github.com/Legrandin/pycryptodome  
https://travis-ci.org/Legrandin/pycryptodome  
https://ci.appveyor.com/project/Legrandin/pycryptodome  

3.  apparently this fork is not published on pypi - or if it is, is not documented in the readme

The readme says that it comes in two versions through pypi:  
https://pypi.org/project/pycryptodome/  
https://pypi.org/project/pycryptodomex/

4.  concerns regarding availability of precompiled versions on major linux distros: deb and rpm based ones.

[...]

PR #627, opened 2018-07-09
Added an explicit dependency to oauthlib in order to avoid depending on the outdated pycrypto library

Rovanion:
See:  https://github.com/pycontribs/jira/issues/619


Rovanion:
Seems to me like the build failures are unrelated to the proposed changes.

asqui:
Looks like @ssbarnea might have done a force push to master? I can't work out how else that first commit appeared on your branch.

In any case, if you remove that first commit from your branch (a938853) it should be good to merge.

ssbarnea:
Wait for me to fix the travis release and once I am done a simple rebase would work.

ssbarnea:
Fix already merged, this is nolonger needed. Lets hope this change would not upset other consumers.

PR #630, opened 2018-07-10
update pyyaml dependency

ssbarnea:
Removes divergent pyyaml dependency as 3.13 was released
and marked as compatible with all python versions.
For the moment we avoid bumping it to >=4 in order to avoid
conflicts with other packages.


hdost:
I have no idea what is going on with GitHub but it looks like it's freaking out, not exactly sure why.

PR #638, opened 2018-07-17
unicode encoding applied to resource setattribute to patch UnicodeEnc…

frank-belldev:
…odingError

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -911,6 +911,11 @@ class RequestType(Resource):
 
 # Utilities
 
+def setattrWithUnicode(obj, name, value):
+    try:
+        setattr(obj, name, value)
+    except Exception:
+        setattr(obj, name.encode('utf-8'), value.encode('utf-8'))
 
 def dict2resource(raw, top=None, options=None, session=None):
     """Convert a dictionary into a Jira Resource object.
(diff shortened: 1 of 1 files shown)


jira/resources.py
```diff
@@ -911,6 +911,11 @@ def __init__(self, options, session, raw=None):
 
 # Utilities
 
+def setattrWithUnicode(obj, name, value):
```
ssbarnea: Why not “upgrading” setattr() instead?
I do not find adding a new method as something easy to maintain or not confusing.
if possible we should fix original method

dbaxa: Please use snake case.

jira/resources.py
```diff
@@ -911,6 +911,11 @@ def __init__(self, options, session, raw=None):
...
+    try:
+        setattr(obj, name, value)
+    except Exception:
```
dchevell: You should only catch specific exceptions you've planned for. When an unexpected exception happens your program _should_ stop, rather than proceeding blindly forward, which makes catching `Exception` undesirable behaviour.

hdost:
@frank-belldev could you add a test for this

frank-belldev:
@hdost ,

I tested it on Python 2.7 and setattr itself seems to include unicode error.
That is, it's broke once name or value includes special character, for example, Spanish or Russian
 character.
I hope that you test it to handle unicodeError.

ssbarnea:
Closing as it was not updated in a long time. Also it has several major issues, including the use of broad "except Exception".

PR #641, opened 2018-07-19
Add FOSSA License Bling to README.rst

hdost:


ssbarnea:
I am against make use of FOSSA public because I don't like their business model: mainly they offer only 5 repositories free, even if they are open source. This was a dealbreaker for pycontribs.

We can use it but I without advertising its use on our README :)

PR #674, opened 2018-08-17
Fix Issue comparsion with None

mateusduboli:
When you try to compare an issue with `None` code blows up and tells that `None` doesn't have a `.id` attribute.

This simple short circuits the comparison so that if `other` is `None` it doesn't tries to access a non existing field.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -590,7 +590,7 @@ class Issue(Resource):
 
     def __eq__(self, other):
         """Comparison method."""
-        return self.id == other.id
+        return other is not None and self.id == other.id
 
 
 class Comment(Resource):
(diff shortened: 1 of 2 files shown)


jira/resources.py
```diff
@@ -519,7 +519,7 @@ def permalink(self):
 
     def __eq__(self, other):
         """Comparison method."""
-        return self.id == other.id
+        return other != None and self.id == other.id
```
hdost: This should be `other is not None`

ssbarnea:
@mateusduboli please fix the linter issue and we will merge it. You can run the lint yourself using `tox -e lint`

PR #677, opened 2018-08-26
Improved support for passing list of fields to search_issues

dchevell:
The public docs for pycontribs don't acknowledge `list` as a supported type for passing fields via `search_issues`, however support for it is implemented.

This fixes a bug introduced by [this](https://github.com/pycontribs/jira/pull/625#) PR, where list copying was implemented using functionality only present in Python 3. This PR uses more universally supported method of copying lists, and also adds `list` as a supported input type in the documentation.

I've intentionally implemented this to work with other iterables (sets, tuples, etc) however that should again be considered an internal implementation detail.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2355,9 +2355,9 @@ class JIRA(object):
         :type maxResults: int
         :param validate_query: Whether or not the query should be validated. (Default: True)
         :type validate_query: bool
-        :param fields: comma-separated string of issue fields to include in the results.
+        :param fields: comma-separated string or list of issue fields to include in the results.
             Default is to include all fields.
-        :type fields: Optional[str]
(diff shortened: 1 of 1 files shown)


dchevell:
This PR hasn't been touched in a long time so I've fixed conflicts introduced in the intervening months. Please re-check if it's good to merge.

hdost:
So I'm not exactly sure why you're getting the failure that you are, going to try to re build just in the case it was some sort of fluke.

hdost:
Nope it looks like one of our non-pinned dependencies broke linting _sigh_ I'll take a look Monday

hdost:
@dchevell ok :face_with_head_bandage: the CI is fixed now. If you rebase now, it should be set.

PR #680, opened 2018-08-30
sprint state change for public API

simzacks:
Modified the update_sprint function, so that it allows modifying the state in the public API.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3148,7 +3148,7 @@ class JIRA(object):
 
     def update_sprint(self, id, name=None, startDate=None, endDate=None, state=None):
 	# For the 'agile' API, all the parameters are required to be in the payload. 
-	
+
         payload = {}
         if name:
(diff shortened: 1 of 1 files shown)


ssbarnea:
Am I wrong to believe that we should eradicate the "agile_rest_path" option completely from the code? if so please update it to make the change in all places.

hdost:
The main question I think we should answer is what version of Jira do we
want to support and then when did that stop being relevant?

simzacks:
My understanding of the code that I changed was that at some point the
public api didn't support state change,  but now it does.
If there's another way to tell what Jira version is being used, I would be
happy to change the if statement to include only supported versions.

simzacks:
@ssbarnea  @hdost - It looks like this conversation died. What do we have to do to get the module to support updating state in the public API? We are using JIRA v7.3.6 and this works. I don't have other versions of Jira to test.

PR #683, opened 2018-09-01
Fix warnings

BoboTiG:
Fix several warnings:
```python
test: AttachmentTests.test_1_add_remove_attachment
tests/tests.py:484 unclosed file <_io.BufferedReader name='.../tests/tests.py'>

test: SessionTests.test_session_with_no_logged_in_user_raises
resilientsession.py:86 invalid escape sequence \\%'
tests/tests.py:1289 invalid escape sequence \\[
tests/tests.py:1738 invalid escape sequence \\.
```
Also fix an unclosed socket (showed by the test `OtherTests.test_session_invalid_login`):
```python
client.py:512 unclosed <ssl.SSLSocket fd=13, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('192.168.0.14', 44314), raddr=('104.192.139.50', 443)>
```

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -514,16 +514,15 @@ class JIRA(object):
     def close(self):
         session = getattr(self, "_session", None)
         if session is not None:
+            try:
+                session.close()
+            except TypeError:
+                # TypeError: "'NoneType' object is not callable"
+                # Could still happen here because other references are also
+                # in the process to be torn down, see warning section in
+                # https://docs.python.org/2/reference/datamodel.html#object.__del__
+                pass
             self._session = None
-            if self.sys_version_info < (3, 4, 0):  # workaround for https://github.com/kennethreitz/requests/issues/2303
-                try:
-                    session.close()
-                except TypeError:
-                    # TypeError: "'NoneType' object is not callable"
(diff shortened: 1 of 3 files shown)


jira/client.py
```diff
@@ -514,16 +514,15 @@ def __del__(self):
...
+            try:
+                session.close()
+            except TypeError:
+                # TypeError: "'NoneType' object is not callable"
+                # Could still happen here because other references are also
+                # in the process to be torn down, see warning section in
+                # https://docs.python.org/2/reference/datamodel.html#object.__del__
+                pass
             self._session = None
-            if self.sys_version_info < (3, 4, 0):  # workaround for https://github.com/kennethreitz/requests/issues/2303
```
ssbarnea: I see that you want too call this code for all versions of Python. Are you sure this is needed? I don't see how this change fits with the others.

BoboTiG: I tested on Python 2.7, 3.6 and 3.7. And this condition  was the source of the socket unclosed. Also tests are OK.
So I think it is safe to say that this code is OK for all Python version :)

PR #708, opened 2018-10-19
Fix create_board failure

merazola:
location_id (project key) was not translated into into the database ID (number).
Tested using the following script:

```
#Initiate connection to JIRA.
print "Authenticating with server..."
authed_jira = JIRA(options, basic_auth=(username, password))
print "Ok."

#Read project name
project = authed_jira.project(prjkey)
prjname = project.name

#Create board for everything

#First create filter for the board
fltrjql = "project = " + prjkey + " ORDER BY RANK"
filter = authed_jira.create_filter(prjname, "Board for all the issues of the project", fltrjql)

#Create board using filter
boardname = prjname + " Issues"
authed_jira.create_board(boardname, prjkey, preset='scrum', location_type='project', location_id=prjkey)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3754,10 +3754,8 @@ class JIRA(object):
             for p in project_ids.split(','):
                 ids.append(self.project(p).id)
             project_ids = ','.join(ids)
-       
         if location_id is not None:
             location_id = self.project(location_id).id
-
         payload['name'] = name
         if isinstance(project_ids, string_types):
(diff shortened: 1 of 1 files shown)


hdost:
We have a test case. This change this should not be breaking existing use cases. Potentially there is a need to change the documentation on the `create_board` method. 
https://github.com/pycontribs/jira/blob/0ee722693c7a24f54ad93339cb139fd20ecc94ff/tests/tests.py#L1270

PR #715, opened 2018-11-01
Fix User regular expression to proper identify the User resource.

pricco:
Today (Nov 1st) JIRA API changed the self link.
They start using `accountId` instead of `username`.

This PR is to fix the regular expression (`resource_class_map`) to identify the User resource.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -1054,7 +1054,7 @@ resource_class_map = {
     r'securitylevel/[^/]+$': SecurityLevel,
     r'status/[^/]+$': Status,
     r'statuscategory/[^/]+$': StatusCategory,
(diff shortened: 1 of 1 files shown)


hdost:
I believe this may be a cloud only change, you may need to add a switch for that.

pricco:
@hdost I added the `username` again to the re.

pricco:
🔥
Urgent patch for production projects:

```
from jira.resources import resource_class_map
resource_class_map[r'user\?accountId.+$'] = resource_class_map[r'user\?username.+$']
```

ssbarnea:
Nice trick to use the emoticon, it worked as the email did standout. I only hope not to see everyone starting to use them making the inbox a xmas tree.

PR #716, opened 2018-11-04
 Added support for returning of generator

maciejs:


ssbarnea:
I am wondering if it would not be better to make ResultList a generator instead. If possible this would be preferable because it would allow us to even switch the default behavior in the future without breaking existing code.

maciejs:
Sounds reasonable. This could be implemented as next step.

hdost:
@ssbarnea is your thought that this change would be required instead of the one here?

ssbarnea:
Mainly I would propose updating this in such way that it would not need a new parameter. ResultList should be a generator.

Making pytho-api more complex will only make it harder to use, test and refactor.

As is python it does not have to be either one or another, it can be both.

jira/client.py
```diff
@@ -374,6 +377,7 @@ def __init__(self,
                  proxies=None,
                  timeout=None,
                  auth=None,
+                 generator=False,
```
ssbarnea: Please make ResultList act as a generator instead of adding a new parameter and different return values.

Issue #721, opened 2018-11-14
Add a method to modify a version that has already been created


Rumbles:
**Is your feature request related to a problem? Please describe.**
I want to start using this tool to manage a deployment process, but one step involves creating a version, and another step would be to mark that version as released if the build is sucessful. Currently I see no way of modifying a version once it has been created.

**Describe the solution you'd like**
Add a method to modify a version

**Describe alternatives you've considered**
I thought of calling the API directly, or deleting the version and then creating the new version with the right status, but neither of these would be great in my opinion

**Additional context**
I have already prepared a fork that would fix this issue, PR to follow

I don't think the way I have done it is necessarily the cleanest way to do it, maybe someone could point me in the right direction if you're not happy with it.

Rumbles:
@ssbarnea has asked for more information, I have been asked to create a workflow that the significant part of this involves 2 steps:

- Each evening when we complete an automatic build we create a new release, this release will be used the next day to track any work, as we need to put a version number on JIRA tickets.
 - Each evening we build a new release with all the commits from that day, if the build is successful the version is marked as released. We then need to create a new version for the following days work.

I was looking at using this tool to do these steps, but while I could see a way to do this in the JIRA API, there was only a way to create a new release in this python JIRA project, no way to modify a version that already exists (that I could see in the documentation).

I need a way to change a release from Unreleased to Released or visa versa.

I have added that functionality with my pull request.

If you need any more specific details, please let me know what info you need, thanks!

mhienle:
Am I missing something here? To me it looks like this functionality is already there. When you call `create_version`, you get the `Version` object (see [here](https://github.com/pycontribs/jira/blob/master/jira/client.py)) which has an `update` method (see [here](https://github.com/pycontribs/jira/blob/master/jira/resources.py)).

Rumbles:
If it's there @mhienle I couldn't see anything for it in the documentation

Do you know where it is documented?

I saw the update method in the code, but I don't see how you use it.

mhienle:
I thought of something like
```python
version = jira.create_version(...)  # https://jira.readthedocs.io/en/master/api.html#jira.JIRA.create_version
...  # do stuff
version.update(released=True)  # not documented
```
In the first line, you could also pass `released=True` or `released=False` to the function `create_version`, so I guess passing `released=True` to `version.update` should work (though I haven't tried).

Not sure if there should be more documentation. Also, the test for `update_version` could show a bit more examples.

[...]

PR #722, opened 2018-11-14
added a way to modify a version once it has been created

Rumbles:
Fixes the issue #721 I just opened

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2811,17 +2811,31 @@ class JIRA(object):
             data['releaseDate'] = releaseDate
         if startDate is not None:
             data['startDate'] = startDate
-
-        versions = self.project_versions(project)
-        for each in versions:
-            if str(each) == name:
-                version = each.id
-        url = self._get_url('version/' + version)
+        version = self.get_version(name, project)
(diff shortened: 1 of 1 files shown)


Rumbles:
@ssbarnea I'm not sure what you mean, the bug has the following text:

Is your feature request related to a problem? Please describe.
I want to start using this tool to manage a deployment process, but one step involves creating a version, and another step would be to mark that version as released if the build is sucessful. Currently I see no way of modifying a version once it has been created.

Describe the solution you'd like
Add a method to modify a version

Describe alternatives you've considered
I thought of calling the API directly, or deleting the version and then creating the new version with the right status, but neither of these would be great in my opinion

Additional context
I have already prepared a fork that would fix this issue, PR to follow

I don't think the way I have done it is necessarily the cleanest way to do it, maybe someone could point me in the right direction if you're not happy with it.

Rumbles:
As per issue, the functionality is there, just not documented

Rumbles:
Sure, I can look in to that, but I don't see any documentation for things in the resources file in readthedocs, can you point me to them? I searched for a string included in the docstring of the update method, but I got no hits in readthedocs

Issue #528, opened 2018-03-06
Cookie-based and Basic auth with password to be deprecated - start using basic authentication with an API token, OAuth, or Atlassian Connect


mansilladev:
See: https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-basic-auth-and-cookie-based-auth/

As of March 15, 2018, in accordance to the [Atlassian REST API Policy](https://developer.atlassian.com/platform/marketplace/atlassian-rest-api-policy/), support for passwords in HTTP basic auth requests && cookie-based authentication will no longer be supported on or after September 15, 2018.

Instead of passwords, you should start using [API tokens](https://confluence.atlassian.com/cloud/api-tokens-938839638.html).

ssbarnea:
Are these the same as https://github.com/pycontribs/jira/blob/427309b9e7b9e13149e1d39d962940364da03129/jira/client.py#L274 or not?

If not, the answer is that we will support them when someone will raise a PR that works with them. ;)

Tagman:
This is a different kind of token. It seems like these are only available in the cloud solution. Our local JIRA server (7.9.xx) doesn't have a site to generate a special API token.

This API token should replace the password in a request. So maybe no further changes to work with them.

Tagman:
This is a different kind of token. It seems like these are only available in the cloud solution. Our local JIRA server (7.9.xx) doesn't have a site to generate a special API token.

This API token should replace the password in a request. So maybe no further changes are needed to work with them.

hdost:
Based on: https://developer.atlassian.com/cloud/jira/platform/jira-rest-api-basic-authentication/

Like @Tagman said, Anyone using these personal tokens should just used the basic authentication method and nothing should be different.

This is a non-issue.

anyayunli:
API token is still favorable to basic_auth in most cases.

PR #734, opened 2018-12-05
updating documentation to tell user to use token

Almenon:
see https://github.com/pycontribs/jira/issues/528 passwords are no longer supported for HTTP BASIC Auth

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -57,9 +57,11 @@ the retrieved session cookie will be used in future requests. Upon cookie expira
 HTTP BASIC
 ^^^^^^^^^^
 
-Pass a tuple of (username, password) to the ``basic_auth`` constructor argument::
+Pass a tuple of (username, token) to the ``basic_auth`` constructor argument::
 
-    auth_jira = JIRA(basic_auth=('username', 'password'))
+    auth_jira = JIRA(basic_auth=('username', 'token'))
+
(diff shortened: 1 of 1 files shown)


Almenon:
This is just a documentation change but I should note that the code should be changed as well.

asqui:
This documentation change looks to be geared towards JIRA Cloud; we should be conscious that there are also JIRA Server users who I think are not subject to this constraint (yet?) and for whom the token creation URL is not correct.

docs/examples.rst
```diff
@@ -57,9 +57,11 @@ the retrieved session cookie will be used in future requests. Upon cookie expira
 HTTP BASIC
 ^^^^^^^^^^
 
-Pass a tuple of (username, password) to the ``basic_auth`` constructor argument::
+Pass a tuple of (username, token) to the ``basic_auth`` constructor argument::
 
-    auth_jira = JIRA(basic_auth=('username', 'password'))
+    auth_jira = JIRA(basic_auth=('username', 'token'))
+
+You can get a token by going to https://id.atlassian.com/manage/api-tokens and clicking "Create API Token"
```
dbaxa: This is for Jira Cloud only.

ssbarnea:
Closing as concernes were not addressed. If we change documentation to mention the token we need to do it in a way that still keeps the password option in.

PR #747, opened 2019-01-17
Removed mention of pycrypto

pgacv2:
Per issue #619.


hdost:
Cherry-piced as #800 on master

PR #758, opened 2019-02-11
Fix flake

cybit:
Fixes so that lintian hopefully works again.
Had to ignore W504 as W503 and W504 are mutually exclusive . W504 seems to be the accepted as the better solution. (https://github.com/PyCQA/pycodestyle/issues/792)
reload(sys) will always fail in newer python versions and there seems to be no solution that works for new and old. (The if clause around it seems to be there for a reason)
To fix W605 I had to replace escaping the dot with putting it into a character class expression. (https://stackoverflow.com/questions/13989640/regular-expression-to-match-a-dot)

### DIFF
--- a/jira/jirashell.py
+++ b/jira/jirashell.py
@@ -265,7 +265,7 @@ def main():
         # would report utf-8
         # see: https://stackoverflow.com/a/23847316/99834
         stdin, stdout, stderr = sys.stdin, sys.stdout, sys.stderr
-        reload(sys)
+        reload(sys)  # noqa
         sys.stdin, sys.stdout, sys.stderr = stdin, stdout, stderr
         sys.setdefaultencoding(os.environ.get('PYTHONIOENCODING', 'utf-8'))
 
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -318,13 +318,13 @@ class Resource(object):
                     if m:
                         user = m.groups()[0]
                     else:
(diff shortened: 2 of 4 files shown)


cybit:
@ssbarnea Could you please rephrase your comment? I do not understand it. 
My goal was only to fix the existing linting errors in travis so that other pull requests can pass.

* reload(sys) needs to be ignored as long as python2.X is being supported. 

* My change on W504 has already been added to master. So it is no longer part of the pull request.
* My change on W605 has already been fixed in master and is no longer part of the pull request.

ssbarnea:
@cybit I didn't see any more errors in travis and I merged 3-4 PRs today. Are you sure they are still failing due to this or maybe they are just outdated.

Linting is run locally with `tox -e lint`, same thing is run by travis and I expect both to give same results.

cybit:
flake8 --config setup.cfg still faills with

```
./jira/jirashell.py:268:9: F821 undefined name 'reload'
./tests/tests.py:269:21: F841 local variable 'e' is assigned to but never used
./tests/tests.py:280:21: F841 local variable 'e' is assigned to but never used
1     F821 undefined name 'reload'
2     F841 local variable 'e' is assigned to but never used

```

using python 3.7 and flake8 3.7.5 
But my other pull request passes on travis now.

So. Sorry for wasting your time because I used the wrong python version. :( I'm closing here.

PR #762, opened 2019-03-06
no-check -> nocheck

yampelo:
Prevents warnings on the server:

```
Use of the 'nocheck' value for X-Atlassian-Token has been deprecated since rest 3.0.0. Please use a value of 'no-check' instead.
```

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -349,7 +349,7 @@ class JIRA(object):
             # 'Accept': 'application/json',  # default for REST
             # 'Pragma': 'no-cache',
             # 'Expires': 'Thu, 01 Jan 1970 00:00:00 GMT'
-            'X-Atlassian-Token': 'no-check'}}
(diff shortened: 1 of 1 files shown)


dbaxa:
For the record it should remain as `no-check`.

ssbarnea:
I am curious why it should remain like this? Based on original message the change was making sense to me but I am glad I asked. 

Only reason for asking is because I want to keep the knowledge logged in this bug, and avoid seeing someone else proposing same change again in few months from now.

dbaxa:
@ssbarnea as per the message ->
> Use of the 'nocheck' value for X-Atlassian-Token has been deprecated since rest 3.0.0. Please use a value of 'no-check' instead.

We (Atlassian) deprecated the 'nocheck' value for `X-Atlassian-Token` and you should use `no-check`.

PR #763, opened 2019-03-07
docs: made StringIO example compatible with py3/py3

wesinator:


ssbarnea:
Python 2.x support was dropped, no further releases will be made for it.

wesinator:
> Python 2.x support was dropped, no further releases will be made for it.

@ssbarnea the master example at https://github.com/pycontribs/jira/blob/master/docs/examples.rst#attachments is for Python 2

Issue #26, opened 2015-02-05
Conflicting documentation sites


FrancescoRizzi:
I think this is due to the ongoing transfer of the project from one site to another, but figured I'd bring it up...

So, https://jira-python.readthedocs.org/en/latest/
And http://jira.readthedocs.org/en/latest/

List different version numbers (and different URLs for the source), but use the same package name for pip (pip install jira).. Of course, one of the two is wrong (the older jira-python if I had to guess?)
HTH

ssbarnea:
Sorted, now the old link was removed. Still, it will take some time for google indexing to bring the new one on the top.

PR #769, opened 2019-03-18
fix link to readthedocs in comment

tomviner:
This seems to be a final broken link from https://github.com/pycontribs/jira/issues/26

https://jira-python.readthedocs.org doesn't work anymore, so reference https://jira.readthedocs.io


tomviner:
I've merged in master, let's see if that helps.

Issue #697, opened 2018-09-26
GDPR: JIRA API changes regarding users


petroswork:
According to [this](https://developer.atlassian.com/cloud/jira/platform/api-changes-for-user-privacy-announcement/?_ga=2.175026080.366259731.1533136712-1276219155.1503476605), the API regarding users will change, several attributes will be removed, one will be added and some may change value/behaviour.

This will require changes to documentation and possibly the methods that depend on the `key` or other attribute to be deprecated. Some likely changes to the code are in class `Resource`, `_READABLE_IDS` and the `add_user` method.

hdost:
Sounds like many of the changes are around user attributes, most of which are checked for in the API. I think the action for this to make the work lighter is to break it down for by their tracking issues and then further subdivide in the case of the issue being to large.

petroswork:
Also, `current_user()` seems to be affected, as it reads the `X-AUSERNAME` header, while it should be using the `X-AACCOUNTID` header instead.

CSimpiFoN:
I suppose these changes were not implemented yet, but the deadline is closing in very fast. Also it seems some changes were already made on Atlassian's side that brake the authentication of this module in some cases.

ssbarnea:
One can always [open a bounty](https://www.bountysource.com/) to try to motivate people ;)

myuwono:
@ssbarnea @CSimpiFoN I've opened #780 for the authentication module.

PR #781, opened 2019-04-26
Support API changes due to GDPR

auxsvr:
Fixes most of https://github.com/pycontribs/jira/issues/697

The changes are backwards incompatible, which is a consequence of the changes to the JIRA API.
Changes to documentation are required.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3323,24 +3323,24 @@ class JIRA(object):
             logging.error(ioe)
         return None
 
-    def current_accountid(self):
-        """Returns the accountId of the current user.
+    def current_user(self):
+        """Returns the username or accountId of the current user, depending
+        on what the server supports, in order of preference.
 
         :rtype: str
         """
-        if not hasattr(self, '_serverInfo') or 'accountid' not in self._serverInfo:
+        if not hasattr(self, '_serverInfo') or\
+           not [i for i in self._serverInfo if i in ('accountid', 'username')]:
 
             url = self._get_url('serverInfo')
             r = self._session.get(url, headers=self._options['headers'])
 
             r_json = json_loads(r)
-            if 'x-aaccountid' in r.headers:
-                r_json['accountid'] = r.headers['x-aaccountid']
-            else:
-                r_json['accountid'] = None
+            r_json['accountid'] = r.headers.get('x-aaccountid')
+            r_json['username'] = r.headers.get('x-ausername')
             self._serverInfo = r_json
             # del r_json['self']  # this isn't really an addressable resource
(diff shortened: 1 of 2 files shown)


auxsvr:
Leaving it implies we return `accountId` where the consumers expect `username`, which will break them, too. Changing the function name will fail explicitly, so that the old code will have to be updated.

jira/client.py
```diff
@@ -3322,24 +3323,24 @@ def backup_download(self, filename=None):
             logging.error(ioe)
         return None
 
-    def current_user(self):
-        """Returns the username of the current user.
+    def current_accountid(self):
```
ssbarnea: Keep function name and *if* possible make it return same resul with both old/new servers. If not possible change its description and document that on newer instances it will receive the accountid.

If both return types can be returned there are two approaches: either to add extra function (current_accountid) or to add optional parameter that mentions that you want accountid instead of username to be returned.

Later if we are forced we can change the meaning of the default parameter.

The main idea is that if someone that code that uses the newer library (or newer jira server) should not break. Is not always possible but we should do out best.

ssbarnea: method name is current_user not currect_username which means that "user" can be anything, either username, userid, a disctionary that describes the user, an user object.

If the name of the function would have being different it would be total different story.

jira/client.py
```diff
@@ -2723,11 +2722,11 @@ def delete_user_avatar(self, username, avatar):
         url = self._get_url('user/avatar/' + avatar)
         return self._session.delete(url, params=params)
 
-    def search_users(self, user, startAt=0, maxResults=50, includeActive=True, includeInactive=False):
+    def search_users(self, query, startAt=0, maxResults=50, includeActive=True, includeInactive=False, accountId=None):
```
auxsvr: This requires backward incompatible change in the signature, as the problematic parameter is positional.

jira/client.py
```diff
@@ -2742,9 +2741,11 @@ def search_users(self, user, startAt=0, maxResults=50, includeActive=True, inclu
         :rtype: ResultList
         """
         params = {
-            'username': user,
+            'query': query,
             'includeActive': includeActive,
             'includeInactive': includeInactive}
+        if query:
```
auxsvr: The rationale here is that empty `query` implies that `accountId` is used for the query.

jira/client.py
auxsvr: You're right in this case, but I don't think it's possible for the rest of the changes.

ssbarnea:
Closing it as recent changes addressed GDPR changes.

PR #786, opened 2019-05-04
Feat: allow saving password in OS keychain/keyring

### LINKED ISSUE #784: Save API tokens / passwords in keyring
**Is your feature request related to a problem? Please describe.**
[Atlassian is deprecating][1] clear password authentication via basic HTTP (at least for cloud accounts), in favor of "personal API tokens". This means that users will need to hard-code their tokens in an alias or config file, in plain text.

[1]: https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-basic-auth-and-cookie-based-auth/

**Describe the solution you'd like**
However, there is a better way, using keyring/keychain provider in the OS. This is already used by the [go-jira][2] client.

[2]: https://github.com/Netflix-Skunkworks/go-jira#keyring-password-source

**Describe alternatives you've considered**
I'm not aware of alternatives that work, besides maybe using a 3rd party password manager to save API tokens, which is ok I guess, though not very user-friendly - not everyone uses 1password/keepass, do they?

**Additional context**
I see the python [keyring] package, is well maintained and supports a couple of python 3 versions (3.5-3.7). Python 2.7 support was recently dropped, and if required, keyring==18 can be used.

asfaltboy:
resolves #784

Considering one thing we could change, is to move the keyring lookup logic inside the JIRA class, so that it's available to library users as well as shell users ...

@ssbarnea how do you feel about this?

### DIFF
--- a/jira/jirashell.py
+++ b/jira/jirashell.py
@@ -7,25 +7,26 @@ over HTTP BASIC or Kerberos.
 """
 
 from __future__ import print_function
+
+import argparse
+import os
+import sys
+import webbrowser
+from getpass import getpass
+
+import requests
+from oauthlib.oauth1 import SIGNATURE_RSA
+from requests_oauthlib import OAuth1
+from six.moves import input
+from six.moves.urllib.parse import parse_qsl
+
+from jira import JIRA, __version__
+
 try:
     import configparser
 except ImportError:
     from six.moves import configparser
 
-from six.moves import input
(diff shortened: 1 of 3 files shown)


asfaltboy:
Hi @pycontribs  , shall we merge make a new release 2.0.1 that includes this feature ^ as well as dozens of commits from 2019 ?

PR #795, opened 2019-05-28
doc-fix: Specify correct parameter name of 'destination'.

akumria:
This was a bit confusing to read / understand initially.


akumria:
It is non-obvious to me why Travis CI failed.

PR #801, opened 2019-06-03
Add access to resources.py for users.

rustyscottweber:
Users don't have access to resources, but they should.
For example (Checking types):
```
if isinstance(o, jira.resources.IssueType):
                return {}
```


rustyscottweber:
My mistake, the file is put into the install bundle and can be imported separately.

Issue #780, opened 2019-04-18
Cloud: Basic auth and cookie based auth are deprecated


myuwono:
**Is your feature request related to a problem? Please describe.**

Basic authentication with passwords and cookie-based authentication no longer work since effectively 14 April 2019. Api token is the supported method going forward for basic auth, i.e. `basic_auth=(username, api_token)`

The problem manifest itself in #778 when users attempt to authenticate with cookie based auth `auth=(username, password) ` (which kept on failing) and basic auth with password `basic_auth=(username, password)`

**Describe the solution you'd like**

It's important to:
- reflect this in the documentation
- prevent jira client from being instantiated when deprecated authentication method is selected for `deployment_type="Cloud"` by throwing an appropriate exception on client creation. 

**Additional context**

- https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-basic-auth-and-cookie-based-auth/ 
- https://ecosystem.atlassian.net/browse/ACJIRA-1466
- https://ecosystem.atlassian.net/browse/ACJIRA-1466

Sefiris:
Putting in my 2 cents, it would be nice if the documentation at least reflected this after I just spent an hour trying to get basic auth working for a POC only to find this issue with the context explaining exactly why it was erroring on
`	response text = {"errorMessages":["You do not have the permission to see the specified issue.","Login Required"],"errors":{}}
`

PR #806, opened 2019-06-17
#780 Update documentation for password/cookie based auth deprecation

myuwono:
Basic authentication with passwords and cookie-based authentication are no longer supported in JIRA cloud. This needs to be reflected in the documentation as appropriate.


myuwono:
@aimestereo fixed! Would be good to see this merged

PR #807, opened 2019-06-24
new: add expand option to JIRA.projects

Addono:
**Issue:** When querying all projects using `JIRA.projects()` there's no option to use the expand feature from the API, for example to be guaranteed to also retrieve the description of a project.

**Solution:** Add an optional parameter to the `JIRA.projects()` which allows specifying an expand parameter, just like e.g. `JIRA.user(id, expand=None)` or `JIRA.group(id, expand=None)`.

**Alternatives:** Earlier I was iterating through all retrieved groups and then individually calling `JIRA.project(project)`, however this invokes an additional API call for every project, which causes major performance degradation when working with large projects.* 

**Changes:** The changed made in this PR are BC and seem too simple to fail - actually it feels too simple so if there's a reason why this wasn't yet implemented then please tell. One concern I have with this PR is that for testing purposes it assumes that the description field is expandable, hence the tests depend on the fact that each project has to have a description, which introduces an implicit dependency.

*In addition, even for this call the same expandable options exist, which seems to me that there's no guarantee that these values are actually present and hence it's more "luck" based.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2228,13 +2228,17 @@ class JIRA(object):
 
     # Projects
 
-    def projects(self):
+    def projects(self, expand=None):
         """Get a list of project Resources from the server visible to the current authenticated user.
 
-        :rtype: List[Project]
+        :param expand: extra information to fetch inside each project
 
+        :type expand: str
         """
-        r_json = self._get_json("project")
+        params = {}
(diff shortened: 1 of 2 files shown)


adehad:
FYI @Addono I've chosen not to reopen this PR due to #865

PR #813, opened 2019-07-01
added support to disable getting application links

mnadeem-acc:
I added support to disable getting application links. It was a quick fix and backwards compatibility is kept.


adehad:
Would be great if you could merge the latest master, we no longer use `pre-commit`  so hopefully your diff will be pretty small and easy to review

ssbarnea:
Closing due to lack of updates.

Issue #701, opened 2018-10-09
Send search request as POST for large queries


zgoda-mobica:
**Is your feature request related to a problem? Please describe.**
Jira API allows search requests to be performed either as GET or POST, with POST suggested for queries that can't be encoded as query params. I have a query with ~60k clauses, ~60 bytes each after encoding as query param.

**Describe the solution you'd like**
For queries that can't be encoded as query params use POST request with JSON payload instead of GET.

ssbarnea:
I totally agree with your proposal. Now you only need to make a PR to address it.

zgoda:
For over a year I used only POST requests for search in Jira and did not notice any issues. I think this should be default - it's hard to know what's the url size boundary and in theory this depends on servlet container implementation (can't say as I use only Tomcat).

PR #844, opened 2019-10-02
Fix for #701

vVv-AA:
Use POST for searching issues.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -628,6 +628,7 @@ class JIRA(object):
         maxResults=50,
         params=None,
         base=JIRA_BASE_URL,
+        verb="GET"
     ):
         """Fetch pages.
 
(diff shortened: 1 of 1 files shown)


vVv-AA:
@ssbarnea @hdost Can you please review this?

jira/client.py
```diff
@@ -3150,12 +3153,13 @@ def _get_json(self, path, params=None, base=JIRA_BASE_URL):
...
         :type base: Optional[str]
-
+        :param verb: HTTP verb to use for the request.
+        :type verb: str
         :rtype: Union[Dict[str, Any], List[Dict[str, str]]]
 
         """
         url = self._get_url(path, base)
-        r = self._session.get(url, params=params)
+        r = getattr(self._session, verb.lower())(url, params=params)
```
dbaxa: While this works, I feel like this code should probably follow a different method/design in order to support using `post` to search for issues.

vVv-AA: I tried to keep it in line with the usage here https://github.com/pycontribs/jira/blob/91461fd736cd9a37cc136e07402c6f0b1e60170b/jira/resilientsession.py#L143

I kept the search as POST by default to avoid user from having to manually having to handle JQLs that are too long.

What approach do you suggest?

PR #854, opened 2019-10-14
Added example to create new fix versions programmatically

alexa-ai:
More examples will be useful to effectively make use of this wonderful repo. This example helps in creating fix versions programmatically, instead of adding fix version for each project manually.


alexa-ai:
@ssbarnea : Many PRs are stuck at Travis failure and get marked as stale after few days. I don't want the same to happen to this PR. I can't figure out why Travis CI fails. "Details" log is not helping. I have added new .py file for additional example, and it should not cause anything to fail. Is Travis CI failing for all PRs or only for few PRs?

alexa-ai:
Thanks @ssbarnea . Let me fix it.

PR #858, opened 2019-10-18
Adds support for different payload types on the user endpoint.

jfboismenu:
The JIRA REST API supports not only `username` but also `key` and `accountId` when resolving a user. This branch adds support for those two parameter types.

The default behaviour does not change, as the username will still be used, but it allows to change the argument type.

This helps better support a local JIRA support, where the `key` for a user is not always equal to the `username`.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2159,13 +2159,13 @@ class JIRA(object):
 
     # Users
 
-    def user(self, id, expand=None):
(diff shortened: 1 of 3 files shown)


jfboismenu:
Sorry. I meant to submit this PR against our own fork of the repo first for code review internally.

PR #859, opened 2019-10-20
added example to create fix version

alexa-ai:


alexa-ai:
@ssbarnea : Travis CI is failing while trying to authenticate JIRA. I added authentication code as used in basic_use.py example program. Still, it does not go thru. Is modifying docs/examples.rst the best way to add examples? 

FYI - I followed your advice and made sure pylint and black are passed.

PR #862, opened 2019-10-23
Modified document to add example to use Atlassian API token

alexa-ai:
Added example to authenticate using Atlassian API token.


alexa-ai:
giving up on this

PR #863, opened 2019-10-27
make Jira issues hashable

senarclens:
This allows them to be used in sets or as dictionary keys.

I did not add this to the Resource class as that is inherited by multiple other classes
with differend __eq__ implementations.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -599,6 +599,10 @@ class Issue(Resource):
         """Comparison method."""
         return other is not None and self.id == other.id
 
+    def __hash__(self):
+        """Allow using Issue in sets and as key for dictionaries."""
+        return hash(self.id)
+
 
 class Comment(Resource):
     """An issue comment."""


senarclens:
Anyone up for quickly reviewing this PR?

senarclens:
@lan17 thx for the feedback. I can adjust to self.key instead if you prefer - both are unique to the issue so it didn't matter when I implemented this (neither in performance nor in functionality).

Wrt the test I wanted to implement one but didn't get the other tests to run and since the change is trivial didn't add one. Plz let me know if anything is required from my side. Thx.

lastthyme (approved):

jira/resources.py
```diff
@@ -599,6 +599,10 @@ def __eq__(self, other):
         """Comparison method."""
         return other is not None and self.id == other.id
 
+    def __hash__(self):
+        """Allow using Issue in sets and as key for dictionaries."""
+        return hash(self.id)
```
dbaxa, line 604: Hmm, should we consider hashing based upon an issue's `permalink` ?

adehad:
Rfer #1051

PR #866, opened 2019-11-07
Retry timeout

mafrosis:
Two things here:

- A small bug fix when the JIRA API endpoint is unavaiable and an exception is raised
- A feature change to set the exponential retry delay to be based off the timeout value. I think it’s reasonable that if users want a low timeout, they will also want a low retry time. (I was waiting multiple minutes for Jira client to decide the API is unavailable).


mafrosis:
Apologies, just noticed tests are failing. I will investigate and update the P/R.

Some feedback on if you will merge the patch would be most welcome.

PR #878, opened 2019-11-27
Documentation: HTTP Basic with remote server

holingpoon:
The documentation was not clear on how to pass authentication info during the initialization of the JIRA object. Adding an example for a particular use case.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -104,7 +104,7 @@ To enable Kerberos auth, set ``kerberos=True``::
 To pass additional options to Kerberos auth use dict ``kerberos_options``, e.g.::
 
     auth_jira = JIRA(kerberos=True, kerberos_options={'mutual_authentication': 'DISABLED'})
-    
+
 .. _jirashell-label:
 
 Issues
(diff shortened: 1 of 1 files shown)


lan17:
It's probably better to just add a link to pydocs on authentication: https://jira.readthedocs.io/en/master/examples.html#authentication

holingpoon:
Good idea! I'll find a spot under that section.

docs/examples.rst
```diff
@@ -101,6 +101,13 @@ To pass additional options to Kerberos auth use dict ``kerberos_options``, e.g.:
 
     auth_jira = JIRA(kerberos=True, kerberos_options={'mutual_authentication': 'DISABLED'})
 
+Authenticating on a remote server
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You may also pass multiple arguments to initialize the JIRA object. For example, to log into a remote Jira server with basic authentication::
+
+   jira = JIRA('https://jira.atlassian.com', basic_auth('username', 'password'))
```
lan17: Basic auth example is a few sections above, by the way on line 52. 

You can modify to also pass in the server, but otherwise, there is no need for duplicate basic auth examples.

PR #894, opened 2020-01-23
Allow setting page_size with unlimited paging

eagleonhill:
By default, it will use the whatever the first maxResults that server returns.
This can be far less than what the server is capable of, and fetching in small batch is very inefficient.


studioj:
@eagleonhill please merge in changes from master. 
could it be you change also requires a documentation update?

adehad:
closing as probably will be superseded by #1394

PR #901, opened 2020-02-14
update_sprint return using Sprint() method

AlfredThompson:
Would require using 'raw' attribute rather than the previously returned 'dict' object


studioj:
@ssbarnea why can't I reopen? Is there a limit? :-)

PR #902, opened 2020-02-14
Allow _get_user_accountid to "find" Unassigned and Automatic "users".

truthbyron:
Return None, -1, and "-1" back from _get_user_accountid to allow `assign_issue` to do that as well.

This seems related to the work being done to handle Jira's migration away from `key` and to `accountId` everywhere. I'm happy to adjust this if needed or add tests if that would be helpful for this change.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1560,8 +1560,10 @@ class JIRA(object):
         return self._get_json("issue/createmeta", params)
 
     def _get_user_accountid(self, user):
-        """Internal method for translating an user to an accountId."""
+        """Internal method for translating an user to an accountId. Return None and -1 unchanged."""
         try:
+            if user is None or user in (-1, "-1"):
+                return user
             accountId = self.search_users(user, maxResults=1)[0].accountId
         except Exception as e:
             raise JIRAError(e)


studioj:
@truthbyron fix conflicts pls
it would be nice if you add a test for this

jira/client.py
```diff
@@ -1560,8 +1560,10 @@ def createmeta(
         return self._get_json("issue/createmeta", params)
 
     def _get_user_accountid(self, user):
-        """Internal method for translating an user to an accountId."""
+        """Internal method for translating an user to an accountId. Return None and -1 unchanged."""
         try:
+            if user is None or user in (-1, "-1"):
```
adehad, line 1565: I think the approach is good, my only suggestion might be to:

```suggestion
            if user in (None, -1, "-1"):
```

but not major

truthbyron: Hey thanks for the feedback! I will try to address it shortly and resolve the conflicts.

adehad:
@truthbyron if you allow maintainers to commit to your fork I should be able to push updates there, I've resolved the conflicts locally so can push them whenever.

adehad:
Refer #1152

Issue #896, opened 2020-01-25
Python Jira needs CI maintenance help


ssbarnea:
That is a call for help on fixing JIRA CI project so we can test and merge change. I am personally against merging any change that does not pass CI but reality is that CI degrated considerably in the last years and cannot be fixed in one hour or two, probably not even an entire weekend.
  
If you care about its future, there are two ways you can pick

- Raise PRs that are addressing the CI issues and not by removing or disabling tests
- Use the sponsor option to show your support, maybe you can convince me to go back to jira instead of the other projects
- Become a maintainer yourself. Well that is the best and if you can deal with first bullet, you have my approval.

truthbyron:
Hi @ssbarnea, I believe this is failing because of [Deprecation of basic authentication with passwords for Jira and Confluence APIs](https://confluence.atlassian.com/cloud/deprecation-of-basic-authentication-with-passwords-for-jira-and-confluence-apis-972355348.html), given the [Travis runs of my PR](https://travis-ci.com/pycontribs/jira/jobs/287461558):

```
E           jira.exceptions.JIRAError: JiraError HTTP 403 url: https://pycontribs.atlassian.net/rest/auth/1/session
E           	text: Basic auth with password is not allowed on this instance
```

I was able to create tie a test Jira account to the testing cloud instance, create an API token, and authenticate using a basic auth tuple of `(email, token)` and get a totally different set (and way more) errors, so I think that method works.

Is there any way we could get a dedicated token generated for the test and admin user and change to using email address instead of username for this block https://github.com/pycontribs/jira/blob/5c27f6ddafffc6110be1db4749fa67025852bcb6/tests/tests.py#L116-L134

I'm not sure how technically able to help I am, but I'm happy to research more/help if you have time  to give any direction (otherwise I'll keep poking around as I can). Thanks for this project and for your time.

csullivannet:
@ssbarnea I've run this as well, setting the environment variables for `CI_JIRA_ADMIN`, `CI_JIRA_ADMIN_PASSWORD`, `CI_JIRA_USER`, `CI_JIRA_USER_PASSWORD` to my own username (my e-mail) and my API token. I was able to successfully authenticate when running `python -m tox` locally.

Further, I disabled the check here to see whether I'm the admin:
https://github.com/pycontribs/jira/blob/5c27f6ddafffc6110be1db4749fa67025852bcb6/tests/tests.py#L163-L165
[...]

Issue #871, opened 2019-11-15
Travis CI is failing *not due* to PR issues


tubaman:
Travis CI is failing even on a trivial PR: https://github.com/pycontribs/jira/pull/870

**To Reproduce**
1. Create a trivial PR
2. See Travis CI fail

**Expected behavior**
Travis CI should only fail if there are issues with a PR

ssbarnea:
That is true but someone needs to address this, and I am currently not using the library.

PS. By addressing it, I do not mean to disable CI testing, i mean finding a way to make it work. Atlassian test instance is working.

tubaman:
What's the Atlassian test instance?

Addono:
Been looking a bit into this today, the problem is that "Encrypted environmental variables" are used, which are not exposed on PR branches, as you will see when you inspect the build logs for a PR:
```
Encrypted environment variables have been removed for security reasons.
See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions
```

And indeed, comparing the exposed variables in a PR:
![image](https://user-images.githubusercontent.com/15435678/69344336-85910b80-0c6f-11ea-9167-2cffd63617ef.png)

Versus one by the maintainer:
![image](https://user-images.githubusercontent.com/15435678/69344406-a22d4380-0c6f-11ea-8ff7-ba4b59a217d7.png)


Solution wise, Travis' [documentation](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions) states this:
> If your build relies on encrypted variables to run, for instance to run Selenium tests with BrowserStack or Sauce Labs, your build needs to take this into account. You won’t be able to run these tests for pull requests from external contributors.
>
> To work around this, **restrict these tests only to situations where the environment variables are available, or disable them for pull requests entirely,** as shown in the following example:
> ...

An alternative solution would be to see if we can for example spin up a JIRA test instance as part of the CI/CD, then use that for testing, and afterwards tear it down. I am currently doing something like that in a Jenkins pipeline, which works fine, although it does take a minute or so for JIRA server to boot. But given that the current Travis pipeline already takes a couple of minutes this shouldn't be too much of an issue.

What could also work would be to make the CI user credentials public. Given that the Jira instance is solely used for testing is there not much damage one can do with these accounts (e.g. delete projects), although that would only be an option if it can be configured such that it has enough rights to pass the tests, but not enough to escalate its rights. For which I do not know if this is feasible.

[...]

PR #909, opened 2020-02-28
896 - GDPR Update and CI fixes

csullivannet:
Intended as the resolution to https://github.com/pycontribs/jira/issues/896

A number of changes were made to the Jira API server side that break the implementation of this library. Primarily, `username`/`name` is generally no longer a valid query or response. This has had the effect of breaking a number of the tests as well as how we authenticate to our test instance in the first place.

I've made the decision to continue passing the credentials in plaintext, due to limitations noted first by @Addono here: https://github.com/pycontribs/jira/issues/871#issuecomment-557099698
Simply put, the tests will fail on PRs as they would not have access to the credentials necessary for the test instance otherwise. Other solutions are possible, but other than the potential for vandalism I didn't see the need for these credentials to be secret.

However, I have moved these variables out of the code and into the environment created by Travis. This gives people the option to define their own set, should they desire. The tests will fail fast if otherwise not set, as discussed here: https://github.com/pycontribs/jira/issues/896#issuecomment-592405680

A couple of other fixes have been made, where fields have either been removed or made mandatory:
- The `key` field has been removed the `myself` response
- `permissions` is now a mandatory field for the `mypermissions` request

Finally, some fixes have been applied to pass linting.

This PR is not likely to be a complete fix for all GDPR related issues. Merely, it fixes the code where tested, or the tests themselves, such that CI will be green again.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -944,7 +944,7 @@ class JIRA(object):
         name,
         project,
         description=None,
-        leadUserName=None,
+        leadAccountId=None,
         assigneeType=None,
         isAssigneeTypeValid=False,
     ):
@@ -956,8 +956,8 @@ class JIRA(object):
         :type project: str
         :param description: a description of the component
         :type description: str
-        :param leadUserName: the username of the user responsible for this component
-        :type leadUserName: Optional[str]
+        :param leadAccountId: the username of the user responsible for this component
+        :type leadAccountId: Optional[str]
         :param assigneeType: see the ComponentBean.AssigneeType class for valid values
         :type assigneeType: Optional[str]
         :param isAssigneeTypeValid: boolean specifying whether the assignee type is acceptable (Default: False)
@@ -971,8 +971,8 @@ class JIRA(object):
         }
         if description is not None:
             data["description"] = description
-        if leadUserName is not None:
-            data["leadUserName"] = leadUserName
+        if leadAccountId is not None:
+            data["leadAccountId"] = leadAccountId
         if assigneeType is not None:
             data["assigneeType"] = assigneeType
 
@@ -1569,7 +1569,7 @@ class JIRA(object):
 
     # non-resource
     @translate_resource_args
-    def assign_issue(self, issue, assignee):
+    def assign_issue(self, issue, assignee_id):
         """Assign an issue to a user. None will set it to unassigned. -1 will set it to Automatic.
 
         :param issue: the issue ID or key to assign
@@ -1585,7 +1585,7 @@ class JIRA(object):
             + str(issue)
             + "/assignee"
         )
-        payload = {"accountId": self._get_user_accountid(assignee)}
+        payload = {"accountId": assignee_id}
         # 'key' and 'name' are deprecated in favor of accountId
         r = self._session.put(url, data=json.dumps(payload))
         raise_on_error(r)
--- a/jira/client.py
+++ b/jira/client.py
@@ -2176,7 +2176,8 @@ class JIRA(object):
 
     # non-resource
     def my_permissions(
-        self, projectKey=None, projectId=None, issueKey=None, issueId=None
+        self, projectKey=None, projectId=None, issueKey=None, issueId=None, * ,
+        permissions
     ):
         """Get a dict of all available permissions on the server.
 
@@ -2188,6 +2189,8 @@ class JIRA(object):
         :type issueKey: Optional[str]
         :param issueId: limit returned permissions to the specified issue
         :type issueId: Optional[str]
+        :param permissions: a CSV list of permission keys to get permissions for
+        :type permissions: str
         :rtype: Dict[str, Dict[str, Dict[str, str]]]
         """
         params = {}
@@ -2199,6 +2202,7 @@ class JIRA(object):
             params["issueKey"] = issueKey
         if issueId is not None:
             params["issueId"] = issueId
+        params["permissions"] = permissions
         return self._get_json("mypermissions", params=params)
 
     # Priorities
--- a/jira/client.py
+++ b/jira/client.py
@@ -481,9 +481,8 @@ class JIRA(object):
             self._create_kerberos_session(timeout, kerberos_options=kerberos_options)
         elif auth:
             self._create_cookie_auth(auth, timeout)
-            validate = (
-                True
-            )  # always log in for cookie based auth, as we need a first request to be logged in
+            # always log in for cookie based auth, as we need a first request to be logged in
+            validate = True
         else:
             verify = self._options["verify"]
             self._session = ResilientSession(timeout=timeout)
@@ -2176,7 +2175,12 @@ class JIRA(object):
 
     # non-resource
     def my_permissions(
-        self, projectKey=None, projectId=None, issueKey=None, issueId=None, * ,
+        self,
+        projectKey=None,
+        projectId=None,
+        issueKey=None,
+        issueId=None,
+        *,
         permissions
     ):
         """Get a dict of all available permissions on the server.
--- a/jira/client.py
+++ b/jira/client.py
@@ -3438,9 +3438,9 @@ class JIRA(object):
             logging.error(ioe)
         return None
 
-    def current_user(self, field="key"):
-        """Returns the username or emailAddress of the current user. For anonymous
-        users it will return a value that evaluates as False.
+    def current_user(self, field="accountId"):
+        """Returns information about the current user, including but not limited
+        to their display name, e-mail, accountId.
 
         :rtype: str
         """
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -544,15 +544,15 @@ class ComponentTests(unittest.TestCase):
             name,
             self.project_b,
             description="stand by!",
-            leadUserName=self.jira.current_user(),
+            leadAccountId=self.jira.current_user(),
         )
         name = "renamed-" + name
         component.update(
-            name=name, description="It is done.", leadUserName=self.jira.current_user()
+            name=name, description="It is done.", leadAccountId=self.jira.current_user()
         )
         self.assertEqual(component.name, name)
         self.assertEqual(component.description, "It is done.")
-        self.assertEqual(component.lead.name, self.jira.current_user())
+        self.assertEqual(component.lead.accountId, self.jira.current_user())
         component.delete()
 
     def test_4_delete(self):
@@ -639,7 +639,7 @@ class FilterTests(unittest.TestCase):
             name=name, description="just some new test filter", jql=jql, favourite=False
         )
         self.assertEqual(myfilter.name, name)
-        self.assertEqual(myfilter.owner.name, self.test_manager.user_admin.name)
+        self.assertEqual(myfilter.owner.accountId, self.test_manager.user_admin.accountId)
         myfilter.delete()
 
     def test_favourite_filters(self):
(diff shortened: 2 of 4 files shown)


Addono:
> However, I have moved these variables out of the code and into the environment created by Travis. This gives people the option to define their own set, should they desire. ...

How would people be supposed to overwrite these values? I tried to overwrite them through the "Environment variables from repository settings", however it seems that the ordering is such that the one in the `.travis.yaml` take precedence.

[This](https://travis-ci.com/Addono/jira/jobs/293295621#L161) pipeline passes, while the environment variables are overwritten to use non-existent credentials for the admin user:
image

.travis.yml
```diff
@@ -74,3 +74,8 @@ env:
   global:
     - secure: "pGQGM5YmHvOgaKihOyzb3k6bdqLQnZQ2OXO9QrfXlXwtop3zvZQi80Q+01l230x2psDWlwvqWTknAjAt1w463fYXPwpoSvKVCsLSSbjrf2l56nrDqnoir+n0CBy288+eIdaGEfzcxDiuULeKjlg08zrqjcjLjW0bDbBrlTXsb5U="
     - PIP_DISABLE_PIP_VERSION_CHECK=1
+    - CI_JIRA_URL="https://pycontribs.atlassian.net"
```
ssbarnea: Do not put credentails in plain text.

csullivannet:
> > However, I have moved these variables out of the code and into the environment created by Travis. This gives people the option to define their own set, should they desire. ...
> 
> How would people be supposed to overwrite these values? I tried to overwrite them through the "Environment variables from repository settings", however it seems that the ordering is such that the one in the `.travis.yaml` take precedence.
> 
> [This](https://travis-ci.com/Addono/jira/jobs/293295621#L161) pipeline passes, while the environment variables are overwritten to use non-existent credentials for the admin user:
> image

My intention was that these variables are read from the environment only as an intentional choice. You can run the tests against another server locally, but not via CI. For a PR to be passable, I would imagine the tests _must_be performed against the same server so that we don't have to chase down config ghosts.

.travis.yml
csullivannet: Unfortunately this is a limitation of Travis. I've added all the variables to the [Travis settings](https://travis-ci.com/pycontribs/jira/settings) and dropped the env auth creds commit. This causes the build to fail on a `KeyError` as they are not passed to my fork for the PR.

This is by design from Travis CI:
https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

If we expect external contributors to fork the project and have their PRs run these same tests, we necessarily must expose the credentials to these users.

Addono:
> My intention was that these variables are read from the environment only as an intentional choice. You can run the tests against another server locally, but not via CI. For a PR to be passable, I would imagine the tests _must_be performed against the same server so that we don't have to chase down config ghosts.

Good point, I agree that it's more important that PR MR are forced to be explicit about config changes.

tests/tests.py
```diff
@@ -105,33 +105,18 @@ def __init__(self):
...
+                "CI_JIRA_USER",
+                "CI_JIRA_USER_PASSWORD",
+            ]
 
-            if "CI_JIRA_USER_PASSWORD" in os.environ:
-                self.CI_JIRA_USER_PASSWORD = os.environ["CI_JIRA_USER_PASSWORD"]
-            else:
-                self.CI_JIRA_USER_PASSWORD = "sd4s3dgec5fhg4tfsds3434"
+            for var in env_auth_vars:
+                setattr(self, var, os.environ[var])
```
Addono: Don't really like this way of getting the environment variables. It's not really clear where the `self.CI_JIRA_ADMIN`. Took me a while before I figured out that the magic happened here, as I started reading somewhere halfway and my IDE couldn't tell me where they came from.

jira/client.py
```diff
@@ -3438,9 +3446,9 @@ def backup_download(self, filename=None):
             logging.error(ioe)
         return None
 
-    def current_user(self, field="key"):
-        """Returns the username or emailAddress of the current user. For anonymous
-        users it will return a value that evaluates as False.
+    def current_user(self, field="accountId"):
```
Addono: This change is not backwards compatible and will break non-cloud versions of Jira.

Addono: Hmm, I noticed that similar BC breaking changes have already been committed to master (e.g. a4e6b1b2). Not too fond of this, but I guess restoring the BC could better be left for a seperate PR.

csullivannet: The CI tests are only against the cloud version, which seems to be the priority for this project. Given the limited resources available to maintain it, I would tend to agree with breaking older on-prem versions in favour of keeping track with the cloud and maintaining BC on a best-effort basis.

The GDPR compliance changes also represent breaking changes within a major version of the API. This represents a fair amount of work to restore functionality to the library as more things get deprecated and dropped. As far as I can tell it also has caused a fair amount of divergence between on-prem and cloud versions, increasing the difficulty of maintenance.

With these factors in mind I'd prefer to prioritize getting fixes in to bring CI back to green and raise issues that arise afterwards, so that the project can gain momentum again. With CI green, it should be easier to put forward smaller PRs that are easier to merge to fix BC piece-by-piece.

tests/tests.py
```diff
@@ -160,7 +145,8 @@ def __init__(self):
                         logging=False,
                         max_retries=self.max_retries,
                     )
-            if not self.jira_admin.current_user():
+
+            if not self.jira_admin.current_user("displayName") == "CI Admin":
```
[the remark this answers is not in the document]
csullivannet: I'll find a value that the local tests and cloud tests can agree on that exists for the admin user or otherwise make changes to the tests so that they agree.

tests/tests.py
csullivannet: Would if be preferable to go back to declaring them line-by-line (but without having a fallback `if`/`else` clause- i.e. they must still be set by the environment)?

.travis.yml
Addono: One option would be to obfuscate them, e.g. store them encrypted together with the password. Security through obfuscation doesn't make it safer, but at least it requires someone to put a bit of effort in to get them and makes it harder for bots to scrape them.

tests/tests.py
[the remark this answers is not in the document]
Addono: It might not be such a big issue, there is much more to be configured before an instance spinned up with `atlas-run-standalone` works.

Shouldn't this check be more like "check if the current user is logged in ([optional] with any admin account)", rather than "check if the current user is logged in with an account with exact name XYZ"?

tests/tests.py
Addono: Seems like the way to go to me.

jira/client.py
Addono: I agree that getting the CI green again should be a top priority. Over the weekend I was working on getting the CI to run against a Jira instance spinned up as part of the pipeline, as to get rid of the credential management issue and have more reproducible builds. So in that effort fixing CI and maintaining BC align.

Anyway, this was just the first BC breaking change I ran into. Shouldn't be considered a blocker for this PR.

tests/tests.py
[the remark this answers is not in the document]
csullivannet: From my reading of the original test it was to check whether the call to `myself` works, which shows you are currently logged in. To be honest at this point I don't remember why exactly I used `displayName`, but I do recall I was having issues with getting this test to pass until I checked the username.

So in terms of the original intention of the test, I agree with you. It would be better for this test to pass as long as it returns anything, although I'm assuming it throws an error or otherwise not `200` if you're not logged in. I'll revisit this test.

.travis.yml
csullivannet: @ssbarnea Would Addono's proposal be an acceptable solution? I'm not certain there's any real workaround to this as it's a hard limitation of Travis. If having the credentials to the cloud test instance in the code (even obfuscated) are completely unacceptable then our efforts might be better redirected towards provisioning local test instances as part of the CI.

iramisvalentincapco:
Hello @Addono and @ssbarnea I was wondering if this PR has met the criteria for approval yet? Given the large number of changes to the JIRA API to be GDPR-compliant this would be crucial for usage of this package lib. 

I understand that having proper passing CI checks is critical for re-enabling PR approvals for this project. My understanding is that this PR would resolve the CI issues that seem to be blocking the  17 other PRs that are sitting in a blocked state.

Addono:
Hi @iramisvalentincapco, the goal of this PR is indeed to get the CI to work again. However, work on that hasn't completed - see the Travis CI pipelines which is still failing.

The reason why stems from the fact that to get the CI to pass again three issues needs to be resolved:
1. Fixing the codebase and test-suite to be compatible with the latest version of Jira. As no fixes are merged, the drifting API specs cause the current test suite to fail. 
2. Provisioning of a Jira instance to test against. Previously, a public Jira Cloud instance was used. However, the credentials to this instance can only be used for branches inside the repository. As giving credentials to this Jira instance to CI runs against PRs from untrusted sources would risk having them exposed.
3. Updating the CI to make use of the changed made to accommodate (1) and (2).

I have seen that a lot of people have been trying to fix 1, 2, 3 or a combination of them. My own experience is that solving all issues concurrently is troublesome, as it's too much to "just" do it all in a weekend and so many changes are required that you lose track of it yourself, hence for the reviewer of the PR it will be even harder to fathom.

iramisvalentincapco:
@Addono @ssbarnea It seems like @csullivannet has addressed #2 via this comment https://github.com/pycontribs/jira/pull/909#discussion_r389665691 is the proposed solution acceptable? 

If so, would it then be acceptable for the main repo to have a "fix/ci" branch or something along those lines so that fixes for #1, #2 and #3 can be more safely merged without impacting the master/release branch?

ssbarnea:
Yes, good idea. Create a new branch called `fix/ci` and this will be used to bring it back to green as a collaborative effort.

Addono:
First contribution to `fix/ci` is ready: Managed to get a containerized version of Jira Software running. However, no other branch than `master` exists, so I cannot create a PR.
https://github.com/Addono/jira/tree/feature/ci-run-jira-in-sidecar-container

The source code for the `jira-run-standalone` container can be found here:
https://github.com/Addono/docker-jira-software-standalone

Barely any of the tests pass, so debugging that would be the next step. Probably some can be fixed by cherry-picking from all open PRs.

ssbarnea:
Now there is a new branch https://github.com/pycontribs/jira/tree/fix/ci -- feel free to feed it.

Addono:
@ssbarnea cool! Created PR #942

.travis.yml
iramisvalentincapco: @ssbarnea Hi, just bumping this into view since it looks like this has been sitting for a bit and it appears to be crucial for this PR.

iramisvalentincapco:
@csullivannet I think after taking in the changes from #942 and changing the merge branch to the new fix/ci branch this PR may be able to get some more movement on the other standing issues with the PR.

Addono:
Yes, more work can be done. But having the CI now test against a Jira Server version instead of Jira Cloud shows that recently (traced it back somewhere around `master~10`) changes were introduced which completely break support for Jira Server.

I think it's important for QA that both Server and Cloud are tested, as their APIs are growing more and more apart. Therefore, having something work on one version does not guarantee that it works on the other. At the very least we can know and document that if a new version would be released now is not suitable for usage with Server environments.

Sadly, this also means that the CI issue isn't completely solved, as now the tests do not run against a Cloud environment. I would suggest reintroducing the tests which run against the Jira Server environment using the encrypted secrets similarly as what was used before. The downside is that this means that this won't be available on runs which originate from PRs. However, this will still be able to run on pushes to `fix/ci`, which would prevent faulty commits from reaching `master`. Re-adding this part of the CI should be pretty trivial and at least get full coverage whilst fixing all the broken tests.

dsm1212:
Ah, sorry I submitted issue #962 about this. Is there any way to get this moving again?

Addono:
> Ah, sorry I submitted issue #962 about this. Is there any way to get this moving again?

Yes, but it requires finding someone who's willing to spend time on getting all tests to pass again for Jira Server instances, from then on it will be possible to merge fixes (assuming they are not Jira Cloud specific).

For Jira Cloud specific fixes it would require finding a way on how to run the CI against a Jira Cloud instance. Which has been troublesome to setup, as because of [this](https://github.com/pycontribs/jira/pull/909#discussion_r388208799) concern of exposing the credentials to the Jira Cloud instance used to run the CI against.

SpoonOfDoom:
> For Jira Cloud specific fixes it would require finding a way on how to run the CI against a Jira Cloud instance. Which has been troublesome to setup, as because of [this](https://github.com/pycontribs/jira/pull/909#discussion_r388208799) concern of exposing the credentials to the Jira Cloud instance used to run the CI against.

Are there any ideas on how to do this? If I understand the discussion above correctly, we're talking about limitations of Travis that cannot be circumvented without having the credentials in the source?

Also, what is the worst case scenario if credentials to the test instance get leaked? As far as I'm aware, the cloud instance usage limits are hard, and there are no costs automatically incurred for traffic, used storage space, etc., is this correct?
I just want to understand the risks we are talking about, because on the surface the danger looks to be pretty mild, to be perfectly honest - especially if, as suggested, the credentials are stored in an encrypted manner to prevent scraping by bots.
I know it's not exactly best practice, but if that's what makes or breaks this project, it might be a valid pragmatic approach.

PR #910, opened 2020-03-02
Fix find_transitionid_by_name

buscoma:
You were passing the whole dict, not only the transition["name"] value.


buscoma:
Doesent mind, I was wrong

PR #912, opened 2020-03-05
📝 Make quickstart example Python 3 compliant

Addono:
Also update the return values to match the current values.


ssbarnea:
Already fixed in master.

Issue #913, opened 2020-03-11
Integrate jira's new REST API endpoint for labels


mikeboiko:
**Is your feature request related to a problem? Please describe.**
I use issue labels in my workflow. One of the features that I would like to use is to get a list of all labels that are currently used in a project.

**Describe the solution you'd like**
Jira recently implemented this feature in their REST API. See links below:
https://ecosystem.atlassian.net/browse/ACJIRA-788
https://developer.atlassian.com/cloud/jira/platform/rest/v3/?utm_source=%2Fcloud%2Fjira%2Fplatform%2Frest%2F&utm_medium=302#api-rest-api-3-label-get

Thanks so much for the great jira python tool!

PR #917, opened 2020-03-20
Added ability to get full list of labels

svermeulen:
For issue #913

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3967,7 +3967,6 @@ class JIRA:
 
         url = self._options["server"] + "/rest/api/3/label"
 
-        print(f"using url = {url}")
         r = self._session.get(url)
         data = json_loads(r)["values"]
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -3963,6 +3963,17 @@ class JIRA:
         return data
 
     @lru_cache(maxsize=None)
+    def labels(self):
+
+        url = self._options["server"] + "/rest/api/3/label"
+
+        print(f"using url = {url}")
+        r = self._session.get(url)
+        data = json_loads(r)["values"]
+
+        return data
+
+    @lru_cache(maxsize=None)
     def avatars(self, entity="project"):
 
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -3544,6 +3544,17 @@ def projectcategories(self):
 
         return data
 
+    @lru_cache(maxsize=None)
+    def labels(self):
+
+        url = self._options["server"] + "/rest/api/3/label"
```
adehad, line 3550: Would transition this to use `self._get_json()`

ssbarnea (changes requested):
Using print does not seem like ready for review code.

studioj:
@svermeulen merge in master and resolve conflicts please

studioj:
@svermeulen any idea if this will be picked up again? => our CI system is working again properly

svermeulen:
Ok I rebased my branch to latest main, and removed the print statement.  Though I don't know how to address @adehad's comment

adehad (changes requested):
This endpoint is only available on v3 of the API for Jira Cloud, this needs to be handled correctly.

`self._options["rest_api_version"]` has a value that can be checked against. 

Or we can instead check the `self._is_cloud` value and if true, use a URL with the v3 url?
That would require extension of the `_get_json()` and `_get_url()` methods to accept a version override.


```python
if not self._is_cloud:
    raise NotImplementedError("Only available on Jira Cloud instances.")
if str(self._options["rest_api_version"]) != "3":
    raise NotImplementedError("Needs v3 of the REST API Version.")

```

jira/client.py
adehad: added an example in https://github.com/pycontribs/jira/pull/917#discussion_r739844741

studioj:
no response after +1yr after requesting changes => close

adehad:
@dvolosnykh I suggested some changes that were not addressed at the time, if you are able to add them in a new PR that would be ideal.

A test would be greatly appreciated too.

PR #927, opened 2020-04-16
Make user search GDPR compliant

julenpardo:
The `username` field is deprecated and Jira is gradually removing it
from the cloud instances. This is the second time such changes break
our integrations; until now, our workaround consisted of first searching
for the user based on the email, and then use the account id from the
response for the other requests. But now we cannot search anymore for
users based on the email, and we need to use the `query` field. Even if
we just pass the same exact value we passed to `username`.

I can imagine there're several places in the code that would require of
being changed for being completely GDPR compliant, but I have no time
at the moment to fix all of them :(

Thanks for your work!

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2977,7 +2977,7 @@ class JIRA(object):
               If maxResults evaluates as False, it will try to get all items in batches.
             includeActive (bool): If true, then active users are included in the results. (Default: True)
             includeInactive (bool): If true, then inactive users are included in the results. (Default: False)
-            query (optional str): Search term. It can just be the email.
+            query (Optional[str]): Search term. It can just be the email.
 
         Returns:
             ResultList[User]
--- a/jira/client.py
+++ b/jira/client.py
@@ -2987,12 +2987,10 @@ class JIRA(object):
 
         params = {
             "username": user,
+            "query": query,
             "includeActive": includeActive,
             "includeInactive": includeInactive,
         }
-
-        if query:
-            params["query"] = query
 
         return self._fetch_pages(User, None, "user/search", startAt, maxResults, params)
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -2964,8 +2964,11 @@ class JIRA(object):
         maxResults: int = 50,
         includeActive: bool = True,
         includeInactive: bool = False,
+        query: Optional[str] = None,
     ) -> ResultList[User]:
         """Get a list of user Resources that match the specified search string.
+        "username" query parameter is deprecated; the expected parameter now is "query", which can just be the full
+        email again. But the first parameter is kept for backwards compatibility.
 
         Args:
             user (str): a string to match usernames, name or email against.
@@ -2974,15 +2977,25 @@ class JIRA(object):
               If maxResults evaluates as False, it will try to get all items in batches.
             includeActive (bool): If true, then active users are included in the results. (Default: True)
             includeInactive (bool): If true, then inactive users are included in the results. (Default: False)
+            query (optional str): Search term. It can just be the email.
 
         Returns:
             ResultList[User]
         """
+        if not user and not query:
+            raise ValueError("Either 'user' or 'query' arguments must be specified.")
+
         params = {
             "username": user,
             "includeActive": includeActive,
             "includeInactive": includeInactive,
         }
+
+        if user:
+            params["username"] = user
+        else:
+            params["query"] = query
+
         return self._fetch_pages(User, None, "user/search", startAt, maxResults, params)
 
     def search_allowed_users_for_issue(
--- a/jira/client.py
+++ b/jira/client.py
@@ -2991,9 +2991,7 @@ class JIRA(object):
             "includeInactive": includeInactive,
         }
 
-        if user:
-            params["username"] = user
-        else:
+        if query:
             params["query"] = query
 
         return self._fetch_pages(User, None, "user/search", startAt, maxResults, params)
--- a/jira/client.py
+++ b/jira/client.py
@@ -2967,7 +2967,7 @@ class JIRA(object):
         query: Optional[str] = None,
     ) -> ResultList[User]:
         """Get a list of user Resources that match the specified search string.
-        "username" query parameter is deprecated; the expected parameter now is "query", which can just be the full
+        "username" query parameter is deprecated in Jira Cloud; the expected parameter now is "query", which can just be the full
         email again. But the first parameter is kept for backwards compatibility.
 
         Args:
--- a/jira/client.py
+++ b/jira/client.py
@@ -2959,7 +2959,7 @@ class JIRA(object):
 
     def search_users(
         self,
-        user: str,
+        user: Optional[str] = None,
         startAt: int = 0,
         maxResults: int = 50,
         includeActive: bool = True,
@@ -2971,7 +2971,7 @@ class JIRA(object):
         email again. But the first parameter is kept for backwards compatibility.
 
         Args:
-            user (str): a string to match usernames, name or email against.
+            user (optional str): a string to match usernames, name or email against.
             startAt (int): index of the first user to return.
             maxResults (int): maximum number of users to return.
               If maxResults evaluates as False, it will try to get all items in batches.
--- a/jira/client.py
+++ b/jira/client.py
@@ -2971,7 +2971,7 @@ class JIRA(object):
         email again. But the first parameter is kept for backwards compatibility.
 
         Args:
-            user (optional str): a string to match usernames, name or email against.
+            user (Optional[str]): a string to match usernames, name or email against.
             startAt (int): index of the first user to return.
             maxResults (int): maximum number of users to return.
               If maxResults evaluates as False, it will try to get all items in batches.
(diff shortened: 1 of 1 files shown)


eliaschneider:
LGTM, A great pull request!

jira/client.py
```diff
@@ -2828,15 +2830,26 @@ def search_users(
...
+        :param query: Search term. It can just be the email.
+        :type query: str
 
 
         :rtype: ResultList
         """
+        if not user and not query:
+            raise ValueError("Either 'user' or 'query' arguments must be specified.")
+
         params = {
```
ngenovictor, line 2842: why not retain existing changes and update params
`
params = {
            "query": user,
            ......
     }
`

julenpardo: @ngenovictor if just using the `query` param is backwards compatible, I'm fine with that too

pallavrustogi: I ran tests and query looks to be backward compatible. it works with api version 2 and 3 both.

jira/client.py
```diff
@@ -2828,15 +2830,26 @@ def search_users(
...
             "username": user,
             "includeActive": includeActive,
             "includeInactive": includeInactive,
         }
+
+        if user:
+            params["username"] = user
```
darlev91, line 2849: @julenpardo Thanks for your efforts on this stuff! We are looking forward for it will be merged! :heart: 
Just one question about changes, isn't it redundant to add this check if the `username` parameter is already in `params` dictionary ?
https://github.com/pycontribs/jira/blob/b5ef5288516117cae6702c54a6d066fe271afc08/jira/client.py#L2843

julenpardo: Hey @darlev91, thanks for your feedback! You're absolutely right, thanks for the catch! I'll push the change asap

julenpardo: @darlev91 I've already made the changes, thanks again for the feedback!

darlev91: @julienadriano Thanks a lot!
What's about `query` parameter ? It is compatible with both JIRA API versions. I think it's better to keep things simple

julenpardo: @darlev91 is it? This is actually what @ngenovictor pointed out above. So we can just use the `query` parameter right away and keep it backwards compatible?

I did it this way just to be 100% sure that it was compatible with both versions, as I can only test it with the latest one, so I would not be sure about replacing the `username` parameter. But if that's the case, then I can just apply @ngenovictor's suggestion

darlev91: @julienadriano yes, if checking JIRA API docs, it says that `query` parameter is required (can contain not only `emailAddress` as a value, and `displayName` too), but it shouldn't replace `username` because it is separate one:
[Link on JIRA API v3](https://developer.atlassian.com/cloud/jira/platform/rest/v3/?utm_source=%2Fcloud%2Fjira%2Fplatform%2Frest%2F&utm_medium=302#api-rest-api-3-user-search-get)
[Link on JIRA API v2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/#api-rest-api-2-user-search-get)
I've checked it already and it really works fine for both API versions, so I think it will be better to add it directly in parameters like separate parameter that should be send on API.

darlev91: @julienadriano @ngenovictor I think this point of view make sense in this case:
```
def search_users(self, query, username=None, startAt=0, maxResults=50, includeActive=True, includeInactive=False):
     ...
     params = {
            "query": query,
            "username": username,
            "includeActive": includeActive,
            "includeInactive": includeInactive,
        }
        ...
```
In this case the client code shouldn't be broken if it uses `displayName` or `emailAddress` as a `query` 
What do you think about it, guys ?

julenpardo: @darlev91 Makes sense to me, I've already pushed the changes

studioj:
likely a duplicate of #940

studioj:
@julenpardo  please black jira/client.py and merge in changes from master

julenpardo:
@studioj Thanks for the feedback, I've already pushed the changes, is there something I should do now to trigger the checks?

studioj:
im not too aquianted with jira api but as stated in above review => why add query, if you can query for user ?

julenpardo:
@studioj I probably did it that way to still keep the other parameter for backward compatibility, as the change appeared to be a canary release affecting only some servers

I'm fine with either approach; I did it this way back then in a rush to fix broken production systems :)

The thing is that we pinned the changes made on my branch as dependency, and it's been that way since the PR was open, so I'm aware that change may break some stuff. Let me test this on Monday

adehad:
Checking the documentation of 

Self-hosted:
https://docs.atlassian.com/software/jira/docs/api/REST/8.13.6/#api/2/user-findUsers
`query` is not present

Cloud:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-user-search/#api-rest-api-2-user-search-get
`query` and `username` are present 

It seems that `query` is only really available on Cloud. That would explain the current failures seen on #940 - no users are returned according to the logs:
image

That also explains why this PR passes those same checks.

It seems like the API doesn't have a problem with passing in incorrect parameters, as discussed in other threads on this PR. So I'm happy approving this PR.

jira/client.py
```diff
@@ -2807,9 +2807,17 @@ def delete_user_avatar(self, username, avatar):
...
+        self,
+        user=None,
+        startAt=0,
+        maxResults=50,
+        includeActive=True,
+        includeInactive=False,
+        query=None,
     ):
         """Get a list of user Resources that match the specified search string.
+        "username" query parameter is deprecated; the expected parameter now is "query", which can just be the full
```
adehad, line 2819: Refer to my previous comment. We should specify here that this only applied to "Cloud" instances, at least for the time being.

```suggestion
        "username" query parameter is deprecated in Jira Cloud; the expected parameter now is "query", which can just be the full
```

adehad (changes requested):

julenpardo:
Many thanks @adehad! I've already committed the suggestion and rebased the head branch with latest master.

jira/client.py
```diff
@@ -2959,30 +2959,39 @@ def delete_user_avatar(self, username: str, avatar: str):
...
         includeInactive: bool = False,
+        query: Optional[str] = None,
     ) -> ResultList[User]:
         """Get a list of user Resources that match the specified search string.
+        "username" query parameter is deprecated in Jira Cloud; the expected parameter now is "query", which can just be the full
+        email again. But the first parameter is kept for backwards compatibility.
 
         Args:
-            user (str): a string to match usernames, name or email against.
+            user (optional str): a string to match usernames, name or email against.
```
adehad, line 2974:
```suggestion
            user (Optional[str]): a string to match usernames, name or email against.
```

jira/client.py
```diff
@@ -2959,30 +2959,39 @@ def delete_user_avatar(self, username: str, avatar: str):
...
             startAt (int): index of the first user to return.
             maxResults (int): maximum number of users to return.
               If maxResults evaluates as False, it will try to get all items in batches.
             includeActive (bool): If true, then active users are included in the results. (Default: True)
             includeInactive (bool): If true, then inactive users are included in the results. (Default: False)
+            query (optional str): Search term. It can just be the email.
```
adehad, line 2980:
```suggestion
            query (Optional[str]): Search term. It can just be the email.
```

jira/client.py
```diff
@@ -2959,30 +2959,39 @@ def delete_user_avatar(self, username: str, avatar: str):
...
+        user: Optional[str] = None,
         startAt: int = 0,
         maxResults: int = 50,
         includeActive: bool = True,
         includeInactive: bool = False,
+        query: Optional[str] = None,
     ) -> ResultList[User]:
         """Get a list of user Resources that match the specified search string.
+        "username" query parameter is deprecated in Jira Cloud; the expected parameter now is "query", which can just be the full
+        email again. But the first parameter is kept for backwards compatibility.
```
adehad, line 2971:
```suggestion
        email again. But the "user" parameter is kept for backwards compatibility, i.e. Jira Server/Data Center.
```

adehad (changes requested):
just a few minor comments on the docstring and the typehint docstring format, glad you managed to get through the rebase alright, it was quite a hefty one !

julenpardo:
@adehad applied, thank you!

adehad (approved):

julenpardo:
Thanks @adehad for your patience and for approving, what's the process now to merge the PR?

adehad:
@julenpardo as I've approved I've requested a review from @ssbarnea who has the permissions for the merge.

We have a somewhat weekly cadence for merging stuff in, most happens on the weekend. But I'm sure this time it won't take a year !

julenpardo:
Many thanks guys!

adehad:
and thank you @julenpardo for your patience ! Hope you contribute in the future too !

lucasolinden:
When will this be included in a proper PIP release?

adehad:
FYI @lucasolinden (and others following) this is now included in the pre-release `pip install jira==3.1.0rc1` - expect a proper release soon.

ssbarnea:
@adehad I just wanted to mention that you have all my support to tag a full release whenever you think is ok. I liked the feedback I seen so far and once we have that release up it should be easier to make new ones w/o having to do pre-releases. If we release often, the need for a pre-release goes down. Thanks for reviving maintenance on jira library!

PR #928, opened 2020-04-17
ResilientSession: do not log request data on ConnectionError

Flupp:
The request data may contain secrets, e.g., a plaintext password when using basic auth. Therefore the request data shall not be logged.

In version 2.0.0, the basic auth login credentials were logged when a ConnectionError was triggered by an invalid SSL certificate. With current master this is not the case anymore, however, the problem might still occur under other circumstances.

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -94,6 +94,7 @@ class ResilientSession(Session):
 
     def __init__(self, timeout=None):
         self.max_retries = 3
+        self.max_retry_delay = 60
         self.timeout = timeout
         super(ResilientSession, self).__init__()
 
@@ -129,7 +130,7 @@ class ResilientSession(Session):
                 msg = "Atlassian's bug https://jira.atlassian.com/browse/JRA-41559"
 
         # Exponential backoff with full jitter.
-        delay = min(60, 10 * 2 ** counter) * random.random()
+        delay = min(self.max_retry_delay, 10 * 2 ** counter) * random.random()
         logging.warning(
             "Got recoverable error from %s %s, will retry [%s/%s] in %ss. Err: %s"
             % (request, url, counter, self.max_retries, delay, msg)
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -167,7 +167,8 @@ class ResilientSession(Session):
                 if response.status_code >= 200 and response.status_code <= 299:
                     return response
             except ConnectionError as e:
-                logging.warning(f"{e} while doing {verb.upper()} {url} [{kwargs}]")
+                logging.warning(f"{e} while doing {verb.upper()} {url}")
+
                 exception = e
             retry_number += 1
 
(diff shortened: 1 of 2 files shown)


studioj:
@Flupp would you be able to write a test for it? since CI is up and running ...

Flupp:
I tried coming up with a test for checking that `ResilientSession` writes no sensitive data to the log.

While doing that I noticed and fixed another problem; see ef184c742b002cc934cb232f5260905d00f324b0.

Flupp:
Addressed CI linting errors.

Flupp:
Rebased and resolved merge conflicts. The problem I fixed in ef184c742b002cc934cb232f5260905d00f324b0 was fixed by 6e513952c854acf67f78bce8c281c0546489a346 in the meantime.

I also silenced a false positive Codacy Static Code Analysis warning by renaming a variable. Some of the remaining warnings disagree with `black`.

tests/tests.py
```diff
@@ -2096,6 +2097,56 @@ def test_session_server_offline(self):
...
+        self.records = []
+
+    def emit(self, record):
+        self.records.append(record)
+
+    def reset(self):
+        self.records = []
+
+
+class ResilientSessionLoggingConfidentiality(unittest.TestCase):
```
studioj, line 2114: just a remark from my QA point of view => TestClasses should best start with Test ...
makes a better differentiator between helper classes and the real testclasses (im not usre if its done consistently in these tests, but its a good idea to evolve to that standard

Flupp: Makes sense; done. However, I used `Tests` as suffix since this seems to be the preferred style in `tests.py`.

Flupp:
Rebased and fixed merge conflicts introduced by d1f244c244021e45a3e86e217563cb7aeca01225.

Flupp:
Rebased and fixed timeout issue in test by introducing configurable maximal retry delay in b331f6a7107048663f0c56e592ba3cf554683200.

jira/resilientsession.py
```diff
@@ -129,7 +130,7 @@ def __recoverable(
                 msg = "Atlassian's bug https://jira.atlassian.com/browse/JRA-41559"
 
         # Exponential backoff with full jitter.
-        delay = min(60, 10 * 2 ** counter) * random.random()
+        delay = min(self.max_retry_delay, 10 * 2 ** counter) * random.random()
```
studioj, line 133: 👍

studioj:
the codacy fail can be ignored i think https://app.codacy.com/gh/pycontribs/jira/pullRequest?prid=7435641

Flupp:
Do I have to do something to ignore the Codacy warning (like adding some annotation in the code) or will you just silently ignore the warning?

studioj:
> Do I have to do something to ignore the Codacy warning (like adding some annotation in the code) or will you just silently ignore the warning?

@ssbarnea should "approve" it in the tool i think

studioj (approved):

adehad (approved):

ssbarnea:
FYI, I removed codacity, too much false positives.

PR #929, opened 2020-04-18
changed search_users to use query instead of username for gdpr compli…

bniemczyk:
…ance


bniemczyk:
fixes error when search_users is used on gdpr strict jira instance, including jira cloud

bniemczyk:
looked at other PRs and saw this is a dup.  closing.

PR #942, opened 2020-05-21
ci: run Jira standalone in a sidecar container

Addono:
Relates to #909


Addono:
The CI isn't green yet, because the tests do not pass. The purpose of this PR is not to fix those, instead it lets the CI run against a local version of Jira Software Server.

PR #957, opened 2020-07-31
Close the file descriptor for add_attachment

yen3:
If the attachment argument is string, the add_attachment function creates a file descriptor then forget to close it. The patch closes the file descriptor after the post action.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -853,6 +853,7 @@ class JIRA(object):
         close_attachment = False
         if isinstance(attachment, str):
             attachment: BufferedReader = open(attachment, "rb")  # type: ignore
+            attachment = cast(BufferedReader, attachment)
             close_attachment = True
         elif isinstance(attachment, BufferedReader) and attachment.mode != "rb":
             self.log.warning(
@@ -878,7 +879,7 @@ class JIRA(object):
                 )
             finally:
                 if close_attachment:
-                    cast(BufferedReader, attachment).close()
+                    attachment.close()
         else:
             method = "MultipartEncoder"
 
@@ -900,7 +901,7 @@ class JIRA(object):
                 )
             finally:
                 if close_attachment:
-                    cast(BufferedReader, attachment).close()
+                    attachment.close()
 
         js: Union[Dict[str, Any], List[Dict[str, Any]]] = json_loads(r)
         if not js or not isinstance(js, Iterable):
--- a/jira/client.py
+++ b/jira/client.py
@@ -849,8 +849,10 @@ class JIRA(object):
         Returns:
             Attachment
         """
+        close_attachment = False
         if isinstance(attachment, str):
             attachment: BufferedReader = open(attachment, "rb")  # type: ignore
+            close_attachment = True
         elif isinstance(attachment, BufferedReader) and attachment.mode != "rb":
             self.log.warning(
                 "%s was not opened in 'rb' mode, attaching file may fail."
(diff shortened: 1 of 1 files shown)


adehad:
I think you have addressed the problem correctly, but part of me wonders why this function is structured the way it is.

Probably a future pull request though, so don't worry about it now. \
I wonder if this function should only accept a string, and the context manager approach be used to open it the intended way (i.e. "rb" mode etc.), this would make the function a lot clearer.

But perhaps I am missing a use case for when the file would already be open ?

yen3:
Hi, @adehad Thanks for your reply.

https://github.com/pycontribs/jira/pull/957/files#diff-442d14b20d591afd0c3de830d2803e30d4816eedc708f9ad129316aa7e5a4a7cL856

The parameter document mentioned that the function could accept a file descriptor. Maybe I misunderstood something.

adehad:
> Hi, @adehad Thanks for your reply.
> 
> https://github.com/pycontribs/jira/pull/957/files#diff-442d14b20d591afd0c3de830d2803e30d4816eedc708f9ad129316aa7e5a4a7cL856
> 
> The parameter document mentioned that the function could accept a file descriptor. Maybe I misunderstood something.

Yep you are correct, i was just rambling to myself haha, sorry for the confusion.

jira/client.py
```diff
@@ -879,13 +881,19 @@ def add_attachment(self, issue, attachment, filename=None):
...
+            try:
+                r = self._session.post(
+                    url,
+                    files={"file": (fname, attachment, "application/octet-stream")},
+                    headers=CaseInsensitiveDict(
+                        {"content-type": None, "X-Atlassian-Token": "nocheck"}
+                    ),
+                )
+            except Exception as e:
+                raise e
```
studioj, line 893: if you raise e here it will not show the proper traceback

```suggestion
                raise
```

or plainly remove the except part

studioj, line 893: if you raise e here it will not show the proper traceback
```suggestion
            except Exception as e:
                raise
```
or better maybe even dont catch exceptions
```suggestion
```

studioj:
it would be nice if you would add a test uploading for example the README file as file descriptor and as string

studioj (approved):

adehad:
@yen3 don't suppose you could run `tox -e lint` over this. It seems like it is just the linting that is causing the checks to fail.

yen3:
@adehad @studioj @ssbarnea Thanks for the review. I try to rebase on the latest master branch so I push the change again. Please review it again. Thank you.

adehad:
@yen3 looks like there is still a minor linting problem, otherwise changes look good to me.

yen3:
@adehad Thanks for the review. I try to fix it. Please review it again.

jira/client.py
```diff
@@ -849,8 +850,10 @@ def add_attachment(
         Returns:
             Attachment
         """
+        close_attachment = False
         if isinstance(attachment, str):
             attachment: BufferedReader = open(attachment, "rb")  # type: ignore
+            close_attachment = True
```
adehad, line 856: Let's put the cast earlier over here:

```suggestion
            attachment: BufferedReader = open(attachment, "rb")
            attachment = cast(BufferedReader, attachment)
            close_attachment = True
```

Then you won't need to cast later on.

You can blame me for just ignoring that type warning, sorry about that !

adehad (changes requested):
added a minor comment on where we do the `cast()`

jira/client.py
yen3: It's a good idea. Thanks for your help.

adehad (approved):

adehad (changes requested):
oops, there's a cheeky black formatting error somewhere here

yen3:
@adehad  I run `black -l 88 client.py` to refine the code format.

adehad (approved):

ssbarnea (approved):

PR #958, opened 2020-08-07
Allow the adding of headers rather than just overwriting all headers

pradcliffe-ns:
In the current update of the options dict you either overwrite all default headers or none. This change allows the addition of headers or changing of existing headers without removing other default headers. That way the current `headers` entries in `DEFAULT_OPTIONS` will not need to be duplicated in code that uses the module and if the default headers set is updated in the future it will be used as expected.


studioj:
@pradcliffe-ns  please merge in the changes from master, solve the conflicts and maybe explain the added functionality improvement in the PR itself

pradcliffe-ns:
> @pradcliffe-ns please merge in the changes from master, solve the conflicts and maybe explain the added functionality improvement in the PR itself (some code sample would be nice)

@studioj I filed this 9 months ago and was trying to be helpful in adding functionality that was useful as explained above. I'm working on different projects now and I'm not sure I have time to add to this. Going back to try and produce example code and work out where in your project those belong is not trivial.

It's a pretty simple change that allows you to add headers without overwriting them all as I explained above, anything in DEFAULT_OPTIONS - https://github.com/pycontribs/jira/blob/b93cdb861412fd24fd833cf1305d8d6196711c4f/jira/client.py#L276 - that you need to change requires overwriting the whole set rather than just making one change. If you don't want that functionality then close the PR.

studioj:
I believe it would be smart to add a test verifying this behavior. But that's my QA person opinion :) in the end up to @ssbarnea  :-) i just try to offload some work of of him. 


@pradcliffe-ns
atm CI tests are failing because of black

pradcliffe-ns:
I have no idea what "because of black" means.

studioj:
> I have no idea what "because of black" means.

@pradcliffe-ns this repo uses the uncompromising code formatter "black" https://pypi.org/project/black/ which is rapidly becoming a standard as code formatting goes.

if you don't know it yet, it's a very interesting project.
=> have a look at their read the docs https://black.readthedocs.io/en/stable/

how to fix your code ...

`pip install black`
`black client.py` 
then save commit and push

ssbarnea:
Please assure that at least running `tox -e lint` is passing before uploading a change. That would run all the linters, including black. its output should be explanatory. If not fixed, we wil lclose the PR.

adehad:
Refer to #1085

PR #976, opened 2020-10-12
Update __init__.py

MikhailToropov:
Fix problem with add_attachment(issue, file). RuntimeError: dictionary keys changed during iteration for Python 3.8

### DIFF
--- a/jira/utils/__init__.py
+++ b/jira/utils/__init__.py
@@ -34,13 +34,15 @@ class CaseInsensitiveDict(dict):
     """
 
     def __init__(self, *args, **kw):
-        super(CaseInsensitiveDict, self).__init__(*args, **kw)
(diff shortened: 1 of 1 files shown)


studioj:
@MikhailToropov  please add a test to check for the fixing of the bug, merge latest CI changes from master in your branch and make sure CI passes before we start a review on this. Thank you

adehad:
I'm tempted to say we should be using the `CaseInsensitiveDict` from [`requests.structures`](https://github.com/psf/requests/blob/1466ad713cf84738cd28f1224a7ab4a19e50e361/requests/structures.py#L15) directly, rather than this implementation. We can then be confident it is well tested too?

adehad:
Refer #1084

PR #978, opened 2020-10-19
Correct lookup of issuetypes when calling create_issue

soerenbe:
Today I got the issue that there where 2 issuetypes with the same name in our Jira installation. The current implementation have 2 issues with this:

1.) It always picks the ID of the first matching name. This may result in a error during create_issue because the other issuetype ist not available in that project. In this case I now raise an Exception beause the request of create_issue is not clear.
2.) When providing a issuetype id `create_issue` not correctly uses this value.


adehad:
@soerenbe if you are able to merge the latest master and then address the errors raised during `tox -e lint`  we can hopefully get this merged !

soerenbe:
Thanks @adehad for having a look at this issue. I fixed the syntax error.

adehad (approved):

PR #940, opened 2020-05-14
GDPR Strict Mode Fix

AlexAbraham1:
Atlassian has changed their rest API and removed the `username` parameter. They have replaced it with `query` which takes a display name or an email address. In the future they might remove support for email addresses and only accept display names.

More info [HERE](https://community.atlassian.com/t5/Jira-questions/The-query-parameter-username-is-not-supported-in-GDPR-strict/qaq-p/1345106)

PR #979, opened 2020-10-21
Fix retrocompatibilty issue with User - in Jira Software API there is…

oscarsj:
… Fix retrocompatibilty issue with User - in Jira Software API there is no accountId, we need to use "key" to identify an user. With this patch the code is compatible both with Jira Cloud (using "accountId") and Jira Software (using "key")

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1567,6 +1567,17 @@ class JIRA(object):
             raise JIRAError(e)
         return accountId
 
+    def _get_user_key(self, user):
+        """Internal method for translating an user to a key.
+            Needed for backwards compatibilty with Jira Software API, not using
+            accountId
+        """
+        try:
+            key = self.search_users(user, maxResults=1)[0].key
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -1567,6 +1567,17 @@ def _get_user_accountid(self, user):
...
         return accountId
 
+    def _get_user_key(self, user):
+        """Internal method for translating an user to a key.
+            Needed for backwards compatibilty with Jira Software API, not using
+            accountId
+        """
+        try:
+            key = self.search_users(user, maxResults=1)[0].key
+        except Exception as e:
```
studioj, line 1577: would it be possible to specify the exception type you're catching here

studioj:
could it be its related to https://github.com/pycontribs/jira/pull/940 ?

adehad:
I think we may close this if this is handled better by #1053

studioj:
asume fixed by #1053 => if not please reopen

PR #987, opened 2020-12-09
Fixes #404 - Allow you to set the reporter as well, not just the assignee

### LINKED ISSUE #404: Cannot set the reporter of a ticket (automatically defaults to 'assignee')
In resources.py, around line 464, there are these two lines of code:

```
if field == 'assignee' or field == 'reporter':
     fields_dict['assignee'] = {'name': value}
```

Unfortunately this means there's no way to set the reporter to anything other than the currently authenticated script creating or updating an issue.  It always defaults to 'assignee'.  Can be easily fixed by doing something like:
```
fields_dict[field] = {'name': value}
```

tlianza:


tlianza:
@ssbarnea - any ideas on the CI error? I think this is a good fix, at least when I run it locally it resolves my issue of setting the reporter.

adehad (approved):
Only legacy Travis CI failing, happy for this to get merged

studioj:
I would add a CI test for this @tlianza.
But I'm a QA person which is a nag for those things :-)

tlianza:
> I would add a CI test for this @tlianza.
> But I'm a QA person which is a nag for those things :-)

Is there an example I can follow?

ssbarnea (approved):

PR #1007, opened 2021-03-16
(fix): Adding accountId parameter on user

LKhoe:
Hello, I found a problem in user function and fixed it.
See you.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,6 +40,11 @@ Jira Python Library
         :alt: Requirements Status
 
 
+Note
+----------
+It's just and fix for the original lib
+
+
 This library eases the use of the Jira REST API from Python and it has been used in production for years.
 
 As this is an open-source project that is community maintained, do not be surprised if some bugs or features are not implemented quickly enough. You are always welcomed to use BountySource_ to motivate others to help.
--- a/jira/client.py
+++ b/jira/client.py
@@ -763,6 +763,12 @@ class JIRA(object):
         resource.find(ids)
         return resource
 
+    def get_changelogs_from_issue(self, issueID_or_key, base=JIRA_BASE_URL):
+        url = f'{self._options["server"]}/rest/api/2/issue/{issueID_or_key}/changelog'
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -40,6 +40,11 @@ Jira Python Library
         :alt: Requirements Status
 
 
+Note
+----------
+It's just and fix for the original lib
```
russoz, line 45: This text does not add much to the end user, IMO.

jira/client.py
```diff
@@ -2630,7 +2642,7 @@ def user(self, id, expand=None):
         :rtype: User
         """
         user = User(self._options, self._session)
-        params = {}
+        params = {'accountId': id}
```
russoz, line 2645: The "accountId" change happened a couple of years ago, but I am not sure (not that I looked for  it) what is the backward compatibility policy of this package.

So, it should be interesting not to force that upon all users.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = jira
+name = jira_333
```
russoz, line 2: why?

setup.py
```diff
@@ -3,4 +3,4 @@
 
 
 if __name__ == "__main__":
-    setuptools.setup(use_scm_version=True)
+    setuptools.setup(version='3.4.0')
```
russoz, line 6: Why are we pinning down a specific version here?

jira/client.py
```diff
@@ -763,6 +763,18 @@ def find(self, resource_format, ids=None):
         resource.find(ids)
         return resource
 
+    def get_changelogs_from_issue(self, issueID_or_key):
```
adehad, line 766: these functions do not follow the style of the rest of the class.

Here we could use `self._get_json()` with the default `base` argument, the next function can use the argument `base=AGILE_BASE_URL`

adehad (changes requested):

PR #1018, opened 2021-05-02
Cleanup project badges

ssbarnea:


README.rst
```diff
@@ -22,13 +19,6 @@ Jira Python Library
 .. image:: https://readthedocs.org/projects/jira/badge/?version=master
     :target: https://jira.readthedocs.io/
 
-.. image:: https://travis-ci.com/pycontribs/jira.svg?branch=master
-    :target: https://travis-ci.com/pycontribs/jira
```
adehad, line 26: Might be able to at the GHA badge here:

https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge

studioj (approved):
Just want to make sure that removing black is not by accident

README.rst
ssbarnea: I considered it but decided to delay until we are sure it is stable ;)

ssbarnea:
> Just want to make sure that removing black is not by accident



> Just want to make sure that removing black is not by accident

I love black, but I think it needs no more advertising. Its use is not relevant to the project.

adehad (approved):

PR #1019, opened 2021-05-02
Switch to locked test dependencies

ssbarnea:
* removes test extra and moves it test-requirements.in
* adds `deps` command for tox which should be used for *manually*
  bumping test dependencies when we see fit

Change-Id: I981c1c90f96051fd400e77efd0f725f27f03d286


adehad (dismissed):
After adding the lock files to the cspell.json ignorePaths, looks good to me.

Is the reasoning for removing the test as an extra because we don't include the tests as part of the whl, so there was no point including it anyway? \
If so, perhaps could also add the `test-requirements.*` into the exclude of the Manifest

dbaxa (approved):

PR #1021, opened 2021-05-03
docs: remove cspell reference in README

adehad:


studioj (approved):

studioj:
@ssbarnea do you know how to "fix" the licenses step?

adehad:
Some of the pip failures are a bit strange too, like this PyQt5 dependency ...

image


It is also strange that mypy is included as a dependency, despite it not being part of our `requirements.txt` file ...

¯\\\_(ツ)\_/¯

PR #1022, opened 2021-05-03
chore: isort imports

adehad:


studioj:
looks good => only issue for new devs is => how to add the pre-commit hook
maybe add a description in the readme to help (me) doing that ;-)

studioj (changes requested):
an addition to the readme for precommit would be a useful addition

adehad:
> an addition to the readme for precommit would be a useful addition

I think this is covered by running `tox -e lint`?

studioj:
Ah I thought it was maybe something like this. 

https://pre-commit.com

studioj (approved):

adehad:
> Ah I thought it was maybe something like this. 
> 
> https://pre-commit.com

It is the same actually. I haven't used it much before but my understanding is that you can use it manually like we have here or automatically per commit as a "pre commit hook" . It is probably more accessible to the community with the current setup to run it manually, but maybe I'm wrong hahah

dbaxa (approved):

studioj:
@ssbarnea can be merged

ssbarnea (approved):

Issue #1009, opened 2021-03-18
Disabling logging is not working


yo-ga:
**Describe the bug**
In [the document](https://jira.readthedocs.io/en/latest/api.html#jira.JIRA), there is a parameter named ```logging```.
The comment in the ```__init__``` says "Determine whether or not logging should be enabled. (Default: True)".
If the programmer tries to disable the log, it doesn't work.

**To Reproduce**
```py
from jira import JIRA
jira_client = JIRA("https://UNABLE_CONNECT_URL", auth=("username", "password"), logging=False)
```
Output:
```
WARNING:root:HTTPSConnectionPool(host='unable_connect_url', port=443): Max retries exceeded with url: /rest/auth/1/session (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x10ae850b8>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',)) while doing POST https://UNABLE_CONNECT_URL/rest/auth/1/session [{'data': '{"username": "username", "password": "password"}', 'headers': {'User-Agent': 'python-requests/2.24.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': 'application/json,*.*;q=0.9', 'Connection': 'keep-alive', 'Cache-Control': 'no-cache', 'Content-Type': 'application/json', 'X-Atlassian-Token': 'no-check'}}]
WARNING:root:Got ConnectionError [HTTPSConnectionPool(host='unable_connect_url', port=443): Max retries exceeded with url: /rest/auth/1/session (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x10ae850b8>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',))] errno:None on POST https://UNABLE_CONNECT_URL/rest/auth/1/session
{'response': None, 'request': <PreparedRequest [POST]>}\{'response': None, 'request': <PreparedRequest [POST]>}
WARNING:root:Got recoverable error from POST https://UNABLE_CONNECT_URL/rest/auth/1/session, will retry [1/3] in 8.383788695379353s. Err: HTTPSConnectionPool(host='unable_connect_url', port=443): Max retries exceeded with url: /rest/auth/1/session (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x10ae850b8>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',))
[...]

Issue #968, opened 2020-09-15
Client API, epic_id in add_issues_to_epic() should be string


SpringPie:
https://github.com/pycontribs/jira/blob/7fa3a454c08a14e2d7d7670fcfa87482e16936ba/jira/client.py#L4288

PR #1023, opened 2021-05-03
add typehints and allow google docstrings

adehad:
## Summary of Changes
1. Add typehints, coaxing mypy where required - i.e. `typing.cast` ,`isinstance()` and as a last resort `assert` 
2. Re-order `JIRAError` argument order to reflect how the base `Exception` class is used, i.e. first argument is a string, prevent misuse of `status_code` for a string of an error message
3. Allow "Google style" docstrings, allowing for more legible docstrings
4. The `basic_use.py` no longer worked, since GDPR the `emailAddress` property has been removed, changing the field used in the example to `.key` at least makes it run fine.
5. Partially (maybe fully?) address #1009 in order to satisfy mypy, refer to commit: "bugfix: respect logging argument" (https://github.com/pycontribs/jira/pull/1023/commits/cc504a1dc5a3975fc6ad1955b6fbf001577038c9).  
#329 review comments were checked, I don't feel the logging changes made here are particularly contentious - they didn't intend to be. I think a further PR should be raised if anything more needs to be done - as the changes here we mainly for the mypy fixes.
The implementation chosen imports `logging` as `_logging`, this should further prevent a developer accidentally using the raw logging module. Exposing the logger under `self.log` may also allow user to use the logger if they monkey patch or base class of the JIRA class.

~~This branch is currently based of the isort branch~~ now rebased

####  Todo:
- [x] ~~Document logging bug fixed on the client.py side, may need changes to the other classes, e.g.  resilient session~~ should be a separate PR if anything further is required than what was added to fix mypy
- [x] Document fix of JIRAError class, argument order causing misuse
- [x] ~~Convert `str = None` to `Optional[str] = None`  if the function has an actual none check~~ not required as inferred by mypy, could maybe check they are added to the docstrings, but might take that up in another PR, a bit burnt out as I severely under estimated the work required here haha
- [x] Check the actual docs build looks ok
- [x] do some more google style docstrings if you aren't tired of them yet, resources.py has a few left
- [x] where possible, convert assertions to properly typehinted variables
- [x] cleanup the commit history, no point separating by files as they all need to work together for mypy to be happy

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,7 +29,12 @@ needs_sphinx = "2.2.0"
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ["sphinx.ext.autodoc", "sphinx.ext.viewcode", "sphinx.ext.intersphinx"]
+extensions = [
+    "sphinx.ext.autodoc",
+    "sphinx.ext.intersphinx",
+    "sphinx.ext.napoleon",
+    "sphinx.ext.viewcode",
+]
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3.7", None),
@@ -42,9 +47,11 @@ intersphinx_mapping = {
 }
 
 autodoc_default_options = {
+    "member-order": "bysource",
     "members": True,
-    "undoc-members": True,
     "show-inheritance": True,
+    "special-members": "__init__",
+    "undoc-members": True,
 }
 
 nitpick_ignore = [
@@ -286,6 +293,11 @@ man_pages = [
 
 # If true, show URL addresses after external links.
 # man_show_urls = False
+
+# -- Options for Napoleon  -----------------------------------------------------
+
+napoleon_google_docstring = True
+napoleon_numpy_docstring = False  # Explicitly prefer Google style docstring
 
 
 # -- Options for Texinfo output ------------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,7 @@ from jira import __version__  # noqa
 # -- General configuration -----------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-needs_sphinx = "2.2.0"
+needs_sphinx = "4.0.0"
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
@@ -37,10 +37,8 @@ extensions = [
 ]
 
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3.7", None),
-    # until https://github.com/psf/requests/issues/5212 is addressed
-    # "requests": ("http://docs.python-requests.org/en/latest/", None),
-    "requests": ("https://requests.kennethreitz.org/en/master/", None),
+    "python": ("https://docs.python.org/3.8", None),
+    "requests": ("https://requests.readthedocs.io/en/latest/", None),
     "requests-oauthlib": ("https://requests-oauthlib.readthedocs.io/en/latest/", None),
     "ipython": ("https://ipython.readthedocs.io/en/stable/", None),
     "pip": ("https://pip.readthedocs.io/en/stable/", None),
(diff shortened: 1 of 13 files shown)


tuukkamustonen (commented):
Took a glance at this, then realized it's a massive diff (as would be expected). As I'm not familiar with the internals of the library, I stopped :)

I see there are 0 checks for PRs, which is a bit scary, but if it passes `mypy` (and other linters) checks, it's a good indication.

The most vital thing (imo) is to have the public API annotations in place (and detailed enough). Internals aid in development, but are normally not needed for users of the library. But this PR covering everything is a nice thing, for sure.

What should be added, is `py.typed` file, so that `mypy` can actually read the types from the package (I don't know why they require that, IDEs don't need it). Can you add that? See #1025.

docs/conf.py
```diff
@@ -287,6 +259,13 @@
 # If true, show URL addresses after external links.
 # man_show_urls = False
 
+# -- Options for Napoleon  -----------------------------------------------------
+
+napoleon_google_docstring = True
+napoleon_numpy_docstring = False  # Explicitly prefer Google style docstring
+napoleon_use_param = True  # for type hint support
```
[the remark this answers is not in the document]
adehad: `py.typed` file added, branch also rebased on to master confirming tests pass.

jira/resilientsession.py
```diff
@@ -166,25 +196,27 @@ def __verb(self, verb, url, retry_data=None, **kwargs):
...
+        return self.__verb("HEAD", str(url), **kwargs)
 
-    def patch(self, url, **kwargs):
-        return self.__verb("PATCH", url, **kwargs)
+    def patch(self, url: Union[str, bytes], data=None, **kwargs) -> Response:
+        return self.__verb("PATCH", str(url), data=data, **kwargs)
 
-    def options(self, url, **kwargs):
-        return self.__verb("OPTIONS", url, **kwargs)
+    def options(self, url: Union[str, bytes], **kwargs) -> Response:
```
tuukkamustonen, line 221: I think you could use [AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr) for `Union[str, bytes]`.

studioj:
i see you're changing the way the documentation is generated

maybe check this PR https://github.com/pycontribs/jira/pull/1039

jira/client.py
```diff
@@ -4295,15 +4508,15 @@ def add_issues_to_sprint(self, sprint_id, issue_keys):
...
         """Add the issues in ``issue_keys`` to the ``epic_id``.
 
-        :param epic_id: The ID for the epic where issues should be added.
-        :type epic_id: int
-        :param issue_keys: The issues to add to the epic
-        :type issue_keys: str
-        :param ignore_epics: ignore any issues listed in ``issue_keys`` that are epics. (Default: True)
-        :type ignore_epics: bool
+        Args:
+            epic_id (int): The ID for the epic where issues should be added.
```
studioj, line 4517: #968 mentions a mistake here maybe take it along

studioj: close the issue if you take it along

examples/basic_auth.py
```diff
@@ -16,7 +19,7 @@
 props = jira.application_properties()
 
 # Find all issues reported by the admin
-issues = jira.search_issues("assignee=admin")
+issues = cast(ResultList[Issue], jira.search_issues("assignee=admin"))
```
studioj, line 22: maybe explain this a little bit more => why cast?

examples/basic_use.py
```diff
@@ -42,6 +40,7 @@
 
 # Or modify the List of existing labels. The new label is unicode with no
 # spaces
+assert issue.fields.labels  # help mypy
```
studioj, line 43: just for my interest what do you mean with help mypy?

examples/basic_use.py
```diff
@@ -51,4 +50,4 @@
 # Linking a remote jira issue (needs applinks to be configured to work)
 issue = jira.issue("JRA-1330")
 issue2 = jira.issue("XX-23")  # could also be another instance
-jira.add_remote_link(issue, issue2)
+jira.add_remote_link(str(issue.id), issue2)
```
studioj, line 53: why cast to str? issue.id is a string already..

jira/resilientsession.py
adehad: Unfortunately we can't use `AnyStr` in this case. This is because this class is a subclass of `Session`, so the typehints need to be consistent with the parent class. Technically we could add a `# type: ignore[override]`, but I felt it unnecessary.


#### Example of mypy failure
```suggestion
    def options(self, url: AnyStr, **kwargs) -> Response:
```

```
jira\resilientsession.py:221: error: Signature of "options" incompatible with supertype "Session"
Found 1 error in 1 file (checked 19 source files)
```

examples/basic_use.py
adehad: So mypy thinks that the `issue.fields.labels` can be `None`, but it has certain ways to infer the type of a variable. This can be via `cast()`, `isinstance()` or `assert` statements - refer: https://mypy.readthedocs.io/en/stable/casts.html

jira/resources.py
```diff
@@ -480,34 +543,41 @@ class Issue(Resource):
...
         Resource.__init__(self, "issue/{0}", options, session)
 
-        self.fields = None
-        """ :type: :class:`~Issue._IssueFields` """
-        self.id = None
-        """ :type: int """
-        self.key = None
-        """ :type: str """
+        self.fields: Optional[Issue._IssueFields] = None
+        self.id: Optional[int] = None
```
adehad, line 568: Referring to both:
https://docs.atlassian.com/software/jira/docs/api/REST/8.13.6/#api/2/issue-getIssue

and 

https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issues/#api-rest-api-2-issue-issueidorkey-get

This Issue id field is a `str`

examples/basic_use.py
adehad: Yep you are completely right. The original typehint was an `Optional[int]` which was incorrect, it is always present, and is a `str`. I verified this both from the API documentation and by checking examples in jira.atlassian.com 

It also helped me remove the assertion in this file

jira/client.py
adehad: I've added it as a linked issue, I'll close after this PR is merged

ssbarnea:
Replase resolve conflicts.

adehad:
@ssbarnea fixed conflicts, happy to address the Codacy failure in a separate PR, seems like we should add pylint and mccabe to our lint step ?

ssbarnea (approved):

ssbarnea:
> @ssbarnea fixed conflicts, happy to address the Codacy failure in a separate PR, seems like we should add pylint and mccabe to our lint step ?
> 
> Looks like a few of them are already picked up by mypy and I've added the `ignore[override]` to keep mypy happy.

We already have pylint as part of pre-commit but maybe we have too many includes. Codacy can be ignored for the moment as what matters regards to linting is to pass pre-commit. Obviously that any change that goes towards making codacy happier is appreciated.

PR #1038, opened 2021-05-13
Fix docstring generation of jira client and minor doc layouting fix

studioj:
https://jira.readthedocs.io/en/master/api.html#jira.JIRA doesn't contain the info about the 'auth=' parameter. First commit should fix that


studioj:
added warnings in the examples for using `auth=(username, password)` and `basic_auth=(username, password)`

studioj:
second commit should fix the layout of the white background in relation with the line length

adehad (approved):

ssbarnea (approved):

Issue #1016, opened 2021-04-30
`rename_version` is not supported yet


ym1220:
Before raising any bug be sure that you used the latest release and that it still reproduces with the latest master. No other version is supported.

**Describe the bug**

**To Reproduce**
```
call the api rename_version as described in API-document
JIRA.rename_version(<JiraProject>,<oldnmae>,<newname>)
```
1. Any additional steps or considerations that happen before or after.

**Expected behavior**

**Stack Trace**
```
'JIRA' object has no attribute 'rename_version'
```

**Version Information**
Python Interpreter: python3
jira-python: 2.0.0
OS: Windows
IPython (Optional): <VERSION>
Other Dependencies:

**Additional context**

studioj:
you might want to have a look at the version resource

https://github.com/pycontribs/jira/blob/master/jira/resources.py#L873

how to find a version id is explained here
https://community.atlassian.com/t5/Jira-questions/How-do-I-get-the-fixversion-ID/qaq-p/395216#:~:text=Browse%20to%20a%20version%20in,where%2010100%20is%20the%20id.&text=so%20how%20or%20where%20do,fixversion%20IDs%20on%20Jira%206.1.

studioj:
@ym1220 
closing this issue 
See following links in the documentation:
- https://jira.readthedocs.io/en/master/api.html#jira.client.JIRA.rename_version
- https://jira.readthedocs.io/en/master/api.html#jira.resources.Version.update

if these items dont solve your question, please reopen and clarify

PR #1039, opened 2021-05-13
adding version to documentation to help out on #1016

studioj:
#1016 suggested a rename version feature.
although not implemented yet it seems possible with the Version.update functionality

This addition to the documentation will at least make it more clear where to look for such solutions

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -906,6 +906,21 @@ class Version(Resource):
         Refer to Atlassian REST API `documentation`_.
 
         .. _documentation: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-project-versions/#api-rest-api-2-version-id-put
+
+        :Example:
+
+            .. code-block:: python
+
+                >> version_id = "10543"
+                >> version = JIRA("https://atlassian.org").version(version_id)
+                >> print(version.name)
+                "some_version_name"
+                >> version.update(name="another_name")
+                >> print(version.name)
+                "another_name"
+                >> version.update(archived=True)
+                >> print(version.archived)
+                True
         """
         data = {}
         for field in kwargs:
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -879,10 +879,11 @@ class Version(Resource):
             self._parse_raw(raw)
 
     def delete(self, moveFixIssuesTo=None, moveAffectedIssuesTo=None):
-        """Delete this project version from the server.
+        """
+        Delete this project version from the server.
 
-        If neither of the arguments are specified, the version is
-            removed from all issues it is attached to.
+        If neither of the arguments are specified, the version is removed from all
+        issues it is attached to.
 
         :param moveFixIssuesTo: in issues for which this version is a fix
             version, add this argument version to the fix version list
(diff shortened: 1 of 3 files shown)


studioj:
i see @adehad  is also fixing this (halfly) with this PR https://github.com/pycontribs/jira/pull/1023/files
although my implementation doesn't follow his changed doc generation

jira/resources.py
```diff
@@ -899,7 +900,13 @@ def delete(self, moveFixIssuesTo=None, moveAffectedIssuesTo=None):
         return super(Version, self).delete(params)
 
     def update(self, **args):
-        """Update this project version from the server. It is prior used to archive versions."""
+        """
+        Update this project version from the server. It is prior used to archive versions.
+
+        link to atlassian restapi `documentation`_.
```
adehad, line 906: As this docstring will be user facing in the docstring, we can make it a little nicer ;)

```suggestion
        Refer to Atlassian  REST API `documentation`_ for all possible values.
```

jira/resources.py
```diff
@@ -899,7 +900,13 @@ def delete(self, moveFixIssuesTo=None, moveAffectedIssuesTo=None):
...
+
+        .. _documentation: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-project-versions/#api-rest-api-2-version-id-get
```
adehad, line 908:
```suggestion
        .. _documentation: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-project-versions/#api-rest-api-2-version-id-put
```

The current link points to the `get` method, the suggested link points to the `put` method which is what is used when updating.

jira/resources.py
```diff
@@ -899,7 +900,13 @@ def delete(self, moveFixIssuesTo=None, moveAffectedIssuesTo=None):
         return super(Version, self).delete(params)
 
     def update(self, **args):
```
adehad, line 902: it bothers me a little that this is `**args`, we would expect it to follow the convention of `**kwargs`. 

```suggestion
    def update(self, **kwargs):
```

This change would also require changes in the function to how it is used.

adehad (changes requested):
As you say I've made some changes in my branch too, I'm happy for this to be merged first, as my PR is a bit large and probably will take a while to get approved. Until then I think let's get this merged, only a few minor comments from me.

jira/resources.py
studioj: Well that would be a change in the public API which we should think twice to change...
I wouldn't do a change like this in this pr.

Maybe we should list up small improvements which might break the API and prepare for them

jira/resources.py
studioj: thx for checking (y)

jira/resources.py
adehad: Changing this variable name would not affect the public API. 

`**` indicates an unspecified number of **keyword** arguments. This variable is a catch all for these keyword arguments and it's name won't affect users in anyway.

studioj: darnit indeed you're right `*args ` and `**kwargs`

studioj: 🤔 this change is not tested. I'll add a test for it too

adehad (approved):
Excellent

ssbarnea (approved):

PR #1040, opened 2021-05-14
cleaning up most travis references

studioj:
since we moved away from travis no point in keeping references
- also removed most of the links to builds which were failing none of them were working anyhow

also restructured some of the code in exceptions as it was having a lot of duplicate if's

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,7 +85,6 @@ Development takes place on GitHub_:
 
 * ``master`` - (default branch) contains the primary development stream. Tags will be used to show latest releases.
 
-.. _GitHub: https://github.com/pycontribs/jira
 
 Setup
 =====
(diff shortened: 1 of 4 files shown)


adehad (approved):
Changes look fine to me. Some of the existing code looks a bit clunky, but that is separate from the intention of the PR so I won't add those comments here.

studioj:
> Changes look fine to me. Some of the existing code looks a bit clunky, but that is separate from the intention of the PR so I won't add those comments here.

Yeah did one cleanup code thing as part of this pr

a lot of duplicate if's really needed some reordering.

I'm trying to learn the library a bit by contributing, still some way to go :-)

ssbarnea (approved):

PR #1045, opened 2021-05-16
make tests more approachable

adehad:
#### Changes

1. Split the `tests.py` file into a series of smaller files.
2. Use `conftest.py` to store the helper methods together with a new base test case for easy access to the same jira `setUp()`
2. Made consistent use of pytest.skip vs unittest.skip, creating a `broken_test` shortcut to help with consistency
3. removed bad logic in `JIRAError` that was supposed to be CI specific, but ended up doing exactly the same thing as normal
4. 100% `jira.exceptions` [coverage](https://codecov.io/gh/pycontribs/jira/tree/feature%2Fmakes_tests_more_approachable/jira)
5. address a few of Codacity warnings from the original codebase
6. start some ground work for testing both Cloud and Server instances, the JiraTestManager now accepts an argument to specify what instance to authenticate with (although currently cloud authentication doesn't work)

#### Change Reasoning:

The single test file with >2000 lines was not approachable to new/casual contributors. This made it more likely for bugs to sneak in. 

The changes in this PR move the test `setUp()` and other helper methods into a separate `conftest.py`. This makes the actual test files simpler to understand. 

The test files have also been named to be easier to to find where a new test should belong.
i.e. I have added a new feature relating to `Attachment` I can add a test in `test_attachment.py`. 
Hopefully this naming scheme also makes it easier to remember that you can run `tox -e py39 -- tests/resources/test_attachment.py` to only test those tests.

I also think this approach may help a contributor who just wants to add tests, they can look at a coverage report, pick a test they want to add and probably find what file they should add the test to pretty quickly.

#### Todo

- [x] ~~Add a helper method to `JiraTestCase` to skip if on Cloud/Server?~~ won't do here
- [x] make jira.exceptions coverage 100%
- [x] Update readme with some suggestions on how to debug tests using VSCode
- [x] a lot of tests have been added to TestIssue e.g. for worklog and votes, that should be moved to a dedicated test_worklog / test_votes file


studioj:
like the idea!

studioj:
Quite a diff :-)

adehad:
> 
> 
> Quite a diff :-)

it really is a bad habit of mine haha

PR #1002, opened 2021-02-26
Export resource_class_map and fix bug in Role.add_user()

wlupton:
The reason for exporting `resource_class_map` is to allow an external module to support additional resource types without the need to edit resources.py (I had edited it previously but this became a maintenance nightmare).

The `Role.add_user()` bug is (I hope) an obvious one: the `groups` argument was ignored.

---

Some of my local extensions conflict with more recent extensions in `master`. For example, I supported notification schemes and permission schemes, but I took a different approach: I also added new resource types, and the methods return resources rather than dictionaries. For example:

Master:

```python
    @lru_cache(maxsize=None)
    def notificationschemes(self):
        # TODO(ssbarnea): implement pagination support
        url = self._options["server"] + "/rest/api/3/notificationscheme"

        r = self._session.get(url)
        data = json_loads(r)
        return data["values"]
```

Mine:

```python
    def notificationschemes(self) -> List[NotificationScheme]:
        """Get a list of notificationscheme Resources."""
        r_json = self._get_json('notificationscheme')
        notificationschemes = [
            NotificationScheme(self._options, self._session, raw_res_json) for
            raw_res_json in r_json['values']]
        return notificationschemes
```

There was already some level of `Role` support, and I took this approach to extending the interface (this is in a wrapper class that calls methods from the existing master class when it can):

```python
    @translate_resource_args
    def project_roles(self, project: IDorName, *, as_list: bool = False,
                      as_resources: bool = False):
        assert self._mode == 'online'
        # this returns a Dict[str, Dict[str, Any]]
        roles = self._jira.project_roles(project)
        # add a 'name' key (needed if will return a list)
        roles = {name: {**value, 'name': name} for name, value in
                 roles.items()}
        # if requested, convert to Dict[str, ProjectRole]
        if as_resources:
            roles = {name: ProjectRole(self._options, self._session, value) for
                     name, value in roles.items()}
        # if requested, convert to List[]
        if as_list:
            roles = list(roles.values())
        return roles
```

Please let me know if you'd like me to contribute any of these extensions.

PR #1047, opened 2021-05-18
Export resource_class_map and fix bug in Role.add_user()

wlupton:
The reason for exporting resource_class_map is to allow an external module to support additional resource types without the need to edit resources.py (I had edited it previously but this became a maintenance nightmare).

The Role.add_user() bug is (I hope) an obvious one: the groups argument was ignored.

(This is a duplicate of PR #1002.)

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -46,6 +46,7 @@ __all__ = (
     "Customer",
     "ServiceDesk",
     "RequestType",
+    "resource_class_map",
 )
 
 logging.getLogger("jira").addHandler(logging.NullHandler())
@@ -938,7 +939,7 @@ class Role(Resource):
         if groups is not None and isinstance(groups, str):
             groups = (groups,)
 
-        data = {"user": users}  # FIXME: groups is not used.
+        data = {"user": users, "group": groups}
         self._session.post(self.self, data=json.dumps(data))
 
 


adehad (approved):
Verified with the API documentation that `group` is the right parameter. Happy for this to get merged.


#### Side Note
However, there is a difference in the Cloud vs Server versions here I believe.

##### Cloud:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-project-role-actors/#api-rest-api-2-project-projectidorkey-role-id-post
- Posting to the  `/rest/api/2/project/{projectIdOrKey}/role/{id}` endpoint

##### Server/Data Center
https://docs.atlassian.com/software/jira/docs/api/REST/8.13.6/#api/2/role-addProjectRoleActorsToRole
- Posting to the `/rest/api/2/role/{id}/actors` endpoint

Looking at the endpoint used in the `__init__` it seems like this is using the Cloud endpoint. So this change will work for those already using this function. But not for the version we have CI against. But as we don't have any tests (the only test is skipped as it is broken: https://github.com/BroadbandForum/jira/blob/02906938c1f96db653aaa4965c640a8f38478ce6/tests/tests.py#L1650 ) the CI won't fail, which I guess we are fine with at the moment. This is more of a bugfix considering the existing situation.

wlupton:
I think you are quoting two different endpoints?
* `/rest/api/2/role/{id}` relates to global role definitions 
* `/rest/api/2/project/{projectIdOrKey}/role/{id}` relates to a project's roles (which must be selected from the global roles)

(I've only ever used the server endpoint.)

adehad:
Ah yes. My mistake. I've updated the original approving comment.

ssbarnea (approved):

PR #1059, opened 2021-05-28
Feature/add board projects

thefifo:
Not sure what tests should be added as I couldn't find any for similar functions (e.g., boards)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4339,12 +4339,12 @@ class JIRA(object):
         """Get a list of project GreenHopperResources.
 
         Args:
-            board_id (int): the board to get sprints from
+            board_id (int): the board to get projects from
             extended (bool): Used only by old GreenHopper API to fetch additional information like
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -4329,6 +4329,49 @@ def sprints(
...
+
+        if (
+            self._options["agile_rest_path"]
+            == GreenHopperResource.GREENHOPPER_REST_PATH
+        ):
+            raise NotImplementedError(
+                "Jira Agile Public API does not support this request"
+            )
+
+        else:
```
studioj, line 4364: this else is not needed since you raise excpetions when the inputs are wrong

studioj (commented):
as a QA person i would like to suggest adding a test => add 2 boards and get them

also i think the method is not really named well. could you be more specific?

studioj:
no response after +1 yr after requesting changes => close

PR #1062, opened 2021-06-01
Update client.py

chenchusheng:
expand =“names"，return


chenchusheng:
expand =“names"，return

studioj (commented):
- please use tox to check your code before pushing => currently the black check fails
- as a QA person i would really like to see a test added to check for that specific functionality so it wont break in the future

studioj:
@chenchusheng?

studioj:
no response after +1 yr after requesting changes => close

PR #1063, opened 2021-06-01
Add (some) missing type hints for `_IssueFields`

julenpardo:
## Description

Tested with mypy 0.782.

Pull request https://github.com/pycontribs/jira/pull/1023 added some type hints, which is great, but didn't define all of them, which now makes mypy unhappy about it.

**Note**: this PR probably doesn't add all of them either, but it does add some of them that do exist (most notably, fields like `summary` or `created`). I haven't found in the Jira REST API documentation a specific list of the fields that have a required value, so the types added by this PR are based in the errors of our codebase, intuition and requests examples provided by Jira docs. I'd be happy to add more fields if necessary.

## Rationale

Because mypy assumes the type `Any` for when there's no type hint, the solution should be to either to define every class attribute with its type, or define no types at all. This design by mypy is intentional for backwards compatibility (among other reasons). Having partially defined types can only break existing set ups.

## Minimal reproducible example

```python
from jira import JIRA


jira_client = JIRA('http://jira.test')
issue = jira_client.issue('TEST-1')
print(issue.fields.summary)
```

```
venv ❯❯❯ mypy script.py
script.py:7: error: "_IssueFields" has no attribute "summary"
Found 1 error in 1 file (checked 1 source file)
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,7 +57,7 @@ autodoc_inherit_docstrings = False
 nitpick_ignore = [
     ("py:class", "JIRA"),  # in jira.resources we only import this class if type
     ("py:obj", "typing.ResourceType"),  # only Py36 has a problem with this reference
-    ("py:class", "jira.resources.MyAny")  # Dummy subclass for type checking
+    ("py:class", "jira.resources.MyAny"),  # Dummy subclass for type checking
     # From other packages
     ("py:mod", "filemagic"),
     ("py:mod", "ipython"),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,7 +69,7 @@ nitpick_ignore = [
     ("py:class", "Response"),
     ("py:mod", "requests-kerberos"),
     ("py:mod", "requests-oauthlib"),
-    ("py:class", "jira.resources.MyAny")
+    ("py:class", "jira.resources.MyAny"),
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,6 +57,7 @@ autodoc_inherit_docstrings = False
 nitpick_ignore = [
     ("py:class", "JIRA"),  # in jira.resources we only import this class if type
     ("py:obj", "typing.ResourceType"),  # only Py36 has a problem with this reference
+    ("py:class", "jira.resources.MyAny")  # Dummy subclass for type checking
     # From other packages
     ("py:mod", "filemagic"),
     ("py:mod", "ipython"),
@@ -69,7 +70,6 @@ nitpick_ignore = [
     ("py:class", "Response"),
     ("py:mod", "requests-kerberos"),
     ("py:mod", "requests-oauthlib"),
-    ("py:class", "jira.resources.MyAny"),
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,6 +69,7 @@ nitpick_ignore = [
     ("py:class", "Response"),
     ("py:mod", "requests-kerberos"),
     ("py:mod", "requests-oauthlib"),
+    ("py:class", "jira.resources.MyAny")
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -22,6 +22,8 @@ if TYPE_CHECKING:
 else:
 
     class MyAny(object):
+        """Dummy subclass of base object class for when type checker is not running.
+        """
         pass
 
 
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -22,8 +22,8 @@ if TYPE_CHECKING:
 else:
 
     class MyAny(object):
-        """Dummy subclass of base object class for when type checker is not running.
-        """
+        """Dummy subclass of base object class for when type checker is not running."""
+
         pass
 
 
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -18,6 +18,13 @@ from jira.utils import CaseInsensitiveDict, json_loads, threaded_requests
 if TYPE_CHECKING:
     from jira.client import JIRA
 
+    MyAny = Any
+else:
+
+    class MyAny(object):
+        pass
+
+
 __all__ = (
     "Resource",
     "Issue",
@@ -574,7 +581,7 @@ class Filter(Resource):
 class Issue(Resource):
     """A Jira issue."""
 
-    class _IssueFields(object):
+    class _IssueFields(MyAny):
         class _Comment(object):
             def __init__(self) -> None:
                 self.comments: List[Comment] = []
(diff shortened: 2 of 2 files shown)


adehad (changes requested):
Great PR. 

I imagine this will also be a pain for basically every customfield property too.

Two options I'm thinking:
1. Use an `Any`-ish typehint `self.fields: Union[Any, Issue._IssueFields]`
2. Inherit `Any` in the `_IssueFields` class during type checking

The inherit `Any` is a bit hacky, so perhaps we will just use `Any` in the future, but:

```python
if TYPE_CHECKING:
    from jira.client import JIRA

    MyAny = Any
else:

    class MyAny(object):
        pass
```

Then:
```python
    class _IssueFields(MyAny):
```

Otherwise the fields you've added seem fine, some of them I can't see on my instance, or the https://jira.atlassian.com instance, but you've set them as optional so don't think it'll hurt to keep them in.

@studioj any thoughts ?

jira/resources.py
```diff
@@ -571,6 +586,7 @@ def __init__(
         self.fields: Issue._IssueFields
         self.id: str
         self.key: str
+        self.raw: Dict[Any, Any] = {}
```
adehad, line 589:
```suggestion
```

This shouldn't be here. 

In `Resource.__init__`
```python
        # Explicitly define as None so we know when a resource has actually
        # been loaded
        self.raw: Optional[Dict[str, Any]] = None
```

If you want to fix this you should instead:

```python
        if raw:
            self._parse_raw(raw)
        self.raw: Dict[str, Any]
```

`_parse_raw` will throw an error is `raw` is None so you can guarantee that `self.raw` is a `Dict[str,Any]` after `_parse_raw` is called.

adehad (changes requested):

jira/resources.py
julenpardo: Thanks for the heads up, I've already did it the `cast()` way after parsing the raw dict, the type checker is working as expected for me

julenpardo:
Thanks @adehad for the feedback. 

Regarding the customfields, the types I've defined have been made under the assumption that they'd be common for any Jira instance (of course my assumptions may be wrong, and the Jira documentation doesn't really help much for this, at least for what I could find), and that, if someone was looking for some custom fields, they would have to check that the variable is bound.

But if subclassing `Any` is a more feasible solution, I'm also good with that too

adehad:
> But if subclassing `Any` is a more feasible solution, I'm also good with that too
 
Great, let's wait a little bit in case anyone else has any thoughts on the matter.

If I was being picky I would suggest doing the `raw` type casting for all the `Resource` subclasses, but that can also be done as a separate PR

julenpardo:
@adehad 

> Great, let's wait a little bit in case anyone else has any thoughts on the matter.

That makes sense, let's wait for other opinions

> If I was being picky I would suggest doing the raw type casting for all the Resource subclasses, but that can also be done as a separate PR

No issue, I can do it in this PR already! I'll push the changes in a while

studioj:
> Great PR.
> 
> I imagine this will also be a pain for basically every customfield property too.
> 
> Two options I'm thinking:
> 
> 1. Use an `Any` typehint `self.fields: Any`, removing `_IssueFields`
> 2. Inherit `Any` in the `_IssueFields` class during type checking
> 
> The inherit `Any` is a bit hacky, so perhaps we will just use `Any` in the future, but:
> 
> ```python
> if TYPE_CHECKING:
>     from jira.client import JIRA
> 
>     MyAny = Any
> else:
> 
>     class MyAny(object):
>         pass
> ```
> 
> Then:
> 
> ```python
>     class _IssueFields(MyAny):
> ```
> 
> Otherwise the fields you've added seem fine, some of them I can't see on my instance, or the https://jira.atlassian.com instance, but you've set them as optional so don't think it'll hurt to keep them in.
> 
> @studioj any thoughts ?

tbh I'm not that acquainted w type hints. :-) I only recently finally made the step to drop legacy python in my professional life. 
type hints are an added feature which is on my to learn list :-)

adehad:
@ssbarnea any thoughts on the maintenance of these typehints?

Least maintenance would be `self.fields: Any`

The `MyAny` proposal is also technically low maintenance from a typehint point of view, but may break at the `mypy` level

julenpardo:
@adehad I've added the types for the `raw` attributes of the rest `Resource` subclasses, let me know if yuo would address something else!

adehad (dismissed):
@julenpardo, great! Although I noticed it looks like you have missed 3 instances at the bottom of the file for:  `Customer`, `ServiceDesk`, `RequestType`

julenpardo:
@adehad Thanks for the catch! Already added, also for `UnknownResource`

adehad (commented):
Nice, looks good. Now just waiting for a decision on the `Any` situation from @ssbarnea

julenpardo:
Hey folks, any follow up regarding defining types as `Any`?

ssbarnea:
I do not have time to look into full details of this but use of Any is almost always a type problem.
It work mentioning that JSON data has a limited number of return data types, so we could narrow it down and avoid use of Any.

adehad:
Hi @julenpardo let's implement the `MyAny` suggestion, I've been given merge permissions now, so I should be able to merge this after that.

julenpardo:
I'm not strong opinionated about this. As of me, having no types at all (specially when just some of them were added) would still be fine. I can understand that the argument in favor of `Any` is a lower maintenance effort, but also its drawbacks.

I originally opened this PR with the intention solely of unbreaking the type checker, so any decision taken here would be okay from my side.

julenpardo:
Thanks @adehad, I posted the previous comment just before seeing yours :) I'll make the changes and ping you back

julenpardo:
@adehad I've made `_IssueFields` subclass `Any` for the type checking, but kept the type hints I also defined in the constructor, should we remove them or just leave it as it is?

docs/conf.py
```diff
@@ -69,6 +69,7 @@
     ("py:class", "Response"),
     ("py:mod", "requests-kerberos"),
     ("py:mod", "requests-oauthlib"),
+    ("py:class", "jira.resources.MyAny"),
```
adehad, line 72: this should be placed a bit higher, above the comment `# From other packages`

adehad (changes requested):
@julenpardo let me get back to you on the existing typehints, I think we may be able to just leave them in. For now could you merge in the latest master to make sure the pipeline is remains all good?

@ssbarnea regarding the `Any` type hint - agree, we are basing the json fields on a `Dict[str,Any]` - which is what we need to rely on until we get recursive type support in mypy: https://github.com/python/typing/issues/182#issuecomment-185996450 and python/mypy#731. We have to fallback to `Any` here because we support accessing the json dict fields with `__getattr__` , so it gets a bit funky, for example if there is a custom field we haven't type hinted

julenpardo:
@adehad Thanks for following up with this, already rebased latest master and addressed the comment

adehad (approved):
Great. I think this is good.

I wanted to check if VSCode would pick up on the types of `_IssueFields` despite inheriting from `Any` during type checking, and I can confirm it does (using PyLance as the language server):

image

ssbarnea:
@adehad I personally found `MyAny` horrible from the naming point of view but on the other hand we can change it later. Indeed when it comes to type hints mypy and behavior of vscode (pylance) are the magic combination. If one of them does complain it is a clear sign we did something wrong.

PR #1074, opened 2021-06-22
Refactor locking of test dependencies

ssbarnea:
Use a single constraints.txt file that is pinning all test requirements. The file is updated manually running tox -e deps.

This should avoid being notified by dependabot about outdated deps, especially test ones.

All projects deps are sourced from inside setup.cfg.


tox.ini
```diff
@@ -115,7 +113,8 @@ usedevelop = false
 deps = pre-commit>=1.17.0
 commands=
     python -m pre_commit run --color=always {posargs:--all}
-extras =
+setenv =
+    PIP_CONSTRAINT=/dev/null
```
adehad, line 117: when creating/recreating the tox env this doesn't play nice with Windows, I had to instead get it working with an empty variable using:

```suggestion
    PIP_CONSTRAINT=
```

PR #1088, opened 2021-07-01
fix: Allow unassigning issues

patrick-ringl-by:


studioj (approved):
looks legit
it would be awesome if you could add an example in the doc section

patrick-ringl-by:
> looks legit
> it would be awesome if you could add an example in the doc section

Done. :)

studioj (approved):

studioj:
@ssbarnea as far as I can see ready, to merge

adehad:
@patrick-ringl-by, do you think you could add this fix into `_get_user_id()` itself? i.e. similar to the #902.

adehad (changes requested):
@patrick-ringl-by, do you think you could add this fix into `_get_user_id()` itself? i.e. similar to the #902.

My vision was something like:

```diff
--    def _get_user_id(self, user: str) -> str:
++    def _get_user_id(self, user: Optional[str]) -> Optional[str]:
        """Internal method for translating an user search (str) to an id.

++        Return None and -1 unchanged.

        This function uses :py:meth:`JIRA.search_users` to find the user
        and then using :py:meth:`JIRA._get_user_identifier` extracts
        the relevant identifier property depending on whether
        the instance is a Cloud or self-hosted Instance.


        Args:
--            user (str): The search term used for finding a user.
++            user (Optional[str]): The search term used for finding a user.
++              None, '-1' and -1 are equivalent to 'Unassigned'.

        Raises:
            JIRAError: If any error occurs.

        Returns:
            str: The Jira user's identifier.
        """
++        if user in (None, -1, "-1"):
++            return user
.
.
.
```

patrick-ringl-by:
One (personal) remark: Positioning the handling of user input validation (which the exceptional handling of `None`,`-1`, and `"-1"` is) further down the call-stack - instead of handling it early - feels wrong to me. I am interested: why you prefer such a solution? :)

Regarding your concrete, proposed patch:

- `-1` (`int`) in the signature of `_def _get_user_id` is (deliberately?) ignored
- the docstring contains a mention of `None` and `-1` but no mention of `"-1"`
- IIRC `None` and `-1` are not equivalent for the JIRA API .. the former should result in an unassignment while the latter should assign to "the default assignee".

This can be shown in a test-case like:

```diff
+
+    def test_assign_to_default_assignee_str(self):
+        self.assertTrue(self.jira.assign_issue(self.issue_1, self.user_normal.name))
+        self.assertEqual(
+            self.jira.issue(self.issue_1).fields.assignee.name, self.user_normal.name
+        )
+        self.assertTrue(self.jira.assign_issue(self.issue_1, "-1"))
+        self.assertEqual(
+            self.jira.issue(self.issue_1).fields.assignee.name, self.user_admin.name
+        )
+
```

Before filing another PR: Since you seem to have in mind *how* this should all work out - what potential solution is the one you favour here? :)

adehad:
Hi @patrick-ringl-by thanks for raising these points, sorry my original comment didn't explain myself well

>  Positioning the handling of user input validation (which the exceptional handling of None,-1, and "-1" is) further down the call-stack  - instead of handling it early - feels wrong to me.

My thoughts is that we are still handling it early, just because we are deeper in a call stack doesn't mean it is fundamentally happening later in the program logic? Can you elaborate on why you feel that a deeper call stack is a potential problem?

We are trying to keep each function to focus on one thing, here we want `_get_user_id()` itself to focus on the user id selection including handling validation for edge cases of the input user.
By putting edge case handling in `assign_issue()` we are making it perform logic for both user validation of an input user **and** actually performing the assignment (you may know this is as increasing the coupling in the code - which you usually want to avoid). 

So in summary:
1. we want to have each function have a single responsibility - `_get_user_id()`  to handle a provided 'user' , `assign_issue()` to actually perform the assignment. This practice makes code easier to read and can help prevent code duplication.
2. I don't think using a separate function (going deeper into a call stack) is influencing the code logic, we are still fundamentally checking the user input edge cases at the start if the `_get_user_id()` function. 

Happy to discuss this further and come to perhaps a better solution.

> -1 (int) in the signature of _def _get_user_id is (deliberately?) ignored
> the docstring contains a mention of None and -1 but no mention of "-1"

Yes, but remember that not everyone that uses this package will also be using `mypy` or other type checkers, so these types may not be enforced. I am happy for us to add `"-1"` to the docstring or any further changes you think may be appropriate, perhaps we should change the type hint to be more verbose? A `Union[Optional[str], Literal[-1]]` is probably the most accurate.

> IIRC None and -1 are not equivalent for the JIRA API .. the former should result in an unassignment while the latter should assign to "the default assignee".

This is exactly right:
https://docs.atlassian.com/software/jira/docs/api/REST/8.13.6/#api/2/issue-assign
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issues/#api-rest-api-2-issue-issueidorkey-assignee-put


The docstring of the `assign_issue()` function states this too:
```
"""Assign an issue to a user. None will set it to unassigned. -1 will set it to Automatic."""
```

We want to support both these cases together with the case for a valid user. Your test case examples therefore are excellent and should be two separate test cases, one for "unassigned" and another for "default_assignee".

studioj:
@patrick-ringl-by , did you understand @adehad s proposal?

adehad:
Refer #1152

PR #1102, opened 2021-07-15
Add subscriptable support for PropertyHolder

dvaerum:
This would be a simple, nice to have feature for dealing with custom fields 😃 

It makes it possible to do
```python
some_custom_field = "customfield_15623"
value = issue.fields[some_custom_field]
```
instead of
```python
some_custom_field = "customfield_15623"
value = getattr(issue.fields, some_custom_field)
```

Alternatively, I will just keep patching it in
```python
from jira import resources as jira_resources

def _jira_issue_property_holder_get_item(cls, item):
    if item.startswith("_"):
        raise KeyError(item)
    else:
        return getattr(cls, item)

jira_resources.PropertyHolder._getitem__ = _jira_issue_property_holder_get_item
```

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -692,7 +692,7 @@ class Issue(Resource):
 
         super(Issue, self).update(async_=async_, jira=jira, notify=notify, fields=data)
 
-    def get_field(self, fieldName: str) -> Any:
+    def get_field(self, field_name: str) -> Any:
         """Obtain the (parsed) value from the Issue's field.
 
         Args:
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -696,7 +696,7 @@ class Issue(Resource):
         """Obtain the (parsed) value from the Issue's field.
 
         Args:
-            fieldName (str): The name of the field to get
+            field_name (str): The name of the field to get
 
         Raises:
             AttributeError: If the field does not exist or if the field starts with an ``_``
@@ -705,13 +705,13 @@ class Issue(Resource):
             Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
         """
 
-        if fieldName.startswith("_"):
+        if field_name.startswith("_"):
             raise AttributeError(
-                f"An issue field_name cannot start with underscore (_): {fieldName}",
-                fieldName,
+                f"An issue field_name cannot start with underscore (_): {field_name}",
+                field_name,
             )
         else:
-            return getattr(self.fields, fieldName)
+            return getattr(self.fields, field_name)
 
     def add_field_value(self, field: str, value: str):
         """Add a value to a field that supports multiple values, without resetting the existing values.
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -1431,3 +1431,9 @@ def cls_for_resource(resource_literal: str) -> Type[Resource]:
 class PropertyHolder(object):
     def __init__(self, raw):
         __bases__ = raw  # noqa
+        
+    def __getitem__(self, item)
+        if item.startswith("_"):
+            raise KeyError(item)
+        else:
+            return getattr(self, item)
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -692,6 +692,27 @@ class Issue(Resource):
 
         super(Issue, self).update(async_=async_, jira=jira, notify=notify, fields=data)
 
+    def get_field(self, fieldName: str) -> Any:
+        """Obtain the (parsed) value from the Issue's field.
+
+        Args:
+            fieldName (str): The name of the field to get
+
+        Raises:
+            AttributeError: If the field does not exist or if the field starts with an ``_``
+
+        Returns:
+            Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
+        """
+
+        if fieldName.startswith("_"):
+            raise AttributeError(
+                f"A issue field_name cannot start with underscore (_): {fieldName}",
+                fieldName,
+            )
+        else:
+            return getattr(self.fields, fieldName)
+
     def add_field_value(self, field: str, value: str):
         """Add a value to a field that supports multiple values, without resetting the existing values.
 
@@ -1431,9 +1452,3 @@ def cls_for_resource(resource_literal: str) -> Type[Resource]:
 class PropertyHolder(object):
     def __init__(self, raw):
         __bases__ = raw  # noqa
-        
-    def __getitem__(self, item)
-        if item.startswith("_"):
-            raise KeyError(item)
-        else:
-            return getattr(self, item)
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -707,7 +707,7 @@ class Issue(Resource):
 
         if fieldName.startswith("_"):
             raise AttributeError(
-                f"A issue field_name cannot start with underscore (_): {fieldName}",
+                f"An issue field_name cannot start with underscore (_): {fieldName}",
                 fieldName,
             )
         else:
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -26,6 +26,18 @@ class IssueTests(JiraTestCase):
         self.assertEqual(issue.key, self.issue_1)
         self.assertEqual(issue.fields.summary, "issue 1 from %s" % self.project_b)
 
+    def test_issue_get_field(self):
+        issue = self.jira.issue(self.issue_1)
+        self.assertEqual(issue.fields.description, issue.get_field(fieldName="description"))
+
+        self.assertRaisesRegex(
+            AttributeError, ": _something", issue.get_field, "_something"
+        )
+
+        self.assertRaisesRegex(
+            AttributeError, "customfield_1234", issue.get_field, "customfield_1234"
+        )
+
     def test_issue_field_limiting(self):
         issue = self.jira.issue(self.issue_2, fields="summary,comment")
         self.assertEqual(issue.fields.summary, "issue 2 from %s" % self.project_b)
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -32,13 +32,11 @@ class IssueTests(JiraTestCase):
             issue.fields.description, issue.get_field(fieldName="description")
         )
 
-        self.assertRaisesRegex(
-            AttributeError, ": _something", issue.get_field, "_something"
-        )
+        with self.assertRaisesRegex(AttributeError, ": _something"):
+            issue.get_field("_something")
 
-        self.assertRaisesRegex(
-            AttributeError, "customfield_1234", issue.get_field, "customfield_1234"
-        )
+        with self.assertRaisesRegex(AttributeError, "customfield_1234"):
+            issue.get_field("customfield_1234")
 
     def test_issue_field_limiting(self):
         issue = self.jira.issue(self.issue_2, fields="summary,comment")
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -29,7 +29,7 @@ class IssueTests(JiraTestCase):
     def test_issue_get_field(self):
         issue = self.jira.issue(self.issue_1)
         self.assertEqual(
-            issue.fields.description, issue.get_field(fieldName="description")
+            issue.fields.description, issue.get_field(field_name="description")
         )
 
         with self.assertRaisesRegex(AttributeError, ": _something"):
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -28,7 +28,9 @@ class IssueTests(JiraTestCase):
 
     def test_issue_get_field(self):
         issue = self.jira.issue(self.issue_1)
-        self.assertEqual(issue.fields.description, issue.get_field(fieldName="description"))
+        self.assertEqual(
+            issue.fields.description, issue.get_field(fieldName="description")
+        )
 
         self.assertRaisesRegex(
             AttributeError, ": _something", issue.get_field, "_something"


studioj:
sounds cool... a solution to a problem we all encountered, i cant say if this is an ideal solution on the other hand :-s seems to be open for much subjectiveness ... I would prefer some other people to have a look before approving.

in any case, please extend the tests. We wont merge anything without a proper test

dvaerum:
> in any case, please extend the tests. We wont merge anything without a proper test.

I will try to have a look next week at work, to see if I can create some proper test for review, pointers are welcome :wink:

studioj:
@adehad your 2 cents on this would be appreciated. As said it's a possible solution to a problem many people are having. (Including me)
It doesn't make it auto completable on the other hand... Which is something we're trying to aim for with the mypy typing etc...
So I'm not 100%sure this is THE solution we should opt for.
On the other hand it's a definitely an improvement for useability.
I was thinking to do something with the actual name of the field. Instead of this custom_blabla thing.

adehad:
> It doesn't make it auto completable on the other hand... Which is something we're trying to aim for with the mypy typing etc...

I think auto-completion/suggestions won't be possible just yet with how mypy and typing work. I'd be gladly proved wrong, but I think let's not worry about that for now.


> I was thinking to do something with the actual name of the field. Instead of this custom_blabla thing. I'm guessing that's actually what @dvaerum wants to fix in the end.
> 
> First look up the custom field number then get the atr since you can only get the string from JIRA().fields()...

So we currently have `_fields` set in the `__init__` which will translate the custom field name to the id.

https://github.com/pycontribs/jira/blob/f26ea39a1cf443e2be7ea503911f0a4fcd668dc2/jira/client.py#L497-L501

Which is used in `search_issues`:

https://github.com/pycontribs/jira/blob/f26ea39a1cf443e2be7ea503911f0a4fcd668dc2/jira/client.py#L2722-L2727

So perhaps we could look into creating a function that goes from field name to id, although I have a feeling that there may be cases where custom fields with the same name can exist - so I personally wouldn't want to rely on a solution like that in my work, but no harm documenting this limitation in such a function.

> @adehad your 2 cents on this would be appreciated. As said it's a possible solution to a problem many people are having. (Including me)

Apart from `__getattr__` raising a `KeyError` (it should raise an `AttributeError` really), my only concern is what this would do to other non-Issue objects. If we are able to verify with some test cases that this is fine then I'm happy with this approach.

What I've done in my work is actually something like:

```python
some_custom_field = "customfield_15623"
value = issue.raw["fields"][some_custom_field]
```

From my perspective the most safe way to implement this would be so add directly to the `Issue` class something like a `get_field()` function:

```python
    def get_field(self, field_name):
        return self.raw["fields"].get(field_name)
```

dvaerum:
> From my perspective the most safe way to implement this would be so add directly to the `Issue` class something like a `get_field()` function:
> 
> ```python
>     def get_field(self, field_name):
>         return self.raw["fields"].get(field_name)
> ```

I see what you @adehad  with a method being a more preferred way to go, but I don't like that it will return `None` if the field does not exist because I feel that one could easily run into some small problems debugging to find a typo this way, instead of just right away being told that something is wrong

> > It doesn't make it auto completable on the other hand... Which is something we're trying to aim for with the mypy typing etc...
> 
> I think auto-completion/suggestions won't be possible just yet with how mypy and typing work. I'd be gladly proved wrong, but I think let's not worry about that for now.

Regarding auto-complete, I also don't know how one could do this, but I did get an idea there could help a little (inspired by @adehad preference suggestion)

```python
def get_field(self, field_name: str) -> Any:
    ...

def get_field_int(self, field_name: str) -> int:
    ...

def get_field_str(self, field_name: str) -> str:
    ...

def get_field_dict(self, field_name: str) -> Dict[str, Any]:
    ...

def get_field_list(self, field_name: str) -> List[Any]:
    ...
```
I just don't know if this is something you wanted, or if it becomes too complete, but my logic is that if you want to get/use a customfield, you already know what type it is, so this way you would get (some) auto-completion for whatever variable is returned

adehad:
> I see what you @adehad with a method being a more preferred way to go, but I don't like that it will return None if the field does not exist because I feel that one could easily run into some small problems debugging to find a typo this way, instead of just right away being told that something is wrong

Yep I was just copying the behaviour of `getattr()` in your original example, I don't mind using square brackets access.

> I just don't know if this is something you wanted, or if it becomes too complete, but my logic is that if you want to get/use a customfield, you already know what type it is, so this way you would get (some) auto-completion for whatever variable is returned

I think for now we can encourage a use like this for the type suggestion:

```python
def get_field(self, field_name: str) -> Any:
    ...

# user to specify specific types
my_int: int = issue.get_field("customfield_15623")
my_str: str = issue.get_field("customfield_15623")

```

I was thinking @studioj might have been referring to autocompletion in the editor, e.g.
image

where currently we have listed some fields that should exist in the `_IssueFields` object to allow some basic autocompletion and type support. Which I think would be difficult to extend for custom fields as they vary between users.

jira/resources.py
```diff
@@ -692,6 +692,15 @@ def update(  # type: ignore[override] # incompatible supertype ignored
...
         super(Issue, self).update(async_=async_, jira=jira, notify=notify, fields=data)
 
+    def get_field(self, field_name: str) -> Any:
+        if field_name.startswith("_"):
+            raise AttributeError(
+                f"A issue field_name cannot start with underscore (_): {item}",
+                field_name,
+            )
+        else:
+            return getattr(self.fields, field_name)
```
adehad, line 702: Please write a docstring for this function and add a comment here that explains why you are using `getattr()`.

My understanding is that you are using `getattr()` so you can get the custom types e.g. `Project` class for the `.project` property

```suggestion
            # use getattr so we are able to return the custom types obtained from `parse_raw()`
            return getattr(self.fields, field_name)
```

```python
    def get_field(self, field_name: str) -> Any:
        """Obtain the (parsed) value from the Issue's field. 

        Args:
            field_name (str): The name of the field to get

        Raises:
            AttributeError: If the field does not exist or if the field starts with an ``_``

        Returns:
            Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
        """

```

jira/resources.py
```diff
@@ -692,6 +692,15 @@ def update(  # type: ignore[override] # incompatible supertype ignored
 
         super(Issue, self).update(async_=async_, jira=jira, notify=notify, fields=data)
 
+    def get_field(self, field_name: str) -> Any:
+        if field_name.startswith("_"):
+            raise AttributeError(
+                f"A issue field_name cannot start with underscore (_): {item}",
+                field_name,
```
adehad, line 699:
```suggestion
                f"A issue field_name cannot start with underscore (_): {field_name}"
```

`item` is not defined

tests/resources/test_issue.py
```diff
@@ -26,9 +26,12 @@ def test_issue(self):
         self.assertEqual(issue.key, self.issue_1)
         self.assertEqual(issue.fields.summary, "issue 1 from %s" % self.project_b)
 
+        self.assertEqual(issue.get_field('summary'), "issue 1 from %s" % self.project_b)
```
adehad, line 29: Please do not do this. You should not edit all the tests to include this new access method. This does not actually 'test' your code's edge cases.

Create a new function that only tests this `get_field()` function, where you create a new issue and test some example fields, you can test a string field (e.g. summary), a list field (e.g. labels) and the project field and ensure that the type is a `Project`.
This means you are able to verify that you can still get the correct types from various different fields. Including the custom types such as `Project`.


Then you should also create at least one (ideally two) separate function that tests incorrect fields, e.g. a field that start with an underscore - to ensure your attribute error is raised and a field that definitely won't exist.

Tests are not just a box ticking exercise !

adehad (changes requested):
Some works needs to be done on the tests and documentation, but I think this approach is shaping up well

tests/resources/test_issue.py
dvaerum: Thanks for the feedback, I think I have underestimated the effort there have been put into the unit test already made, I will put some more effort into it tomorrow or the day after tomorrow (hopefully)

jira/resources.py
dvaerum: Sry, slipped my mind will get it fixed

jira/resources.py
dvaerum: This one is embarrassing :cold_sweat:

tests/resources/test_issue.py
adehad: No problem ! Feel free to ask me any further questions if anything is unclear and feel free to take your time, we all have stuff going on (and you should enjoy yourself too !)

jira/resources.py
studioj: write a test for it :-) then this wont happen.

> does it show I'm a QA person :-)

studioj (commented):
tests still need some rework

just forget your changes in all the other tests as @adehad mentions

just add two testmethods
one where you raise an arrtibute error on purpose

self.assertRaises(AttributeError, get_field, "_name_starts_with_underscore")

and another one which follows the happy path

rynkk:
Hey, any updates on this?
I am currently stuck with dynamically manipulating the `__class__` of the `_IssueFields`-instance to include my very professional Mixin:
<details>
  <summary>GetItemMixin</summary>

```py

 class GetItemMixin:
    def __getitem__(self, item: str):
        try:
            return getattr(self, item)
        except AttributeError:
            raise ValueError(f"Field {item} does not exist in fields, is it included in the fields parameter?")

    def __setitem__(self, key: str, value: Any):
        if key in self and self[key] is not None and value is not None:
            assert isinstance(value, type(self[key]))  # some help with setting the correct value kind
        setattr(self, key, value)

    def __contains__(self, item):
        try:
            getattr(self, item)
            return True
        except AttributeError:
            return False
```

</details>

dvaerum:
> Hey, any updates on this? I am currently stuck with dynamically manipulating the `__class__` of the `_IssueFields`-instance to include my very professional Mixin:
> GetItemMixin

I ended up getting quite busy until new-year stuff is starting to calm down again. I may be able to have a look at it again because I also need it myself at work. Cannot keep patching this feature in, in multiple scripts :sweat_smile:

tests/resources/test_issue.py
```diff
@@ -26,6 +26,20 @@ def test_issue(self):
...
+            issue.fields.description, issue.get_field(fieldName="description")
+        )
+
+        self.assertRaisesRegex(
+            AttributeError, ": _something", issue.get_field, "_something"
+        )
+
+        self.assertRaisesRegex(
+            AttributeError, "customfield_1234", issue.get_field, "customfield_1234"
+        )
```
adehad, line 41: (non-blocking) I have a preference for the [context manager style](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) as I find it easier to read!

```suggestion
        with self.assertRaisesRegex(AttributeError, ": _something"):
            issue.get_field("_something")

        with self.assertRaisesRegex(AttributeError, "customfield_1234"):
            issue.get_field("customfield_1234")
```

adehad (changes requested):
Awesome, looking much better @dvaerum.

Just a minor point regarding the naming convention of variables.

There is also a non-blocking comment, so feel free to ignore if you feel strongly about it.

jira/resources.py
```diff
@@ -692,6 +692,27 @@ def update(  # type: ignore[override] # incompatible supertype ignored
 
         super(Issue, self).update(async_=async_, jira=jira, notify=notify, fields=data)
 
+    def get_field(self, fieldName: str) -> Any:
```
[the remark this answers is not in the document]
dvaerum: I have changed `fieldName` back to `field_name`, it just looked like the project was using the camelCase naming scheme, so I wanted to follow that 😅 

Example:
https://github.com/pycontribs/jira/blob/5a41be8c3cfb2ed6f090852d28873939641c834f/jira/resources.py#L728-L735

jira/resources.py
```diff
@@ -692,6 +692,27 @@ def update(  # type: ignore[override] # incompatible supertype ignored
...
+        Raises:
+            AttributeError: If the field does not exist or if the field starts with an ``_``
+
+        Returns:
+            Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
+        """
+
+        if fieldName.startswith("_"):
+            raise AttributeError(
+                f"A issue field_name cannot start with underscore (_): {fieldName}",
```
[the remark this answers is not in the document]
dvaerum: fixed

dvaerum:
> Awesome, looking much better @dvaerum.
> 
> Just a minor point regarding the naming convention of variables.
> 
> There is also a non-blocking comment, so feel free to ignore if you feel strongly about it.

Done, I have added all your suggestions, but it doesn't look like it will run the tests :/

jira/resources.py
```diff
@@ -692,6 +692,27 @@ def update(  # type: ignore[override] # incompatible supertype ignored
...
+            Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
+        """
+
+        if fieldName.startswith("_"):
+            raise AttributeError(
+                f"An issue field_name cannot start with underscore (_): {fieldName}",
+                fieldName,
+            )
+        else:
+            return getattr(self.fields, fieldName)
```
adehad, line 714: @dvaerum , yep need to use the new variable name in the function body (and docstring) too:

```suggestion
            field_name (str): The name of the field to get

        Raises:
            AttributeError: If the field does not exist or if the field starts with an ``_``

        Returns:
            Any: Returns the parsed data stored in the field. For example, "project" would be of class :py:class:`Project`
        """

        if field_name.startswith("_"):
            raise AttributeError(
                f"An issue field_name cannot start with underscore (_): {field_name}",
                field_name,
            )
        else:
            return getattr(self.fields, field_name)
```

dvaerum: aaah, so because one fails the rest gets canceled, makes sense, I don't know why I didn't look in the one there was red :joy:

adehad (approved):
Awesome, thanks for getting this through!

dvaerum:
> Awesome, thanks for getting this through!

Thanks for working with me on this :D

PR #1106, opened 2021-07-27
fix bug with accountId is not passed

nakonovalov:
fix "The 'accountId' query parameter needs to be provided"

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -1077,7 +1077,7 @@ class User(Resource):
         session: ResilientSession,
         raw: Dict[str, Any] = None,
     ):
-        Resource.__init__(self, "user?username={0}", options, session)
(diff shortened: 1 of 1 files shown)


adehad:
Unfortunately at the moment this will fail for Jira Server/Data Center, I'll approve the CI run but most likely it will fail.

This is on our radar though, just need to find the best way to go about supporting both Jira Cloud and Jira Server/Data Center.

adehad:
@nakonovalov I've made a separate branch (refer #1109) that I think addresses this, would be great to get confirmation from you that it works for you too.

nakonovalov:
Thank you!

studioj:
> Thank you!

does that mean you checked on your side that it works?

adehad:
Closing in favour of #1109

PR #1107, opened 2021-07-27
allow testing of Jira Cloud

adehad:
## Summary of Changes
1. Create a new pytest mark to opt in to running a test on Jira Cloud
2. Create a new CI that uses GitHub Actions secrets to get credentials to authenticate with Jira Cloud instance - carrying on from the spirit of #909 

## Developer TODO:
- [ ] Port tests relating to 'User' if possible - as that is a key aspect that has changed with GDPR and is affecting users


jira/client.py
```diff
@@ -3945,31 +3952,31 @@ def create_project(
 
         ps_list: List[Dict[str, Any]]
 
-        if not permissionScheme:
+        if permissionScheme is None:
```
studioj, line 3955: while more correct this slightly changes the API => changes are then not only a patch but also a change of functionality. Although I think that people which have an impact by this change were using the library wrong anyway :-)

I also notice these args are not in the docstring 
added https://github.com/pycontribs/jira/issues/1132 for future reference

jira/client.py
```diff
@@ -3945,31 +3952,31 @@ def create_project(
...
             ps_list = self.permissionschemes()
             for sec in ps_list:
                 if sec["name"] == "Default Permission Scheme":
                     permissionScheme = sec["id"]
-                break
-            if not permissionScheme:
+                    break
```
studioj, line 3960: also a bug fix here? seems to be logical to have the break there but again this changes functionality

jira/client.py
```diff
@@ -4042,8 +4051,7 @@ def create_project(
             "key": key,
             "projectTypeKey": ptype,
             "projectTemplateKey": template_key,
-            "lead": assignee,
-            # "leadAccountId": assignee,
+            "leadAccountId" if self._is_cloud else "lead": assignee,
```
studioj, line 4054: cool i didnt know you could change dict keys like this 👍

setup.cfg
```diff
@@ -160,5 +160,8 @@ timeout = 80
 filterwarnings =
     ignore::pytest.PytestWarning
 
+markers =
```
studioj, line 163: how to use this?

tests/conftest.py
```diff
@@ -73,14 +51,29 @@ def setUp(self) -> None:
...
-        self.test_manager = JiraTestManager()
+
+        initialized = False
+        try:
+            self.test_manager = JiraTestManager()
+            initialized = self.test_manager.initialized
+        except Exception as e:
+            # pytest with flaky swallows any exceptions re-raised in a try, except
+            # so we log any exceptions for aiding debugging
+            LOGGER.exception(e)
```
studioj, line 62: shouldn't you raise the exception after catching then?

studioj: that gives also a better test result I'd guess

tests/conftest.py
```diff
@@ -120,20 +113,23 @@ class JiraTestManager(object):
...
 
-    def __init__(self, jira_hosted_type="Server"):
+    def __init__(self, jira_hosted_type=os.environ.get("CI_JIRA_TYPE", "Server")):
         """Instantiate and populate the JIRA instance"""
         self.__dict__ = self.__shared_state
 
         if not self.__dict__:
             self.initialized = False
             self.max_retries = 5
+            self._cloud_ci = False
```
studioj, line 128: I'll voice again my worries ... shouldn't we have 2 classes where we override the differences between cloud and server?
maybe part of a larger refactoring?
The if'ing seems like something easy forgotten in later updates :-)

tests/conftest.py
```diff
@@ -225,6 +196,67 @@ def set_basic_auth_logins(self, **jira_class_kwargs):
             self.jira_sysadmin = JIRA(self.CI_JIRA_URL, **jira_class_kwargs)
             self.jira_normal = JIRA(self.CI_JIRA_URL, **jira_class_kwargs)
 
+    def _project_exists(self, project_key: str) -> bool:
```
studioj, line 199: seems like functionality to add to the library :-)

tests/conftest.py
```diff
@@ -225,6 +196,67 @@ def set_basic_auth_logins(self, **jira_class_kwargs):
...
+                break
+            sleep(wait_between_checks_secs)
+
+        if self._project_exists(project_key):
+            raise TimeoutError(
+                " Project '{project_key}' not deleted after {time_to_wait_for_delete_secs} seconds"
+            )
+
+    def _create_project(
+        self, project_key: str, project_name: str, allow_exist: bool = False
```
studioj, line 242: I'd name this `allow_exist` differently .. had to look up the code as to what you exactly meant by it.
- force_recreate
- remove_if_exists

a few suggestions => everything is subjective of course

tests/resources/test_group.py
```diff
@@ -1,15 +1,22 @@
-from tests.conftest import JiraTestCase
+from tests.conftest import JiraTestCase, allow_on_cloud
 
 
+@allow_on_cloud
 class GroupsTest(JiraTestCase):
+    def setUp(self) -> None:
+        JiraTestCase.setUp(self)
```
studioj, line 7: is there a reason why not to do 

`super().setUp(self)`

studioj (commented):
If CI passes and runs the tests and you take the suggestions into consideration 
you 'll get the thumbs up :-)  👍

jira/client.py
studioj: i didnt find any bug reports related to this...

tests/conftest.py
studioj: should i add a feature request? :-D

adehad:
Thanks for the review! Might get a chance to address them at this tomorrow. It is a bit tricky to have the new Cloud tests run on this PR, I've tried to check it on a fork but it still had some troubles with getting the secrets? Any advice would be appreciated. 
https://github.com/adehad/jira/actions/runs/1109941719

The fork does confirm that the CI is triggered at the right time, ie after the Server CI has been run, which is good to know at least

jira/client.py
adehad: Yep I only discovered it when trying to get the tests to work on the Jira Cloud instance!

setup.cfg
adehad: The `conftest.py` has an `allow_on_cloud` that can be used, see example in `test_group.py`

tests/resources/test_group.py
adehad: No reason, happy to change!

tests/conftest.py
adehad: I was having trouble with raised exceptions not actually causing the test to fail, so instead I log the exception information like this and then do an assertion just below, as the assertions do work correctly. Let me know if this is still unclear and if I should change the comment to explain this better!

tests/conftest.py
adehad: Definitely agree that this isn't optimal, and probably needs some refactoring down the line.
My intention with the `@allow_on_cloud` decorator is that we could leverage the existing tests and port over key tests to run on both without too much code duplication. So the only time you would to do the if else checks would be when we are supporting both in a test case and there is a key difference between Server and Cloud that needs accounting.

studioj:
> I've tried to check it on a fork but it still had some troubles with getting the secrets? Any advice would be appreciated.
> https://github.com/adehad/jira/actions/runs/1109941719

i think you can create your own jira cloud instance, I did so a few months back ... have you tried running it against your own instance already?

adehad:
> > I've tried to check it on a fork but it still had some troubles with getting the secrets? Any advice would be appreciated.
> > https://github.com/adehad/jira/actions/runs/1109941719
> 
> i think you can create your own jira cloud instance, I did so a few months back ... have you tried running it against your own instance already?
> 

I have and it passes locally, see: https://github.com/pycontribs/jira/issues/896#issuecomment-889475886

studioj:
Getting my new local windows env up and running is not really trivial... I'll maybe try to make a contributor guide for windows.

I'll start with trying to understand how tox works and is configured. For now it's mere magic 😅.

tests/conftest.py
adehad: force_recreate sounded good to me

tests/conftest.py
adehad: go for it !

adehad:
So I've done some more testing on my fork and I think GitHub Actions isn't suitable for running actions with shared secrets.

Main reason is that it does not have any mechanism to share secrets with pull requests. 
If you try to setup the same secrets in your fork, it will still scrub about the secrets to an empty string before they can be used (probably because of the matching name).

The other option I am pursuing right now is using Azure Pipelines, which has a handy feature where we can comment on a pull request to be able to trigger the pipeline. The good thing about this is we can control who's comments to actually listen to to be able to trigger the pipeline. This means @studioj and myself for example can be given permissions to do this. Which allows us to vet the pull requests before we run any pipelines on them.
https://docs.microsoft.com/en-gb/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers 
The use case I imagine is that, we verify that the Server tests pass and that there is nothing malicious in the code, we can then trigger the pipeline via a comment. The pipeline status is then reported directly onto the pull request.


The ideal situation would be if GitHub implements an "Approve and Run" button for Actions which enable sharing of secrets. This has been discussed for a long time, but it sounds like it hasn't gotten anywhere unfortunately, e.g.: https://github.community/t/allow-secrets-to-be-shared-with-forks-from-trusted-actions/16525/12 . For now I think the Azure Pipeline option might be best.

tests/conftest.py
studioj: what i would do though is catch a more specific exception here. Which counts for all down the line actually.
to me it feels like JIRAError
this 'No project could be found' seems to be coming from jira itself ... 
similarily i would add a raise in an `else` statement if that string was not found

studioj: if now you have a connection error it would return True

adehad:
@studioj can you submit a pull request to my fork? I think I've gotten azure pipelines to work:
https://github.com/adehad/jira/pull/3
i.e. I can use a comment to trigger the pipeline.

So if you can fork my fork and then submit a pull request to my fork, I think if you write the comment it won't do anything. But when I write it, then it will trigger the pipeline

tests/conftest.py
adehad: now updated, once instance seems like it should be a genuine case, so I've left it.

PR #1109, opened 2021-07-29
jira cloud user support

adehad:


studioj (approved):
everything seems understandable and correct...

what worries me architecturally is that maybe we should split up the cloud vs the server part into two classes ... 
this seems like a patch for a much larger problem... (i'm guessing more problems will pop up)

I'll check out your code to play around a little bit (y)

I'd be willing to help contribute to this refactoring... (but we should plan it a bit)

adehad:
> everything seems understandable and correct...
> 
> what worries me architecturally is that maybe we should split up the cloud vs the server part into two classes ... 
> this seems like a patch for a much larger problem... (i'm guessing more problems will pop up)
> 
> I'll check out your code to play around a little bit (y)
> 
> I'd be willing to help contribute to this refactoring... (but we should plan it a bit) 

Agreed, for a 4.0 or other major version release I think it makes sense to revisit how we do this and implement a more scalable solution

kot0dama:
Hi,

I think delete_user() also needs an update as it's not working with emails anymore on JIRA Cloud.

I got a workaround for now:
- fetch the user's accountId through search_user() 
- pass it to delete_user() along with the username (something like `"{username}&accountId={accountId}"`)

Let me know if I can help in any way!

Thank you,
Loïc

adehad:
Hi @kot0dama it would be great if you could raise an issue for this - (I would use the bug template). Would also be great for you to include the workaround you have in case this is useful for anyone else in the meantime.

We are happy to review any pull requests you have too.

If you do want to take a crack at some code this kind of behaviour we currently handle with a `self_is_cloud` property

So perhaps we could do something like `"{'accountId' if self._is_cloud else 'username'}={username}"`
We can then specify in the docstring that the `username` argument should be the `accountId` in Jira Cloud.

kot0dama:
Hi @adehad, here's the bug report including my workaround:  #1241 
I'll followup on your comment there !

Thank you

PR #1110, opened 2021-07-31
fix to Client Class which no longer accepts the parameter "user" and instead requires accountid to find users by account id

mangopipeline:
this doesn't work due to changes on the restApi...
I'm not sure if this is the right fix, but it was the only way I could get jire.user(user_id) to work...
let me know what you guys think

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -1077,7 +1077,7 @@ class User(Resource):
         session: ResilientSession,
         raw: Dict[str, Any] = None,
     ):
-        Resource.__init__(self, "user?username={0}", options, session)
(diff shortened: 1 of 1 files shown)


adehad:
Hi @mangopipeline, thanks for this. Could you check if the branch #1109 works for you? 

This pull request currently won't work with our CI as we test on Jira Server too, you can verify this by looking at #1106 - where this exact change went through a CI run.

studioj:
@mangopipeline did you have time to verify @adehad s suggestions. 
If they work for you please close the PR.

Thank you for contributing!

adehad:
Closing in favour of #1109

PR #1113, opened 2021-08-03
Paginate result lists as a generator

turran:


adehad:
Can you elaborate on what this is and how it should be used? From my initial quick look I'm struggling to see what problem this is trying to solve.

jira/client.py
```diff
@@ -4254,7 +4299,9 @@ def boards(
                 Board(self._options, self._session, raw_boards_json)
                 for raw_boards_json in r_json["views"]
             ]
-            return ResultList(boards, 0, len(boards), len(boards), True)
+            return ResultList(
+                Board, None, boards, None, None, None, 0, len(boards), len(boards), True
```
studioj, line 4303: thats a  lot of parameters ....

jira/client.py
```diff
@@ -136,7 +136,12 @@ def _field_worker(
 class ResultList(list, Generic[ResourceType]):
     def __init__(
```
studioj, line 137: can we do this without this large extension of parameters? 
Wherever you're calling this it doesnt make too much sense as to what is happening

turran: I have reopened this PR and made the changes that you have requested. A new class holding the parameters is now available. More refactoring can be done, like the _get_items_from_pages or the ResultList with useless parameters, but I have no clear view on hoe to proceed.

adehad: @turran when writing this implementation do you have any thoughts on what would make a better user experience?

I have been wondering if we can have a look at options for breaking changes in a major version bump, and am considering if we can perhaps use httpx and move to true async support. 

But perhaps that would be a separate package. We've inherited a lot of baggage so might also be worth considering what a clean slate might gain.

turran: @adehad Good point. IMHO when building a library you need to define where you put the criteria of what the library can do for the user and what not. So far, the current Jira Python bindings aim to mimic as much as possible the REST API and doing synchronous calls for that, at the HTTP level. There are plenty of situations where the current 1:1 python - REST API can be abstracted to give a better user experience, but that's not the current situation.

Still, the Jira REST API itself can be batched and already has asynchronous operations internally, like the UPM interface where you get a task id and then the user queries an specific endpoint with that task id to know the status of it. This can be abstracted on the python library again as a 1:1 with sync calls or put some logic above to do the task handling automatically for you, but that would require async code for sure and httpx seems like a good option there.

Now, for APIs that support batches, like searching issues by JQL, yielding them is enough and won't require much heavy change on the current code. At the end, the REST API that work with batches behaves like a generator already :)

adehad: I think I'll need to see the ergonomics of this through an example to really make a call on whether this should exist as code in the library or maybe in the documentation as an example of optimizing / improving performance.

So if you still have some capacity to write a test that would be ideal. I would perhaps make a new `tests/test_paginator.py` and use an endpoint you deem suitable for the example.

As you say the `tests/resources/` is probably a good place to see the new tests structure. I think the `search_issues()` would be a good one to make an example for as that is where I can imagine most people trying to leverage this. You can try searching Atlassian's own Jira https://jira.atlassian.com/secure/Dashboard.jspa as they'll probably have much more tickets, and you won't need to authenticate to simply trawl through.


---- 

Thinking out loud

Maybe we can provide a `JiraAsync` class that uses a httpx session under the hood (vs a requests session) to start a process moving. A bit intimidating to look at all the functionality we support. But perhaps we can use the async class internally in the non-async class and simply have the non-async methods await the async methods. So users familiar with the sync calls could still benefit from async-ing the paginated endpoints for example.



----

Thanks again for taking the time for the PR and suggestions, hope we can figure out something that would work, and sorry that I'm a bit overly cautious for some changes !

studioj:
stale for 4 months => close for now

PR #1139, opened 2021-09-02
add PermissionScheme Resource

Steffen223:
make PermissionScheme Resource also available when using pycontribs/jira.
tox lint passed. But I had issues with my environment using tox tests. Using this branch in a our script worked successful

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -83,6 +83,7 @@ from jira.resources import (
     User,
     Version,
     Votes,
+    PermissionScheme,
     Watchers,
     Worklog,
 )
@@ -2018,6 +2019,15 @@ class JIRA(object):
             Votes
         """
         return self._find_for_resource(Votes, issue)
+
+    @translate_resource_args
+    def project_permissionscheme(self, project):
+        """Get a PermissionScheme Resource from the server.
+ 
+        :param project: ID or key of the project to get the permissionscheme for
+        :rtype: PermissionScheme
+        """
+        return self._find_for_resource(PermissionScheme, project)
 
     @translate_resource_args
     def add_vote(self, issue: str) -> Response:
--- a/jira/client.py
+++ b/jira/client.py
@@ -2024,8 +2024,12 @@ class JIRA(object):
     def project_permissionscheme(self, project):
         """Get a PermissionScheme Resource from the server.
 
-        :param project: ID or key of the project to get the permissionscheme for
-        :rtype: PermissionScheme
+
+        Args:
+            project (str): ID or key of the project to get the permissionscheme for
+
+        Returns:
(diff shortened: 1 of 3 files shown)


jira/client.py
```diff
@@ -2019,6 +2020,15 @@ def votes(self, issue: str) -> Votes:
         """
         return self._find_for_resource(Votes, issue)
 
+    @translate_resource_args
+    def project_permissionscheme(self, project):
+        """Get a PermissionScheme Resource from the server.
+
+        :param project: ID or key of the project to get the permissionscheme for
+        :rtype: PermissionScheme
+        """
```
adehad, line 2029: Let's use the Google style docstring 

```suggestion

        Args:
            project (str): ID or key of the project to get the permissionscheme for

        Returns:
        PermissionScheme: The permission scheme
        """
```

jira/client.py
```diff
@@ -2019,6 +2020,15 @@ def votes(self, issue: str) -> Votes:
         """
         return self._find_for_resource(Votes, issue)
 
+    @translate_resource_args
+    def project_permissionscheme(self, project):
```
adehad, line 2024:
```suggestion
    def project_permissionscheme(self, project: str):
```

jira/resources.py
```diff
@@ -798,6 +799,17 @@ def __init__(
...
+class PermissionScheme(Resource):
+    """Permissionscheme information on an project."""
+
+    def __init__(self, options, session, raw=None):
+        Resource.__init__(
+            self, "project/{0}/permissionscheme?expand=user", options, session
+        )
+        if raw:
+            self._parse_raw(raw)
+
```
adehad, line 811:
```suggestion
        self.raw: Dict[str, Any] = cast(Dict[str, Any], self.raw)
```

adehad (changes requested):
looking good ! Tests can be a bit tricky as you need to setup docker, but if you can try something locally feel free to add a test file to the `test/resources/` folder and we can see how it runs on the CI

Steffen223:
Hello @adehad , a test has been added to https://github.com/Steffen223/jira/blob/master/tests/resources/test_project.py#L204

adehad:
Test failure is unrelated to this, seems like we need to reconfigure something on our github actions end. Will try and sort this out and then re-run the jobs.

adehad:
Looks like a few linting error popped up could you run `tox -e lint` ?
Can you also first rebasing / merge master back in? As we just merged in an update to our lint

Steffen223:
Hello @adehad , I linted, updated and merged from master.

adehad (approved):

adehad:
Awesome, thank you for your contribution !

PR #1145, opened 2021-09-17
Locate the exact user by key if there are multiple users returned from query

skyscribe:
Previously only the first user is taken when an issue is assigned, things would stop working when multiple user records are returned from server, and intended user is not listed at the head of result list.

A second filter is applied if multiple user records are returned, so if there is an exact match, it should be returned instead of the first one in original query list.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1663,7 +1663,7 @@ class JIRA(object):
                 users = self.search_users(query=user)
             else:
                 users = self.search_users(user=user)
-            
+
             if len(users) > 1:
                 matches = [u for u in users if u.key == user]
             user_obj = matches[0] if matches else users[0]
--- a/jira/client.py
+++ b/jira/client.py
@@ -1666,12 +1666,7 @@ class JIRA(object):
             
             if len(users) > 1:
                 matches = [u for u in users if u.key == user]
-                if len(matches) >= 1:
-                    user_obj = matches[0]
-                else:
-                    user_obj = users[0]               
-            else:
-                user_obj = users[0]
+            user_obj = matches[0] if matches else users[0]
 
         except Exception as e:
             raise JIRAError(str(e))
--- a/jira/client.py
+++ b/jira/client.py
@@ -1661,16 +1661,18 @@ class JIRA(object):
             user_obj: User
             if self._is_cloud:
                 users = self.search_users(query=user)
-                if len(users) > 1:
-                    matches = [u for u in users if u.key == user]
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -1660,9 +1660,19 @@ def _get_user_id(self, user: str) -> str:
...
             user_obj: User
             if self._is_cloud:
-                user_obj = self.search_users(query=user, maxResults=1)[0]
+                users = self.search_users(query=user)
             else:
-                user_obj = self.search_users(user=user, maxResults=1)[0]
+                users = self.search_users(user=user)
+            
+            if len(users) > 1:
+                matches = [u for u in users if u.key == user]
```
adehad, line 1668: I wonder if we want a different fallback property for jira cloud, as `key` is deprecated there.

jira/client.py
```diff
@@ -1660,9 +1660,19 @@ def _get_user_id(self, user: str) -> str:
...
+                if len(matches) >= 1:
+                    user_obj = matches[0]
+                else:
+                    user_obj = users[0]               
+            else:
+                user_obj = users[0]
```
adehad, line 1674:
```suggestion
            user_obj = matches[0] if matches else users[0]
```

and `matches = []` before the `if len(users) > 1`

jira/client.py
```diff
@@ -1660,9 +1660,19 @@ def _get_user_id(self, user: str) -> str:
         try:
             user_obj: User
             if self._is_cloud:
-                user_obj = self.search_users(query=user, maxResults=1)[0]
+                users = self.search_users(query=user)
             else:
-                user_obj = self.search_users(user=user, maxResults=1)[0]
```
adehad, line 1665: I'm a bit concerned about removing the `maxResults`, can we set it to a reasonable default still?

adehad (changes requested):
Looks mostly fine to me. Can I confirm the situation is: you are searching by the exact key and you aren't getting it from the first result?

jira/client.py
skyscribe: I have only tested on my jira cloud environment, not sure about the newer properties yet.

jira/client.py
skyscribe: Reset as 20 - this should be reasonable still? @adehad

adehad: yep

studioj (approved):

adehad (approved):

PR #1157, opened 2021-09-23
Fix JWT QSH generation for urls with repeated parameters

robinhughes:
- Fix QshGenerator to combine repeated query string parameters into csv
- add test for QshGenerator.generate_qsh

This fixes an issue where e.g. `search_issues(jql, fields=["field1", "field2"])` would fail to authenticate because the generated qsh was invalid. The url requested is like `...&fields=field1&fields=field2&....` These repeated parameters must be combined into a single csv parameter according to https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#qsh when computing the QSH for JWT auth.

### DIFF
--- a/tests/test_qsh.py
+++ b/tests/test_qsh.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from jira.client import QshGenerator
+
+
+class MockRequest(object):
+    def __init__(self, method, url):
+        self.method = method
+        self.url = url
+
+
+def test_qsh():
+    gen = QshGenerator("http://example.com")
+
+    for method, url, expected in [
+        ("GET", "http://example.com", "GET&&"),
+        # empty parameter
+        ("GET", "http://example.com?key=&key2=A", "GET&&key=&key2=A"),
+        # whitespace
+        ("GET", "http://example.com?key=A+B", "GET&&key=A%20B"),
+        # tilde
+        ("GET", "http://example.com?key=A~B", "GET&&key=A~B"),
+        # repeated parameters
+        (
+            "GET",
+            "http://example.com?key2=Z&key1=X&key3=Y&key1=A",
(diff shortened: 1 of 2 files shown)


tests/test_qsh.py
```diff
@@ -0,0 +1,37 @@
...
+class MockRequest(object):
+    def __init__(self, method, url):
+        self.method = method
+        self.url = url
+
+
+def test_qsh():
+    gen = QshGenerator("http://example.com")
+
+    for method, url, expected in [
```
adehad, line 14: Awesome test cases, let's arrange them into a pytest friendly way !

Specifically with a [`@pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html)

An example can be found here:
https://github.com/pycontribs/jira/pull/1109/files#diff-3ce874fa74c2993e05bfbc5fbc2df93e0d043ab22ae5dd0c1f7c4f0ce56e4216R13

So you would just be pulling out the for loop into the function argument.

```python
@pytest.mark.parametrize(
    "method,url,expected",
     [("GET", "http://example.com", "GET&&"), ...],
    ids=["empty parameter", ...]
)
def test_qsh(method, url, expected): ...
```

adehad (changes requested):
Looking good! There's a problem with our github actions pipeline at the moment so it may fail, I'll try and sort this out soon so we can get this merged.

tests/test_qsh.py
robinhughes: Sure! I've made the suggested change.

adehad:
Ah lint failing might be because we've updated our lint steps, try merging in master before running `tox -e lint` again !

adehad (approved):

adehad:
Thank you for your contribution !

Issue #1164, opened 2021-10-04
Not able to create a project


hsekowski:
**Describe the bug**
create_project function does not work as expected in this version
It worked in 2.0.0

**To Reproduce**
```
from jira import JIRA

jira = JIRA("http://localhost:8080/", basic_auth=("admin", "admin"))
project_key = 'PRJ'
project = jira.create_project(project_key)
```
1. Any additional steps or considerations that happen before or after.

**Expected behavior**

**Stack Trace**
```python
✗ python xyz/__init__.py
Traceback (most recent call last):
  File "xyz/__init__.py", line 35, in <module>
    project = jira.create_project(project_key)
  File "/Users/abc/Library/Caches/pypoetry/virtualenvs/issue-exploder-J4Ji3stn-py3.7/lib/python3.7/site-packages/jira/client.py", line 3775, in create_project
    r = self._session.post(url, data=json.dumps(payload))
  File "/Users/abc/Library/Caches/pypoetry/virtualenvs/issue-exploder-J4Ji3stn-py3.7/lib/python3.7/site-packages/jira/resilientsession.py", line 175, in post
    return self.__verb("POST", url, **kwargs)
  File "/Users/abc/Library/Caches/pypoetry/virtualenvs/issue-exploder-J4Ji3stn-py3.7/lib/python3.7/site-packages/jira/resilientsession.py", line 168, in __verb
    raise_on_error(response, verb=verb, **kwargs)
  File "/Users/abc/Library/Caches/pypoetry/virtualenvs/issue-exploder-J4Ji3stn-py3.7/lib/python3.7/site-packages/jira/resilientsession.py", line 54, in raise_on_error
    r.status_code, error, r.url, request=request, response=r, **kwargs
jira.exceptions.JIRAError: JiraError HTTP 400 url: http://localhost:8080/rest/api/2/project
	text: You must specify a valid project lead.
	
[...]

PR #1165, opened 2021-10-04
fix: https://github.com/pycontribs/jira/issues/1164

hsekowski:
current_user code backported from 2.0.0
Works fine for on-prem create_project (not tested for cloud)

Other occurrences not tested as less important from functional point of view:
1) logging in maintenance.py
2) conftest.py & test_component.py

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3778,22 +3778,20 @@ class JIRA:
             self.log.error(ioe)
         return None
 
-    def current_user(self, field: str = "key") -> str:
-        """Returns the username or emailAddress of the current user. For anonymous
-        users it will return a value that evaluates as False.
+    def current_user(self):
+        if not hasattr(self, '_serverInfo') or 'username' not in self._serverInfo:
 
-        Returns:
-            str
-        """
-        if not hasattr(self, "_myself"):
+            url = self._get_url('serverInfo')
+            r = self._session.get(url, headers=self._options['headers'])
 
-            url = self._get_url("myself")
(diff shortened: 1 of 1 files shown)


hsekowski:
Checking and comparing REST API documentation for both Server and Cloud, it's obvious we have to know if call is to one or another. As an example - server accepts lead as a call parameter (https://docs.atlassian.com/software/jira/docs/api/REST/8.19.1/#project-createProject) while cloud requires leadAccountId (https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-projects/#api-rest-api-3-project-post).

adehad:
Can you tell me if the branch: https://github.com/pycontribs/jira/tree/feature/test_jira_cloud

works for you?

The only thing that might need changing is in the call to current_user() where we may want to specify the field as `accountId` instead of the default `key`

https://github.com/pycontribs/jira/blob/4617bb82e972379f83a26bf6853b83061b6a6a8e/jira/client.py#L3989

hsekowski:
@adehad  ,
Is the branch ( https://github.com/pycontribs/jira/tree/feature/test_jira_cloud ) intended to be for cloud only ?
I tested it for on prem and get
`response text = {"errorMessages":[],"errors":{"projectLead":"You must specify a valid project lead."}}`

adehad:
@hsekowski it is intended to support both. The plan is to merge it into main, it specifically adds compatibility for Jira Cloud by addressing some regressions.

Issue #339, opened 2017-02-10
Passing certificates from trusted CAs


hunsalz:
If your Jira runs in an environment with your own trusted CA you will have to instantiate jira with passing 'verify' the path of a CA_BUNDLE file or directory with certificates of trusted CAs. See http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

```
options = {
   "server" : 'https://your-jira.your-domain.org',
   "verify" : 'your-ca.crt'
}
jira = JIRA(options, basic_auth=('user', 'password')
```

Otherwise you will receive:
```
requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:749)
```

Unfortunately it doesn't work with the config.py module as well. The reason is the boolean assumption in line 85
```
verify = config.getboolean(profile, 'verify')
```

Simple working solution is:
```
verify = config.get(profile, 'verify')
if verify.lower() == 'true':
   verify = True
elif verify.lower() == 'false':
   verify = False
```
Maybe the boolean decision is imaginable in a more pythonic way ...?

ssbarnea:
Dealing with certificates is outside the scope of this library. We are using requests library so you can use it to change [SSL certificate bundle via REQUESTS_CA_BUNDLE](http://docs.python-requests.org/en/master/user/advanced/).

You are more than welcome to chaise all SSL related bugs directly on [requests library issue tracker](https://github.com/kennethreitz/requests/issues?q=CA_BUNDLE).

hunsalz:
@ssbarnea : I suppose you misunderstood me. I don't suggest to add any new SSL functionality. It's about the misinterpretation of the 'verify' config param in config.py module. The verify parameter isn't strictly a boolean value. In fact it's allowed to pass a path to a directory with certificates or a path to a single certificate and this will be processed by OpenSSL.

PR #1202, opened 2021-10-29
add the `.verify` property correctly

adehad:
Addressed the concerns raised by #339 

http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

`pip install git+https://github.com/pycontribs/jira.git@bugfix/correctly_add_ssl_cert`

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -378,7 +378,7 @@ class JIRA:
                 * rest_api_version -- the version of the REST resources under rest_path to use. Defaults to ``2``.
                 * agile_rest_path - the REST path to use for Jira Agile requests. Defaults to ``greenhopper`` (old, private
                   API). Check :py:class:`jira.resources.GreenHopperResource` for other supported values.
-                * verify -- Verify SSL certs. Defaults to ``True``.
+                * verify (Union[bool, str]) -- Verify SSL certs. Defaults to ``True``.
                 * client_cert -- a tuple of (cert,key) for the requests library for client side SSL
                 * check_update -- Check whether using the newest python-jira library version.
                 * headers -- a dict to update the default headers the session uses for all API requests.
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -3403,6 +3395,13 @@ def _add_client_cert_to_session(self):
         client_cert: Tuple[str, str] = self._options["client_cert"]  # to help mypy
         self._session.cert = client_cert
 
+    def _add_ssl_cert_to_session(self):
+        """
+        Adds the client certificate to the request if configured through the constructor.
```
kchason, line 3400: Is this adding the client certificate, or the certificate authority?

adehad: Thanks, this was a copy paste job I failed to update. I've now updated both the name of the function and improved the docstring for clarity.

jira/client.py
```diff
@@ -480,7 +483,6 @@ def __init__(
             self._create_oauth_session(oauth, timeout)
         elif basic_auth:
             self._create_http_basic_session(*basic_auth, timeout=timeout)
-            self._session.headers.update(self._options["headers"])
```
studioj, line 483: i see this is removed but not added somewhere else ....

studioj (changes requested):
I would maybe look for adding a test for this... this has the potential to break a lot :-) and probably will sneak through CI without a proper test :-S. I wouldn't know how to test it though :-D

jira/client.py
adehad: I removed this as it is already called in L502 (current main) regardless of the auth session type:

https://github.com/pycontribs/jira/blob/db56d5c98dd5a792e6a347c3ff87f7f8cc752599/jira/client.py#L499-L502

adehad:
>I would maybe look for adding a test for this... this has the potential to break a lot :-) and probably will sneak through CI without a proper test :-S. I wouldn't know how to test it though :-D

Yeah I have no idea how to test this, really the `ResilientSession` class should accept these arguments as parameters and set it during the `__init__`.

studioj (approved):

studioj:
> > I would maybe look for adding a test for this... this has the potential to break a lot :-) and probably will sneak through CI without a proper test :-S. I wouldn't know how to test it though :-D
> 
> Yeah I have no idea how to test this, really the `ResilientSession` class should accept these arguments as parameters and set it during the `__init__`.

:-) see it as a challenge :D you have my approval...

adehad:
>:-) see it as a challenge :D

😆

PR #1238, opened 2021-11-13
Allow verify option as path to cert in config

lamcw:
This aligns behavior in `get_jira` with the feature implemented in #1202 :

```ini
[my-jira-profile]
url = abc.atlassian.net
verify = /path/to/cert
```


adehad (approved):
thanks, good catch

PR #1261, opened 2021-12-09
Remove unnecessary shebang lines

csomh:
None of these files are marked or installed as executables, so having a
shebang line is not required.

This will simplify Fedora packaging. Fedora Packaging Guidelines require
files which are not installed as executables to not to have shebang
lines.
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shebang_lines

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 """
 This module implements a friendly (well, friendlier) interface between the raw JSON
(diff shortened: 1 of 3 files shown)


studioj (approved):
i dont see any issues with this change

ssbarnea (approved):

Issue #723, opened 2018-11-15
Get statuses that relates to a specific project


aaron417:
**Is your feature request related to a problem? Please describe.**
The problem I meet is I'd like to get "statuses" configuration that belongs to a specific project like the REST API defined in JIRA rest api doc

GET /rest/api/2/project/{projectIdOrKey}/statuses

**Describe the solution you'd like**
a new API in jira like "project_statuses(id)"

**Describe alternatives you've considered**
None

**Additional context**
None

wilada:
+1 - I've also tried to get this by using `find('project/{projectKey}/statuses')`, but it appears `_parse_raw` assumes the raw json is always a dictionary, and not a list. From jirashell:
```
~/.virtualenvs/a/lib/python3.7/site-packages/jira/client.py in find(self, resource_format, ids)
    671         """
    672         resource = Resource(resource_format, self._options, self._session)
--> 673         resource.find(ids)
    674         return resource
    675 

~/.virtualenvs/a/lib/python3.7/site-packages/jira/resources.py in find(self, id, params)
    199             path = self._resource.format(id)
    200         url = self._get_url(path)
--> 201         self._load(url, params=params)
    202 
    203     def _get_url(self, path):

~/.virtualenvs/a/lib/python3.7/site-packages/jira/resources.py in _load(self, url, headers, params, path)
    322         if path:
    323             j = j[path]
--> 324         self._parse_raw(j)
    325 
    326     def _parse_raw(self, raw):

~/.virtualenvs/a/lib/python3.7/site-packages/jira/resources.py in _parse_raw(self, raw)
    328         if not raw:
    329             raise NotImplementedError("We cannot instantiate empty resources: %s" % raw)
--> 330         dict2resource(raw, self, self._options, self._session)
    331 
    332     def _default_headers(self, user_headers):

~/.virtualenvs/a/lib/python3.7/site-packages/jira/resources.py in dict2resource(raw, top, options, session)
    921 
    922     seqs = tuple, list, set, frozenset
--> 923     for i, j in iteritems(raw):
    924         if isinstance(j, dict):
    925             if 'self' in j:

~/.virtualenvs/a/lib/python3.7/site-packages/six.py in iteritems(d, **kw)
    585 
    586     def iteritems(d, **kw):
--> 587         return iter(d.items(**kw))
    588 
    589     def iterlists(d, **kw):

AttributeError: 'list' object has no attribute 'items'
```

wilada:
[...]

PR #1267, opened 2021-12-31
Add projectstatuses method to get project statuses (missing endpoint)

Kayx23:
Added `projectstatuses` method for `/project/{projectIdOrKey}/statuses`

Relevant issue: https://github.com/pycontribs/jira/issues/723

Atlassian Doc for this REST endpoint: 
* [Jira Cloud REST v3](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-projects/#api-rest-api-3-project-projectidorkey-statuses-get)
* [Jira Cloud REST v2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-projects/#api-rest-api-2-project-projectidorkey-statuses-get)
* [Jira Server](https://docs.atlassian.com/software/jira/docs/api/REST/8.13.15/#project-getAllStatuses)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2933,7 +2933,7 @@ class JIRA:
         Returns:
             List[Status]
         """
-        r_json = self._get_json("project/{projectIdOrKey}/statuses")
+        r_json = self._get_json(f"project/{projectIdOrKey}/statuses")
         statuses = [
             Status(self._options, self._session, raw_stat_json)
             for raw_stat_json in r_json
--- a/jira/client.py
+++ b/jira/client.py
@@ -2916,7 +2916,7 @@ class JIRA:
     def statuses(self) -> List[Status]:
         """Get a list of all status Resources from the server.
 
-        Refer to :py:meth:`JIRA.issue_types_for_project` for getting statuses 
+        Refer to :py:meth:`JIRA.issue_types_for_project` for getting statuses
         for a specific issue type within a specific project.
 
         Returns:
--- a/jira/client.py
+++ b/jira/client.py
@@ -2927,18 +2927,18 @@ class JIRA:
         ]
         return statuses
 
-    def project_statuses(self, projectIdOrKey: str) -> List[Status]:
-        """Get a list of statuses available within a project from the server.
+    def project_statuses_by_issue_type(self, projectIdOrKey: str) -> List[IssueType]:
+        """Get a list of issue types available within the project with available statuses within each issue type, as each project has a set of valid issue types and each issue type has a set of valid statuses.
 
         Returns:
-            List[Status]
+            List[IssueType]
         """
         r_json = self._get_json(f"project/{projectIdOrKey}/statuses")
-        statuses = [
-            Status(self._options, self._session, raw_stat_json)
+        issue_types = [
+            IssueType(self._options, self._session, raw_stat_json)
             for raw_stat_json in r_json
         ]
-        return statuses
+        return issue_types
 
     def status(self, id: str) -> Status:
         """Get a status Resource from the server.
--- a/jira/client.py
+++ b/jira/client.py
@@ -2916,6 +2916,9 @@ class JIRA:
     def statuses(self) -> List[Status]:
         """Get a list of all status Resources from the server.
 
+        Refer to :py:meth:`JIRA.issue_types_for_project` for getting statuses 
+        for a specific issue type within a specific project.
+
         Returns:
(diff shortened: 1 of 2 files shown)


jira/client.py
```diff
@@ -2927,6 +2927,19 @@ def statuses(self) -> List[Status]:
         ]
         return statuses
 
+    def projectstatuses(self, projectIdOrKey: str) -> List[Status]:
```
adehad, line 2930: (non-blocking)

@studioj what are you thoughts on using the more conventional naming convention here?

e.g.:

```suggestion
    def project_statuses(self, projectIdOrKey: str) -> List[Status]:
```

My only concern is that we don't really conform to this anyway, e.g. `statuscategories()` below. So perhaps we want to have a separate merge request where we add a deprecation warning to the current non snake_case methods, and also convert them to a wrapper of the snake case ones. Something like:

```python

def projectstatuses(self, projectIdOrKey: str) -> List[Status]:
    warnings.warn("Please use project_statuses()", DeprecationWarning)
    return self.project_status(projectIdOrKey)
   

def project_statuses(self, projectIdOrKey: str) -> List[Status]:
    ... 
```

adehad (approved):
Looks good to me ! Will wait a bit too see if @studioj spots anything I've missed.

Thanks !

jira/client.py
studioj: i would indeed do that in a separate PR, but tbh if we add something new i would suggest making it snake case yes

studioj: so this method should be snake case
project_statuses

studioj (changes requested):
cool thank you for your addition.
We do prefer PR's where there is a test added so the functionality can keep on running properly

Kayx23:
Apologize for the back and forth... Please review again. I realized I misunderstood what the endpoint returns when writing the test case. I thought it returns all statuses within a project, but it actually returns issue types with statuses within each issue type. I've made some revisions. 

CI test for servers passed... cloud failed because `Test Manager setUp failed`. Was this expected?

studioj:
> CI test for servers passed... cloud failed because `Test Manager setUp failed`. Was this expected?

seems like this is for all PR's i hope @adehad  will have the time to check... i'm not that knowledgeable with the cloud CI part. it 

seems like an ENV var isn't set. That part has been added pretty recent so maybe it still contains some edge cases

adehad:
Sorry yes regarding the Cloud CI, currently it doesn't play nicely with merge requests from forks (it doesn't pass the credentials properly even though we think we've set all the right settings to allow this!)

@studioj from my perspective if the Server tests pass that is good, then on our end we can raise a PR from a branch within this repo and we should be able to activate this test to run on Cloud with the `@allow_on_cloud` decorator, and hopefully we should see no problems and can merge it.

studioj:
It's a start😀
Better late than not and of course safety first

jira/client.py
```diff
@@ -2927,6 +2927,19 @@ def statuses(self) -> List[Status]:
         ]
         return statuses
 
+    def project_statuses_by_issue_type(self, projectIdOrKey: str) -> List[IssueType]:
+        """Get a list of issue types available within the project with available statuses within each issue type, as each project has a set of valid issue types and each issue type has a set of valid statuses.
```
adehad, line 2931: Hi @Kayx23 !

Sorry I'm just reviewing this again and think this function name and docstring could be clarified.

Technically this function is returning issue types rather than project statuses.

So perhaps the function name should be `issue_types_for_project()`

I also think the short summary for the function could be made clearer with an example.

Perhaps we can modify this like so:

```suggestion
    def issue_types_for_project(self, projectIdOrKey: str) -> List[IssueType]:
        """Get a list of issue types available within the project.
        
         Each project has a set of valid issue types and each issue type has a set of valid statuses.
         The valid statuses for a given issue type can be extracted via: `issue_type_x.statuses`
```

We can then probably also update the docstring of `statuses()` to point to this function to help people find this use case!

```python
        """Get a list of all status Resources from the server.

           Refer to :py:meth:`JIRA.issue_types_for_project` for getting statuses
           for a specific issue type within a specific project.
```

adehad (changes requested):

jira/client.py
Kayx23: Good suggestion! This makes a lot of sense. I've added them in.

studioj (approved):

PR #1276, opened 2022-01-11
Support requesting board configuration Fixes #1242

### LINKED ISSUE #1242: Support board issue endpoint
### Problem trying to solve

With the old greenhopper API, I could get a board, find the filter used for the board, and then find the issues. I see no way to do that since the 3.1.1 release.

### Possible solution(s)

There seems to be an endpoint that does exactly what I need, but it's not implemented as far as I can see.

https://docs.atlassian.com/jira-software/REST/7.3.1/#agile/1.0/board-getIssuesForBoard

### Alternatives

I tried to work around this with a JQL search, but it seem I can't query issues within a certain board.

### Additional Context

_No response_

jswetzen:
It's previously not possible to get configuration (and hence the JQL) for a board. This PR adds that new resource and supports fetching configuration by board ID using the `/rest/agile/1.0/board/{boardId}/configuration` endpoint.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -58,6 +58,7 @@ from jira.resilientsession import ResilientSession, raise_on_error
 from jira.resources import (
     Attachment,
     Board,
+    BoardConfiguration,
     Comment,
     Component,
     Customer,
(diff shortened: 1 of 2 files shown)


studioj (changes requested):
please add a test to complete this PR we really want to only move tested code in. Thank you

PR #1290, opened 2022-02-02
Remove unused setuptools runtime dependency

### LINKED ISSUE #1289: Remove setuptools runtime dependency
### Problem trying to solve

I'm trying to remove `setuptools` from my application dependencies (so I can use [`pip-tools`](https://pypi.org/project/pip-tools/1.8.0/) with [`pip install --no-deps`](https://pip.pypa.io/en/stable/cli/pip_install/#install-no-deps), as per [pip's recommendation](https://pip.pypa.io/en/stable/topics/repeatable-installs/))

### Possible solution(s)

Remove [`setuptools` requirement](https://github.com/pycontribs/jira/blob/main/setup.cfg#L62) from `install_requires`

### Alternatives

For my use, use `pip-tools --allow-unsafe` or `pip install` without `--no-deps`, but this doesn't solve the underlying issue

### Additional Context

Setuptools is only used in the `jira` package during build-time (as the pyproject.toml (aka PEP517) build backend), and not during runtime (a [repository search](https://github.com/pycontribs/jira/search?q=setuptools) confirms this)

EpicWink:
Closes #1289

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -218,6 +218,3 @@ yanc==0.3.3
     # via jira (setup.cfg)
 zipp==3.6.0
     # via importlib-metadata
-
-# The following packages are considered to be unsafe in a requirements file:
-# setuptools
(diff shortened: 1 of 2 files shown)


adehad:
Thanks!

PR #1292, opened 2022-02-04
Update search_issues return type hint

roansong:
When `json_result` is `True`, the return value is a Dict, not a list of Dicts.


roansong:
I edited this on the Github UI - will take a look at the failing tests offline :)

PR #1336, opened 2022-03-23
gha: use secrets from cloud environment

ssbarnea:


ssbarnea:
I will have a look on monday, I know where to look.

PR #1354, opened 2022-04-14
update default behaviour of `jira.search_issues`

### LINKED ISSUE #1240: search_issue does not return the comments field.
### Bug summary

I am aware that a similar question has been asked before, however the response was to use the following:
`jira.search_issues(jql, fields="comment, the other fields" ... )`

I would like to retrieve ALL the fields + the comments. At runtime, my script does not know all the fields, so I cannot use the aforementioned workaround.

A workaround that I have considered is to send a small query with `maxResults:1` and without the `fields="comment"` parameter to the `search_issues(...)` function. This will allow me to obtain all of the keys of the values within the `fields` parameter. I could then join them into a single string, and append ",comment" to it, and make my future requests with this new string as the fields parameter as such:
```
all_fields = ",".join(field_names) + ",comment"
jira.search_issues(jql, fields=all_fields)
```

However, this feels like a workaround. Would it be possible to add some parameter `includeComments=True` within the `search_issues` function to do this automatically?

### Jira Instance type

### Jira instance version

8.14.0

### jira-python version

main

3.8

- [ ] Linux
- [ ] macOS
- [X] Windows

### Reproduction steps

```python
# 1. Call search_issues(...) without specifying fields paramter
# 2. Observe that the result does not include the comments field
# 3. That's the issue
```

### Stack trace

```python
none really
```

### Expected behaviour

Comments to be included, since the documentation says that omitting the fields parameter will return all the available fields.

### Additional Context

_No response_

Yusuf-TJ:
The search_issues documentation states that when nothing is passed to the field parameter, all fields are returned. 
https://jira.readthedocs.io/api.html?highlight=search_issues#jira.client.JIRA.search_issues
The "*all" is used as the field parameter when nothing is passed to achieve this.

Fixes #1240

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2896,7 +2896,7 @@ class JIRA:
         if isinstance(fields, str):
             fields = fields.split(",")
         else:
-            fields = list(fields or [])
+            fields = list(fields or ["*all"])
 
         # this will translate JQL field names to REST API Name
         # most people do know the JQL names so this will help them use the API easier


jira/client.py
```diff
@@ -2896,7 +2896,7 @@ def search_issues(
         if isinstance(fields, str):
             fields = fields.split(",")
         else:
-            fields = list(fields or [])
+            fields = list(fields or ["*all"])
```
adehad, line 2899:
```suggestion
            fields = list(["*all"] if fields is None else fields)
```

Let's do an explicit check with `None`, that way you can pass an empty list if you actually don't want everything.

I don''t really like how we call `list()` on a `str`

```python
>>> list("hey")
['h', 'e', 'y']
```

Do you think you can tidy this up further? i.e. `if isinstance(fields, str)`

adehad (changes requested):
Additionally let's add tests in `test_issue.py` to cover the possibility of fields: a str, a list of str, and None.

jira/client.py
Yusuf-TJ: I was wondering if it would be better to change the default value to '*all" as the None value actually represents all.
This way we only need to convert string inputs to list and pass through list inputs.

adehad: Yep that sounds good too

Yusuf-TJ: Alright then, I'll send a new PR

adehad: (feel free to just update this branch with more commits!)

PR #1359, opened 2022-04-15
use importlib.metadata instead of pkg_resources

### LINKED ISSUE #1248: Usage of `pkg_resources` slows down imports
### Bug summary

The project uses `pkg_resources` for setting version information:

https://github.com/pycontribs/jira/blob/main/jira/__init__.py#L2-L7

If possible, this should be removed in favor of importlib or static `version.py` generation.

Refs:

* https://github.com/pypa/setuptools/issues/510
* https://github.com/protocolbuffers/protobuf/issues/7393

### Jira Instance type

### Jira instance version

any

### jira-python version

main

any

- [X] Linux
- [X] macOS
- [X] Windows

### Reproduction steps

```python
# Run: `python -c "import jira"`
# Observe `pkg_resources` taking a long time to import for larger projects.
```

### Stack trace

```python
N/A
```

### Expected behaviour

I expect `pkg_resources` to not be used so that the startup time is reasonable.

### Additional Context

_No response_

adehad:
# Summary of Changes
1. Remove `pkg_resources` to set `__version__` instead using `importlib.metadata`
2. Remove `pkg_resources` wrapper for parsing version, and directly use `packaging` (note: comes with `setuptools` install, but added to requirements for completeness)

Closes #1248

### DIFF
--- a/jira/__init__.py
+++ b/jira/__init__.py
@@ -1,8 +1,8 @@
 """The root of JIRA package namespace."""
 try:
-    import pkg_resources
+    import importlib.metadata
 
-    __version__ = pkg_resources.get_distribution("jira").version
+    __version__ = importlib.metadata.version("jira")
 except Exception:
     __version__ = "unknown"
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -41,7 +41,7 @@ from typing import (
 from urllib.parse import parse_qs, quote, urlparse
 
 import requests
-from pkg_resources import parse_version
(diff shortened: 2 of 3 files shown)


adehad:
@vkomarov-r7 @3V3RYONE can you give this a review and see if this meets your expectations?

3V3RYONE (commented):
Looks good to me @adehad !! The changes make sense, nice work!

PR #1360, opened 2022-04-15
updated `jira.search_issues` default behaviour to include all fields

### LINKED ISSUE #1240: search_issue does not return the comments field.

Yusuf-TJ:
The `jira.search_issues` documentation states that when nothing is passed to the field parameter, all fields are returned.
https://jira.readthedocs.io/api.html?highlight=search_issues#jira.client.JIRA.search_issues
The "*all" has been set as the default field parameter to achieve this.

Same fix as #1354 
Resolves #1240

### DIFF
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -25,6 +25,27 @@ class IssueTests(JiraTestCase):
         issue = self.jira.issue(self.issue_1)
         self.assertEqual(issue.key, self.issue_1)
         self.assertEqual(issue.fields.summary, "issue 1 from %s" % self.project_b)
+    
+    def test_issue_search_field(self):
+        issues = self.jira.search_issues("key=%s" % self.issue_1)
+        self.assertEqual(self.issue_1, issues[0].key)
+        self.assertIsInstance(issues, list)
+        
+        issues = self.jira.search_issues("key=%s" % self.issue_1, json_result=True)
+        self.assertIsInstance(issues, dict)
+
+        issues = self.jira.search_issues("key=%s" % self.issue_1, fields="comment,assignee")
+        self.assertTrue(hasattr(issues[0].fields, "comment"))
+        self.assertTrue(hasattr(issues[0].fields, "assignee"))
+        self.assertFalse(hasattr(issues[0].fields, "reporter"))
+
+
+        issues = self.jira.search_issues("key=%s" % self.issue_1)
+        self.assertTrue(hasattr(issues[0].fields, "reporter"))
+        self.assertTrue(hasattr(issues[0].fields, "comment"))
+        
+        issues = self.jira.search_issues("key=%s" % self.issue_1, fields=[])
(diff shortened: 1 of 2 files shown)


adehad (changes requested):
Changes look good, let's get some tests in!

adehad:
@Yusuf-TJ Let's re-use this same pull request, easier to keep track of review comments and whether they have been addressed or not.

tests/resources/test_issue.py
```diff
@@ -26,6 +26,28 @@ def test_issue(self):
...
 
+    def test_issue_search_field(self):
+        issues = self.jira.search_issues("key=%s" % self.issue_1)
+        self.assertEqual(self.issue_1, issues[0].key)
+        self.assertIsInstance(issues, list)
+
+        issues = self.jira.search_issues("key=%s" % self.issue_1, json_result=True)
+        self.assertIsInstance(issues, dict)
+
+        issues = self.jira.search_issues(
```
adehad, line 37: Could you also include a brief comment to explain what each block is testing

e.g.
```suggestion

        # Check that a field we do not request does not appear
        issues = self.jira.search_issues(
```

I think it may also be possible to do these as individual test functions
e.g. `issues[0]` is the same as `self.issue_1` (as we proved in the first assertion in this test)

That way we can specify in the function name itself what we are testing

`test_issue_search_finds_issue`
`test_issue_search_can_return_json`
`test_issue_search_only_includes_provided_fields`
`test_issue_search_default_behaviour_included_fields`

adehad (changes requested):
Looking good, a few minor changes, but really like the test cases!

adehad (approved):
Awesome

PR #1364, opened 2022-04-19
Handle #925 rate limit error 

### LINKED ISSUE #925: Handle 429 rate limiting messages from Jira
**Is your feature request related to a problem? Please describe.**
 Atlassian recommends usage of rate limiting with Jira in this blog post: https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html

This would result in getting a 429 status code.  [From what I can tell](https://github.com/pycontribs/jira/blob/91461fd736cd9a37cc136e07402c6f0b1e60170b/jira/resilientsession.py#L100) this library will never retry these requests.

**Describe the solution you'd like**
Use the existing exponential backoff logic to (eventually) retry a request that was ratelimited.

**Describe alternatives you've considered**
I have not really considered any alternative here. 

**Additional context**

Admittedly this may be handled elsewhere and I may have missed it,  I have not rolled out the ratelimiting yet so I do not have real world testing here.

StefanoBoriero:
* add status code 429 to the list of retriable errors

FIXES #925

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -88,7 +88,7 @@ def raise_on_error(r: Optional[Response], verb="???", **kwargs):
 class ResilientSession(Session):
     """This class is supposed to retry requests that do return temporary errors.
 
-    At this moment it supports: 502, 503, 504
+    At this moment it supports: 429
     """
 
     def __init__(self, timeout=None):
@@ -113,11 +113,23 @@ class ResilientSession(Session):
                 f"Got ConnectionError [{response}] errno:{response.errno} on {request} {url}\n{vars(response)}\n{response.__dict__}"
             )
         if isinstance(response, Response):
-            if response.status_code in [502, 503, 504, 401]:
-                # 401 UNAUTHORIZED still randomly returned by Atlassian Cloud as of 2017-01-16
+            if response.status_code in [429]:
+                number_of_tokens_issued_per_interval = response.headers[
+                    "X-RateLimit-FillRate"
+                ]
+                token_issuing_rate_interval_seconds = response.headers[
+                    "X-RateLimit-Interval-Seconds"
+                ]
+                maximum_number_of_tokens = response.headers["X-RateLimit-Limit"]
+                retry_after = response.headers["retry-after"]
                 msg = f"{response.status_code} {response.reason}"
-                # 2019-07-25: Disabled recovery for codes above^
-                return False
+                logging.warning(
+                    f"Request rate limited by Jira: request should be retried after {retry_after} seconds.\n"
+                    + f"{number_of_tokens_issued_per_interval} tokens are issued every {token_issuing_rate_interval_seconds} seconds. "
+                    + f"You can accumulate up to {maximum_number_of_tokens} tokens.\n"
+                    + "Consider adding an exemption for the user as explained in: "
+                    + "https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html"
+                )
             elif not (
                 response.status_code == 200
                 and len(response.content) == 0
--- a/tests/test_resilientsession.py
+++ b/tests/test_resilientsession.py
@@ -1,6 +1,11 @@
 import logging
+from unittest.mock import Mock, patch
+
+import pytest
+from requests import Response
 
 import jira.resilientsession
+from jira.exceptions import JIRAError
 from tests.conftest import JiraTestCase
 
 
@@ -53,3 +58,43 @@ class ResilientSessionLoggingConfidentialityTests(JiraTestCase):
     def tearDown(self):
         jira.resilientsession.logging.getLogger().removeHandler(self.loggingHandler)
         del self.loggingHandler
+
+
+status_codes_retries_test_data = [
+    (429, 4, 3),
+    (401, 1, 0),
+    (403, 1, 0),
+    (404, 1, 0),
+    (502, 1, 0),
+    (503, 1, 0),
+    (504, 1, 0),
+]
+
+
+@patch("requests.Session.get")
+@patch("time.sleep")
+@pytest.mark.parametrize(
+    "status_code,expected_number_of_retries,expected_number_of_sleep_invocations",
+    status_codes_retries_test_data,
+)
+def test_status_codes_retries(
+    mocked_sleep_method: Mock,
+    mocked_get_method: Mock,
+    status_code: int,
+    expected_number_of_retries: int,
+    expected_number_of_sleep_invocations: int,
+):
+    mocked_response: Response = Response()
+    mocked_response.status_code = status_code
+    mocked_response.headers["X-RateLimit-FillRate"] = "1"
+    mocked_response.headers["X-RateLimit-Interval-Seconds"] = "1"
+    mocked_response.headers["retry-after"] = "1"
+    mocked_response.headers["X-RateLimit-Limit"] = "1"
+    mocked_get_method.return_value = mocked_response
+    session: jira.resilientsession.ResilientSession = (
+        jira.resilientsession.ResilientSession()
+    )
+    with pytest.raises(JIRAError):
+        session.get("mocked_url")
+    assert mocked_get_method.call_count == expected_number_of_retries
+    assert mocked_sleep_method.call_count == expected_number_of_sleep_invocations


tests/test_resilientsession.py
```diff
@@ -53,3 +58,16 @@ def test_logging_with_connection_error(self):
     def tearDown(self):
         jira.resilientsession.logging.getLogger().removeHandler(self.loggingHandler)
         del self.loggingHandler
+
+
+@patch("requests.Session.get")
+def test_throttling_error_is_retried(mocked_method: Mock):
```
adehad, line 64: we should patch `time.sleep()` to ensure that it is called too, so we actually retry with the exponential back off!

Otherwise test looks good, really appreciate it

jira/resilientsession.py
```diff
@@ -113,11 +113,10 @@ def __recoverable(
...
             )
         if isinstance(response, Response):
-            if response.status_code in [502, 503, 504, 401]:
-                # 401 UNAUTHORIZED still randomly returned by Atlassian Cloud as of 2017-01-16
+            if response.status_code in [429, 502, 503, 504]:
+                # 401 UNAUTHORIZED still randomly returned by Atlassian Cloud as of 2017-01-16 and are handled by jira.JiraCookieAuth.handle_401
                 msg = f"{response.status_code} {response.reason}"
-                # 2019-07-25: Disabled recovery for codes above^
-                return False
+                return True
```
adehad, line 119: this seems to do a lot more than handle the 429 error code, as we are now returning True for a bunch of other error codes.

Additionally, this is in a short circuit clause that returns True... It doesn't actually reach the `time.sleep(delay)` code !

Can I make a few suggestions?
1. We use a separate elif clause for 429, that does not have a short circuit return value. i.e. it just does some logging and has a comment explaining what it is trying to do. For example: "handle the rate limit code".
2. Within that elif clause we have a `logging.warning` message that we have been rate limited, and we include the information in the headers as found in the #925 reference link, i.e. `X-RateLimit-Remaining` 
3. The warning message also includes a suggestion to add the user to the exemption list if they are hitting this too often.
4. We add a debug message to this exist clause that actually logs the msg, currently the `msg` variable is set but then not used...


### For any future readers

Really apologise for the state of this file... (using the root logger for logging too...). The implementation really isn't great

I'm going to try and raise a PR that moves the code towards the suggestion here:
https://stackoverflow.com/questions/23267409/how-to-implement-retry-mechanism-into-python-requests-library/23271433#comment58969710_23271433

adehad (changes requested):
Really appreciate you making this, I've added a few comments

(also the GitHub way of referencing an issue would be to use the # symbol and then the number, i.e. `#925`)

jira/resilientsession.py
StefanoBoriero: Thanks for the constructive feedback, you raised some good points! Let me try to address them one by one

1. Rather than adding a new elif clause, can I suggest to just get rid of the other 5xx status codes and keep retries only for the 429 code? I find if confusing for readers without context to see them handled in a separate clause, that is functionally equivalent to handling them in the non-retriable codes execution path.
2-3-4. Sure will do that.

tests/test_resilientsession.py
StefanoBoriero: Good catch, it's actually not invoked right now as you pointed out in the other comment

jira/resilientsession.py
adehad: Yep good shout, to be honest I got confused reading that why your test passed for the 502 error... missed the `not` in the `elif`

tests/test_resilientsession.py
```diff
@@ -53,3 +58,36 @@ def test_logging_with_connection_error(self):
...
+    )
+    with pytest.raises(JIRAError):
+        session.get("mocked_url")
+    assert mocked_get_method.call_count == 4
+    assert mocked_sleep_method.call_count == 3
+
+
+@patch("requests.Session.get")
+@patch("time.sleep")
+def test_5xx_error_is_not_retried(mocked_sleep_method: Mock, mocked_get_method: Mock):
```
adehad, line 83: let's `pytest.parametrize()` with the rest of the previous `status_code`s that were tested, just so we can make sure any changes to this behaviour are intentional rather than regressions

StefanoBoriero: Done, I added all the status_codes previously present in this class and couple of more that usually shouldn't be retried like 403 and 404

jira/resilientsession.py
```diff
@@ -137,6 +140,7 @@ def __recoverable(
         if isinstance(response, Response):
             logging.debug("response.headers: %s", response.headers)
             logging.debug("response.body: %s", response.content)
+            logging.debug(msg)
```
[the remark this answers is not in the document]
StefanoBoriero: No worries I missed that too, deleted now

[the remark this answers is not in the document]
StefanoBoriero: Same is true for the header "X-RateLimit-Remaining" which is logged with the rest of the headers on line 141, but with the difference that is logged at debug level. If you think that's enough, I'd remove the log of this specific header in the handling of the 429 response code in favour of the more generic header dump

tests/test_resilientsession.py
adehad: Perfect

jira/resilientsession.py
```diff
@@ -113,11 +113,14 @@ def __recoverable(
...
-                # 401 UNAUTHORIZED still randomly returned by Atlassian Cloud as of 2017-01-16
+            if response.status_code in [429]:
+                rate_limit_remaining = response.headers["X-RateLimit-Remaining"]
                 msg = f"{response.status_code} {response.reason}"
-                # 2019-07-25: Disabled recovery for codes above^
-                return False
+                logging.warning(
+                    f"""Request rate limited by Jira: number of tokens remaining {rate_limit_remaining}. Consider adding exemption for the user as explained in
+                    https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html
+                    """
```
adehad, line 122:
```suggestion
                    f"Request rate limited by Jira: number of tokens remaining {rate_limit_remaining}.\n" 
                    + "Consider adding an exemption for the user as explained in: "
                    + "https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html"
                    """
```

Technically we could have implicit string concat without the `+`, but I think this may make `black` behave better.

#### Why?


`"""` in this way will actually print all the indents!

https://www.online-python.com/
```python
if True:
    if True:
        if True:
            print(f"""Request rate limited by Jira: number of tokens remaining {5}. Consider adding exemption for the user as explained in
                  https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html
                  """)
```

Gives:
```
Request rate limited by Jira: number of tokens remaining 5. Consider adding exemption for the user as explained in
                  https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html
                  
```

adehad (changes requested):
Awesome stuff, just a minor change and otherwise looks good.

---

Regarding:

https://github.com/pycontribs/jira/pull/1364#discussion_r859147195
> Same is true for the header "X-RateLimit-Remaining" which is logged with the rest of the headers on line 141, but with the difference that is logged at debug level. If you think that's enough, I'd remove the log of this specific header in the handling of the 429 response code in favour of the more generic header dump

I prefer the current implementation with the specific error message at the warning level. My reasoning would be:
 
1. It is more like for the log level to be set to info or warning levels than to debug levels.
2. It is a way for the code to document itself, and additionally be more meaningful for the end user than just a dump of headers they need to know the implementation details to understand

But happy to hear your thoughts

jira/resilientsession.py
```diff
@@ -113,11 +113,14 @@ def __recoverable(
...
-            if response.status_code in [502, 503, 504, 401]:
-                # 401 UNAUTHORIZED still randomly returned by Atlassian Cloud as of 2017-01-16
+            if response.status_code in [429]:
+                rate_limit_remaining = response.headers["X-RateLimit-Remaining"]
                 msg = f"{response.status_code} {response.reason}"
-                # 2019-07-25: Disabled recovery for codes above^
-                return False
+                logging.warning(
+                    f"""Request rate limited by Jira: number of tokens remaining {rate_limit_remaining}. Consider adding exemption for the user as explained in
+                    https://confluence.atlassian.com/adminjiraserver/improving-instance-stability-with-rate-limiting-983794911.html
```
adehad, line 121: Actually was just reading the page, and rate_limit_remaining might not be useful here
if we are rate limit that means we have hit the limit? So we might just be logging 0...

Maybe instead to we should log the `X-RateLimit-Limit` and `retry-after` values which represent how many tokens we have and how long we need to wait before the next retry

StefanoBoriero: I think the most interesting headers in this scenario would be `X-RateLimit-FillRate` and `X-RateLimit-Interval-Seconds`, so that a user would get the sense of how often tokens are issued and how many tokens are issued each time, to better understand how to perform and possibly configure retries accordingly. `X-RateLimit-Limit` is the maximum value of tokens that you can accumulate over time, it's still good info but I believe the former ones are more actionable. Anyhow I'll add all of the above.

And yes, is my understanding as well that in a throttled request scenario the value of `X-RateLimit-Remaining` would be 0 so I'll remove it.

jira/resilientsession.py
StefanoBoriero: Cool I didn't know about that, thanks for the detailed info! Python is not my main language, I can put together some working code but I'm lacking knowledge on these details :sweat_smile:

adehad: No problem at all! I think you have an excellent grasp already

adehad (approved):
Awesome, thanks for pushing through this.

adhamselman:
Hello,
https://github.com/pycontribs/jira/pull/1364/files#diff-0605333df4b31ca7f97b2591e3bcba9c71ea201f4a88e383316e7f6c1c478e06R127
Just wondering, what is the point of only logging the value of 'retry-after' instead of taking it into consideration when implementing the backoff?

adehad:
@adhamselman 
Completely understand where you are coming from, and perhaps there is a further improvement by not bothering to retry if this duration is already greater than the `max_retry_delay`

adhamselman:
Not just that. The backoff delay may be calculated to (for example) 20 seconds but the 'retry-after' value may be 30.
I'm not entirely sure but if the retry value is 30 and we retry after 20 seconds, i think the server may penalize you for being too aggressive and increase the wait time.

PR #1366, opened 2022-04-23
tidy ResilientSession implementation

adehad:
## Summary of Changes
- Resilient Session code cleanup, removing dead code and other code repetition
- Updated workaround for retrying the `MultipartEncoder` scheme for attachments

## TODO
- [x] Probably broke some tests that need to be fixed
- [x] Make sure docstrings are ok
- [x] add some tests for the new retry object

## Special Thanks:
@datashaman : https://stackoverflow.com/questions/23267409/how-to-implement-retry-mechanism-into-python-requests-library/23271433#comment58969710_23271433

```
Co-Authored-By: datashaman <59514+datashaman@users.noreply.github.com>
```


jira/client.py
```diff
@@ -991,18 +986,17 @@ def file_stream() -> MultipartEncoder:
                     fields={"file": (fname, attachment, "application/octet-stream")}
```
adehad, line 986: `attachment` here is IO stream, I wonder if this needs to `.seek(0)` so that on retries it actually works...

adehad: yep

adehad:
@StefanoBoriero , could I bother you to have a look at this?

As you were recently in this code I thought you might be able to critique this !

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
     """
     request = kwargs.get("request", None)
-    # headers = kwargs.get('headers', None)
 
-    if r is None:
+    if resp is None:
         raise JIRAError(None, **kwargs)
 
-    if r.status_code >= 400:
+    if not resp.ok:  # equivalent to .status_code < 400 & >200
```
adehad, line 46:
```suggestion
    if not resp.ok:  # equivalent to .status_code < 400 & >=200
```

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
-        if r.status_code == 403 and "x-authentication-denied-reason" in r.headers:
-            error = r.headers["x-authentication-denied-reason"]
-        elif r.text:
+        if resp.status_code == 403 and "x-authentication-denied-reason" in resp.headers:
+            error = resp.headers["x-authentication-denied-reason"]
+        elif resp.text:
             try:
-                response = json.loads(r.text)
-                if "message" in response:
+                resp_data = json.loads(resp.text)
```
adehad, line 52:
```suggestion
                resp_data = resp.json()
```

StefanoBoriero:
> @StefanoBoriero , could I bother you to have a look at this?
> 
> As you were recently in this code I thought you might be able to critique this 

With pleasure! I'll find some time in the next few days to do so

jira/resilientsession.py
```diff
@@ -100,20 +104,137 @@ def __init__(self, timeout=None):
...
+            except ConnectionError as e:
+                exception = e
+
+            # Decide if we should keep retrying
+            response_or_exception = response if response is not None else exception
+            retry_number += 1
+            if is_retrying() and self.__recoverable(
+                response_or_exception, url, method.upper(), retry_number
+            ):
+                _prepare_retry_method.prepare(processed_kwargs)
```
StefanoBoriero, line 169: I'm not too sure what this function is supposed to do: it returns the input unchanged but the returned value is ignored, am I missing something?

jira/resilientsession.py
```diff
@@ -100,20 +104,137 @@ def __init__(self, timeout=None):
...
+            else:
+                retry_number = self.max_retries + 1  # exit the while loop, as above max
+
+        if exception is not None:
+            # We got an exception we could not recover from
+            raise exception
+        elif raise_on_error(response, **processed_kwargs):
```
StefanoBoriero, line 176: I'm not too sure about the rationale of having the function `raise_on_error` returning a boolean and using it in an if condition: couldn't it follow the same behaviour of the `raise_for_status` function from requests.Response? Raise an exception if some conditions are met, otherwise no operation. Then you could do
```
raise_on_error(response)
return response
```
outside of the if statement, and you wouldn't need the Runtime exception (not 100% sure about this last statement)

StefanoBoriero (commented):
Good improvements! I left some comments mostly about refactoring, let me know what you think

jira/resilientsession.py
```diff
@@ -100,20 +104,137 @@ def __init__(self, timeout=None):
...
     def __recoverable(
         self,
         response: Optional[Union[ConnectionError, Response]],
-        url: str,
-        request,
+        url: Union[str, bytes],
+        request_method: str,
         counter: int = 1,
     ):
+        """Return whether the request is recoverable and hence should be retried.
```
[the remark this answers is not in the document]
adehad: what arguments would this second function take? Really this is just a bunch of logging and then two lines for the delay

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
 
 
 class ResilientSession(Session):
     """This class is supposed to retry requests that do return temporary errors.
 
-    At this moment it supports: 429
+    :py:meth:`__recoverable` handles all retry-able errors.
     """
 
     def __init__(self, timeout=None):
```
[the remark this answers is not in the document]
adehad: fine with me, `timeout` isn't used to my knowledge for some reason... But that is out of scope for this change for sure haha

jira/resilientsession.py
adehad: yeah, the default does that, but in the case for adding an attachment we pass in a custom one which actually regenerates the data and headers for the request. 

Not an ideal situation, but happy to take alternatives, I went through a few iterations, but as this is where the retry behaviour and delays are this implementation has the least code!

jira/resilientsession.py
adehad: the reasoning is based around the static type checking, the [`TypeGuard` type](https://peps.python.org/pep-0647), it is basically instructing the type checker that we have checked that the `Response` is actually a `Response` and not `None`.

So we could definitely do what you described, but mypy will complain that `Optional[Response]` not a `Response` is being output from the function.
We could do a `cast()` to just tell mypy we know that it will be a `Response`, but I think this can let bugs slip through. 

Finally, `raise_on_error()` is automatically called for each request (unlike `raise_for_status()`) so it is more of an implementation detail than a user function, so I'm not too worried about how a user will use it.

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
+    def prepare(self, original_request_kwargs) -> dict:
+        return original_request_kwargs
+
+
+class PassthroughRetryPrepare(PrepareRequestForRetry):
+    def prepare(self, original_request_kwargs) -> dict:
+        return super().prepare(original_request_kwargs)
+
+
+def raise_on_error(resp: Optional[Response], **kwargs) -> TypeGuard[Response]:
```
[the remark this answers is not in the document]
adehad: Refer to: https://github.com/pycontribs/jira/pull/1366#discussion_r864326830 where I talk more about the TypeGuard and link its PEP

adehad:
Thanks @StefanoBoriero, I've tried to address most of the concerns, with some discussions points.

I've organised my commits to tackle individual comments (as GitHub a bit weird this way) and left it up to you to resolve threads that I've addressed.

jira/client.py
```diff
@@ -2672,7 +2662,7 @@ def create_temp_project_avatar(
         if size != size_from_file:
             size = size_from_file
 
-        params = {"filename": filename, "size": size}
+        params: Dict[str, Union[int, str]] = {"filename": filename, "size": size}
```
studioj, line 2665: its probably personal preference, but for me this doesnt improve readability. What would be the advantage of adding type annotation here?

jira/resilientsession.py
```diff
@@ -1,119 +1,262 @@
...
+    Args:
+        resp (Response): The Jira API request's response.
+
+    Returns:
+        str: The error message parsed from the Response. An empty string if no error.
+    """
+    resp_data: Dict[str, Any] = {}  # json parsed from the response
+    parsed_error = ""  # error message parsed from the response
+
+    if resp.status_code == 403 and "x-authentication-denied-reason" in resp.headers:
```
studioj, line 75: probably so but validated on both server and cloud?

jira/resilientsession.py
```diff
@@ -1,119 +1,262 @@
...
+    if "message" in resp_data:
+        # Jira 5.1 errors
+        parsed_error = resp_data["message"]
+    elif "errorMessages" in resp_data:
+        # Jira 5.0.x error messages sometimes come wrapped in this array
+        # Sometimes this is present but empty
+        error_messages = resp_data["errorMessages"]
+        if len(error_messages) > 0:
+            if isinstance(error_messages, (list, tuple)):
+                parsed_error = error_messages[0]
```
studioj, line 92: why only the first element? could they be appended?

jira/resilientsession.py
```diff
@@ -1,119 +1,262 @@
...
+        """
+        self.timeout = timeout  # TODO: Unused?
+        self.max_retries = max_retries
+        self.max_retry_delay = max_retry_delay
         super().__init__()
 
         # Indicate our preference for JSON to avoid https://bitbucket.org/bspeakmon/jira-python/issue/46 and https://jira.atlassian.com/browse/JRA-38551
         self.headers.update({"Accept": "application/json,*.*;q=0.9"})
 
+    def _jira_prepare(self, **_kwargs) -> dict:
```
studioj, line 128: why _kwargs and not kwargs here? I might not be fully up to date on the standards here

studioj (commented):
quite a lengthy PR ... I'd like to test this out before giving my "official" approval :) I'll try to do so this week

jira/client.py
```diff
@@ -1016,8 +1006,9 @@ def file_stream() -> MultipartEncoder:
         )
         if jira_attachment.size == 0:
             raise JIRAError(
-                "Added empty attachment via %s method?!: r: %s\nattachment: %s"
-                % (method, r, jira_attachment)
+                "Added empty attachment?!: r: {}\nattachment: {}".format(
```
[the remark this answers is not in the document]
adehad: Looks like pyupgrade doesn't always convert to an f-string:

https://github.com/asottile/pyupgrade#f-strings

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

jira/client.py
adehad: So now that we no longer use our own typehints for the requests methods (before resilient session had an override for the typehints), mypy wasn't happy as it inferred the type as `Dict[str, object]`

jira/resilientsession.py
adehad: I think either way it is out of scope for this PR. Hasn't been reported amongst the issues so it probably works as intended.
Amazingly, I couldn't find the endpoint in the Cloud docs page...

jira/resilientsession.py
adehad: it's a good point, this was the original implementation, but I'll join them with a newline character, as no harm done

jira/resilientsession.py
adehad: nah it just a bad variable name, I'll updated the names

(recall, `**variable_name` just takes any additional keyword arguments passed into the function that are not explicitly listed in the function definition)

jira/resilientsession.py
adehad: Maybe it was for older jira versions? 
As the jira docs don't report it anywhere
https://developer.atlassian.com/search/?q=x-authentication-denied-reason&p=1


There is reference to:
https://developer.atlassian.com/search/?q=x-seraph-loginreason&p=1

jira/resilientsession.py
```diff
@@ -123,103 +266,31 @@ def __recoverable(
...
-
-    def patch(self, url: Union[str, bytes], data=None, **kwargs) -> Response:  # type: ignore
-        return self.__verb("PATCH", str(url), data=data, **kwargs)
+        if is_recoverable:
+            # Exponential backoff with full jitter.
+            delay = min(self.max_retry_delay, 10 * 2**counter) * random.random()
+            LOG.warning(
+                "Got recoverable error from %s %s, will retry [%s/%s] in %ss. Err: %s"
+                % (request_method, url, counter, self.max_retries, delay, msg)  # type: ignore[str-bytes-safe]
+            )
```
adehad, line 283: I wonder if this should be in the `__init__` as maybe if it gets here it is too late to warn the developer about it?

jira/resilientsession.py
```diff
@@ -123,103 +266,31 @@ def __recoverable(
...
+            if isinstance(response, Response):
+                LOG.debug(
+                    "Dumping Response headers and body. "
+                    + f"Log level debug in '{__name__}' is not safe for production code!"
+                )
```
adehad, line 288: I wonder if this should be in the `__init__` as maybe if it gets here it is too late to warn the developer about it?

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
     """
     request = kwargs.get("request", None)
-    # headers = kwargs.get('headers', None)
 
-    if r is None:
+    if resp is None:
         raise JIRAError(None, **kwargs)
 
-    if r.status_code >= 400:
+    if not resp.ok:  # equivalent to .status_code < 400 & >=200
```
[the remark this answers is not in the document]
StefanoBoriero: Great, I'd like to resolve this but I can't find the proper way of doing so, could it be because I didn't have write permission to the repo when the PR was opened? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#resolving-conversations
Anyhow consider this resolved

jira/resilientsession.py
```diff
@@ -1,94 +1,98 @@
...
+                    and len(resp_data["errors"]) > 0
+                    and isinstance(resp_data["errors"], dict)
                 ):
                     # Jira 6.x error messages are found in this array.
-                    error_list = response["errors"].values()
+                    error_list = resp_data["errors"].values()
                     error = ", ".join(error_list)
                 else:
-                    error = r.text
+                    error = resp.text
```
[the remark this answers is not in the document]
StefanoBoriero: LGTM, please feel free to resolve this.

jira/resilientsession.py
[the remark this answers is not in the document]
StefanoBoriero: Ok thanks I understand now, consider this resolved.

jira/resilientsession.py
StefanoBoriero: It's fine with me, I was asking out of curiosity because I couldn't figure it out! Consider this resolved.

jira/resilientsession.py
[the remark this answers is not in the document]
StefanoBoriero: It would take `counter`, that is the number of retries performed so far, to calculate the delay. I don't feel strongly about this so feel free to keep the current structure, it's just a personal opinion about separating the two concerns.

jira/resilientsession.py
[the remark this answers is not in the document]
StefanoBoriero: Cool, I agree about the scope, let's keep it in mind if there's going to be another iteration on this. Consider this discussion resolved

jira/resilientsession.py
StefanoBoriero: Ok, thanks for the explanation. It doesn't bother much, leave it as is and consider this resolved.

jira/resilientsession.py
adehad: I'll leave this unresolved until I document the class!

jira/resilientsession.py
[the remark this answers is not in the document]
adehad: Ahh didn't realise this was a limitation, strange that GitHub allows you to create your own, but not resolve your own !

Thanks for commenting on the threads instead!

studioj (dismissed):
rework looks ok and the code is covered in the functional tests ... only the extra doc still todo

jira/utils/__init__.py
```diff
@@ -56,24 +56,24 @@ def threaded_requests(requests):
...
 
     Raises:
         JIRAError: via :py:func:`jira.resilientsession.raise_on_error`
 
     Returns:
         Union[List[Dict[str, Any]], Dict[str, Any]]: the json
     """
-    raise_on_error(r)  # if 'r' is None, will raise an error here
-    r = cast(Response, r)  # tell mypy only Response-like are here
+    raise_on_error(resp)  # if 'r' is None, will raise an error here
```
adehad, line 71:
```suggestion
    raise_on_error(resp)  # if 'resp' is None, will raise an error here
```

adehad (commented):
Thanks @studioj , I'll make some time to finish the PR

jira/client.py
```diff
@@ -1785,8 +1777,7 @@ def assign_issue(self, issue: Union[int, str], assignee: Optional[str]) -> bool:
         url = self._get_latest_url(f"issue/{issue}/assignee")
         user_id = self._get_user_id(assignee)
         payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}
-        r = self._session.put(url, data=json.dumps(payload))
-        raise_on_error(r)
+        self._session.put(url, data=json.dumps(payload))
```
studioj, line 1780: no more raise?
or is that a given?
that could result maybe in different behavior in the parts where this raise wasnt specifically called...

jira/resilientsession.py
```diff
@@ -1,119 +1,288 @@
...
+
+    Returns:
+        TypeGuard[Response]: True if the passed in Response is all good.
     """
     request = kwargs.get("request", None)
-    # headers = kwargs.get('headers', None)
 
-    if r is None:
+    if resp is None:
         raise JIRAError(None, **kwargs)
```
studioj, line 64: since you're touching this part ... maybe make the exception slighlty more elaborate

jira/resilientsession.py
```diff
@@ -1,119 +1,288 @@
...
-    # https://answers.atlassian.com/questions/11457054/answers/11975162
-    if (
-        r.status_code == 200
-        and len(r.content) == 0
-        and "X-Seraph-LoginReason" in r.headers
-        and "AUTHENTICATED_FAILED" in r.headers["X-Seraph-LoginReason"]
-    ):
-        pass
+
+    return True  # if no exception was raised, we have a valid Response
```
studioj, line 78: I don't like returning True ....
if no exception => success

jira/resilientsession.py
```diff
@@ -1,119 +1,288 @@
...
+        """
+
+        retry_number = 0
+        exception: Optional[Exception] = None
+        response: Optional[Response] = None
+        response_or_exception: Optional[Union[ConnectionError, Response]]
+
+        processed_kwargs = self._jira_prepare(**kwargs)
+
+        def is_retrying() -> bool:
```
studioj, line 193:
```suggestion
        def is_allowed_to_retry() -> bool:
```

jira/client.py
adehad: not required even in the previous implementation as we already called it internally.

jira/resilientsession.py
adehad: this is required for the `TypeGuard` output type, which should only return True if it ensures that the input argument is the correct type (in this case a `Response`).

https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards

studioj (approved):

PR #1383, opened 2022-05-24
Raise `JIRAError` when no user found

lishoujun:
_get_user_id will out of index when no user find, I add an if and return to fix it

my user use my bot to report issues,  I use _get_user_id to get their accountid, but some of them are not user of my jira.  
In this case, currrent code will get a empty list.  
By return none , I don't add reporter to the issue


lishoujun:
@adehad  review pls

adehad:
Thanks @lishoujun, I'll try and have a look this weekend.

Could you elaborate a bit more on your use case? ie when this function returns None how does this help you are trying to do

lishoujun:
@adehad  I have edit the first  comment so other can know  my case easily

adehad:
Thanks @lishoujun for updating the description.

I am a bit concerned about returning None, as None has a specific meaning of "unassigned" in other Jira cases.
I think the behaviour to raise an exception is correct, but perhaps not an IndexError ! As we need a way to indicate to the user that the user was not found.

Perhaps it makes sense to raise a `JiraError(f"No matching users found for: '{user}'")`? This can then be caught in your script and those users ignored?

@studioj any thoughts?

studioj:
> Perhaps it makes sense to raise a `JiraError(f"No matching users found for: '{user}'")`? This can then be caught in your script and those users ignored?
> 
> @studioj any thoughts?

sounds better like that yes

lishoujun:
@adehad  I have update the commit as discuss.   pls check again.

studioj (approved):

adehad:
Thanks for bearing with us @lishoujun , now merged!

lishoujun:
😄

PR #1384, opened 2022-05-25
Remove setup_requires for setuptools_scm_git_archive

fdebrabander:
This dependency is not needed.

The documentation at https://github.com/Changaco/setuptools_scm_git_archive
states that there are two files needed for this to work .git_archival.txt and
.gitattributes. Both files are missing in this repository.

See also https://github.com/pycontribs/jira/discussions/1365

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,6 @@ zip_safe = False
 # These are required during `setup.py` run:
 setup_requires =
     setuptools_scm >= 1.15.0
-    setuptools_scm_git_archive >= 1.0
 install_requires =
     defusedxml
     packaging
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,7 +50,7 @@ include_package_data = True
 zip_safe = False
 # These are required during `setup.py` run:
 setup_requires =
-    setuptools_scm >= 1.15.0
+    setuptools_scm >= 7.0.0
 install_requires =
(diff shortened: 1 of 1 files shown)


studioj (approved):

ssbarnea (changes requested):
If you fix this, please ensure that you require setuptools_scm > 7 like it is described in the link you provided. You should also check that pip can install git archives made by github, so you might have to add the missing files.

setup.cfg
```diff
@@ -51,7 +51,6 @@ zip_safe = False
 # These are required during `setup.py` run:
 setup_requires =
     setuptools_scm >= 1.15.0
```
studioj, line 53: as @ssbarnea says maybe bump the required version for setuptools_scm

studioj (changes requested):

setup.cfg
adehad, line 53:
```suggestion
    setuptools_scm >= 7.0.0
```

adehad:
man it's actually a mess, I'll make a PR to move to a single `pyproject.toml`, too much legacy stuff kicking about

adehad:
Refer to #1443 
Thanks for initiating this @fdebrabander, you are a co-author on the merged commit, and thank you for your contribution !

PR #1008, opened 2021-03-17
Added 'notify' argument to the comment update() method

ronniel1:
The underlying update() method used, already supports the 'notify' argument.
Added that argument to Comment's update() method and propagated it on to Resource's update() method

PR #1387, opened 2022-06-01
Add notify to comment

studioj:
my own implementation of https://github.com/pycontribs/jira/pull/1008

I didn't want to add the functionality to `jira.add_comment` since I'm not sure if it has the same way of adding the notification as add_user => `data["notification"]='false'` OR `data["notifyUsers"] = "false"`

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4330,16 +4330,16 @@ class JIRA:
             directoryId (int): The directory ID the new user should be a part of (Default: 1)
             password (Optional[str]): Optional, the password for the new user
             fullname (Optional[str]): Optional, the full name of the new user
-            notify (bool): Whether or not to send a notification to the new user. (Default: False)
-            active (bool): Whether or not to make the new user active upon creation. (Default: True)
-            ignore_existing (bool): Whether or not to ignore and existing user. (Default: False)
-            applicationKeys (Optional[list]): Keys of products user should have access to
+            notify (bool): Whether to send a notification to the new user. (Default: False)
+            active (bool): Whether to make the new user active upon creation. (Default: True)
+            ignore_existing (bool): Whether to ignore and existing user. (Default: False)
+            application_keys (Optional[list]): Keys of products user should have access to
 
         Raises:
             JIRAError:  If username already exists and `ignore_existing` has not been set to `True`.
 
         Returns:
-            bool: Whether or not the user creation was successful.
+            bool: Whether the user creation was successful.
 
 
         """
--- a/jira/client.py
+++ b/jira/client.py
@@ -1852,22 +1852,16 @@ class JIRA:
         data: Dict[str, Any] = {"body": body}
 
         if is_internal:
-            data.update(
-                {
-                    "properties": [
-                        {"key": "sd.public.comment", "value": {"internal": is_internal}}
-                    ]
-                }
-            )
-
+            data["properties"] = [
+                {"key": "sd.public.comment", "value": {"internal": is_internal}}
+            ]
         if visibility is not None:
             data["visibility"] = visibility
 
         url = self._get_url("issue/" + str(issue) + "/comment")
         r = self._session.post(url, data=json.dumps(data))
 
-        comment = Comment(self._options, self._session, raw=json_loads(r))
-        return comment
+        return Comment(self._options, self._session, raw=json_loads(r))
 
     # non-resource
     @translate_resource_args
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -763,14 +763,45 @@ class Comment(Resource):
             self._parse_raw(raw)
         self.raw: Dict[str, Any] = cast(Dict[str, Any], self.raw)
 
-    def update(self, fields=None, async_=None, jira=None, body="", visibility=None):
-        """Update a comment"""
-        data = {}
+    def update(  # type: ignore[override]
+        # The above ignore is added because we've added new parameters and order of parameters is different.
+        # Will need to be solved in a major version bump.
+        self,
+        fields: Optional[Dict[str, Any]] = None,
+        async_: Optional[bool] = None,
+        jira: "JIRA" = None,
+        body: str = "",
+        visibility: Optional[Dict[str, str]] = None,
+        is_internal: bool = False,
+        notify: bool = True,
+    ):
+        """Update a comment
+
+        Keyword arguments are marshalled into a dict before being sent.
+
+        Args:
+            fields (Optional[Dict[str, Any]]): DEPRECATED => a comment doesn't have fields
+            async_ (Optional[bool]): If True the request will be added to the queue, so it can be executed later using async_run()
+            jira (jira.client.JIRA): Instance of Jira Client
+            visibility (Optional[Dict[str, str]]): a dict containing two entries: "type" and "value".
+              "type" is 'role' (or 'group' if the Jira server has configured
+              comment visibility for groups) and 'value' is the name of the role
+              (or group) to which viewing of this comment will be restricted.
+            body (str): New text of the comment
+            is_internal (bool): Defines whether a comment has to be marked as 'Internal' in Jira Service Desk (Default: False)
+            notify (bool): Whether to notify users about the update. (Default: True)
+        """
+        data: Dict[str, Any] = {}
         if body:
             data["body"] = body
         if visibility:
             data["visibility"] = visibility
-        super().update(data)
+        if is_internal:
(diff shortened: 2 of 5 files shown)


studioj:
how do i ignore since i dont want to change the api i dont think it's smart to listen to mypy here
```
jira/resources.py:766: error: Signature of "update" incompatible with supertype "Resource"  [override]
jira/resources.py:766: note:      Superclass:
jira/resources.py:766: note:          def update(self, fields: Optional[Dict[str, Any]] = ..., async_: Optional[bool] = ..., jira: Optional[JIRA] = ..., notify: bool = ..., **kwargs: Any) -> Any
jira/resources.py:766: note:      Subclass:
jira/resources.py:766: note:          def update(self, fields: Optional[Dict[str, Any]] = ..., async_: Optional[bool] = ..., jira: Optional[JIRA] = ..., body: str = ..., visibility: Optional[Dict[str, str]] = ..., is_internal: bool = ..., notify: bool = ...) -> Any
```

pyproject.toml
```diff
@@ -5,5 +5,4 @@ requires = [
   "setuptools_scm_git_archive >= 1.0",
   "wheel",
 ]
-requires-python = ">=3.8"
```
studioj, line 8: this triggerred failing builds as it was an unknown parameter

tests/resources/test_comment.py
```diff
@@ -4,12 +4,17 @@
 class CommentTests(JiraTestCase):
     def setUp(self):
         JiraTestCase.setUp(self)
-        self.issue_1 = self.test_manager.project_b_issue1
-        self.issue_2 = self.test_manager.project_b_issue2
-        self.issue_3 = self.test_manager.project_b_issue3
+        self.issue_1_key = self.test_manager.project_b_issue1
```
studioj, line 7: rename to signify it's the key which is stored in this variable

tests/resources/test_comment.py
```diff
@@ -4,12 +4,17 @@
...
+        self.issue_2_key = self.test_manager.project_b_issue2
+        self.issue_3_key = self.test_manager.project_b_issue3
+
+    def tearDown(self) -> None:
```
studioj, line 11: safer to delete everything afterwards

jira/client.py
```diff
@@ -4330,16 +4324,16 @@ def add_user(
             directoryId (int): The directory ID the new user should be a part of (Default: 1)
             password (Optional[str]): Optional, the password for the new user
             fullname (Optional[str]): Optional, the full name of the new user
-            notify (bool): Whether or not to send a notification to the new user. (Default: False)
-            active (bool): Whether or not to make the new user active upon creation. (Default: True)
-            ignore_existing (bool): Whether or not to ignore and existing user. (Default: False)
-            applicationKeys (Optional[list]): Keys of products user should have access to
+            notify (bool): Whether to send a notification to the new user. (Default: False)
```
crimsonxiii, line 4327: had to look it up (https://mcelroytutoring.com/blog-post.php?id=4165#:~:text=Here's%20the%20trick%3A%20%22Whether%20or,the%20outcome%20itself%20is%20undetermined.) but uneed seems better to remove the 'or not'

tests/resources/test_comment.py
```diff
@@ -61,9 +66,11 @@ def test_add_comment_with_issue_obj(self):
         comment.delete()
 
     def test_update_comment(self):
```
crimsonxiii, line 68: I would split into two tests (one test, one assert)

jira/resources.py
```diff
@@ -763,14 +764,46 @@ def __init__(
...
-        """Update a comment"""
-        data = {}
+    # mypy complains about the signature as the order of the parameters is different.
+    # If we follow mypy here we will break backwards compatibility with version <= 3.2 which is probably not advised
+    @typing.no_type_check
+    def update(
+        self,
+        fields: Optional[Dict[str, Any]] = None,
+        async_: Optional[bool] = None,
+        jira: "JIRA" = None,
```
crimsonxiii, line 774: more for myself, but why the quotes around JIRA?
I didn't see such a variant on https://docs.python.org/3/library/typing.html

crimsonxiii (approved):

jira/client.py
```diff
@@ -4330,16 +4324,16 @@ def add_user(
...
+            active (bool): Whether to make the new user active upon creation. (Default: True)
+            ignore_existing (bool): Whether to ignore and existing user. (Default: False)
```
adehad, line 4329: (non-blocking), I find it clearer to be explicit on what True does for boolean arguments.

e.g. 
```
notify (bool): True will send a notification to the new user. (Default: False)
```

jira/resources.py
```diff
@@ -763,14 +764,46 @@ def __init__(
             self._parse_raw(raw)
         self.raw: Dict[str, Any] = cast(Dict[str, Any], self.raw)
 
-    def update(self, fields=None, async_=None, jira=None, body="", visibility=None):
-        """Update a comment"""
-        data = {}
+    # mypy complains about the signature as the order of the parameters is different.
+    # If we follow mypy here we will break backwards compatibility with version <= 3.2 which is probably not advised
+    @typing.no_type_check
+    def update(
```
adehad, line 770: I believe the correct way for this case is to [`# type: ignore[override]`](https://mypy.readthedocs.io/en/stable/common_issues.html?#incompatible-overrides) - which will still type check the rest of the funciton but indicate to mypy this is a intentional break from the super class

jira/resources.py
adehad: This is because we only import the `JIRA` class when type checking, and current Python versions actually try to evaluate the type hint.

Some more reading here: https://mypy.readthedocs.io/en/stable/runtime_troubles.html#annotation-issues-at-runtime

adehad (changes requested):
Looks good, one main suggestion about the type hinting

pyproject.toml
adehad: my bad, sorry about that

jira/client.py
studioj: It's a personal preference I share... Though I would prefer to open a new PR and apply this consistently amongst all doc strings.

jira/resources.py
studioj: I'll fix this!

pyproject.toml
adehad: Just looked into this, I think this needs to be under the `[project]` header:
https://stackoverflow.com/a/72462723/7724187

adehad (changes requested):

pyproject.toml
studioj: also found that but there's like much inclarity as to what does things actually do ... many people are confused :-D as am I.

I'll follow your suggestion

studioj: didnt work :-S

jira/resources.py
studioj: didnt work :-S

studioj: ok worked out if i move it to the first line of update

jira/resources.py
```diff
@@ -763,14 +763,46 @@ def __init__(
...
         self.raw: Dict[str, Any] = cast(Dict[str, Any], self.raw)
 
-    def update(self, fields=None, async_=None, jira=None, body="", visibility=None):
-        """Update a comment"""
-        data = {}
+    # mypy complains about the signature as the order of the parameters is different.
+    # If we follow mypy here we will break backwards compatibility with version <= 3.2 which is probably not advised
+
+    def update(  # type: ignore[override]
+        self,
```
adehad, line 770: would be better if we could keep this comment closer to its use, and I don't think all of this is correct

```suggestion

    def update(  # type: ignore[override] # We've added new parameters and order of parameters is different
        self,
```

maybe we could even do, if the formatter is happy with it

```suggestion

    def update(  # type: ignore[override]
        # We've added new parameters and order of parameters is different
        self,
```

jira/client.py
studioj: https://github.com/pycontribs/jira/tree/unify_docstrings_for_bool_to_refer_whats_true

studioj:
@adehad would be nice if this is released "soon" (maybe together with the other approved PR) my team would really like to use this feature :-)
I don't know the effort to release, glad to help if needed

adehad (approved):

adehad:
@studioj yep I think a new release is due soon too! I'll also make progress on my PR, would be great if you can review it once I've made the changes

PR #1394, opened 2022-06-03
Configurable batch size for `JIRA._fetch_pages()` and dependant methods

### LINKED ISSUE #1393: "search_issues" - performance improvements
### Problem trying to solve

When using `JIRA.search_issues` to retrieve all relevant Issues (`maxResults=False`) the API-Calls get batches of 50 Issues.
This is quite the performance hurdle, as there is an enormous overhead on multiple levels when trying to get more than 1000 issues (or even less).

I benchmarked `JIRA.search_issues` with 1. batches of size 50 and 2. batches of size 500. The JQL encompassed ~7000 issues and only the summary-field was pulled (10 runs).

Avg-time:
1. 45.5 seconds
2. 6.1 seconds

### Possible solution(s)

Add an optional parameter `batch_size` to `JIRA._fetch_pages`.
1. If it is specified and maxResults evaluates as falsy every API-call should query `batch_size`-issues. 
If the fist request returns less issues than specified by `batch_size` AND it is not the last request: Emit a warning (along the lines of "Tried to retrieve {batch_size } issues in one request, but JIRA-server returns at most {actual_amount_retrieved}. Falling back to batch size of {actual_amount_retrieved}.") and fallback to the actual amount of issues received.

2. If it is specified and `maxResults` is truthy -> Emit warning, that `batch_size` should only be used in conjunction with a falsy `maxResults` and ignore it.

3. If it is not specified, default to current behaviour

All methods using `fetch_pages` would have the new parameter added and pass it over.

### Alternatives

_No response_

### Additional Context

PR will follow, shortly.

---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

rynkk:
Add `batch_size` argument to `JIRA._fetch_pages` and dependant methods.
This allows for  significant performance improvements when trying to retrieve a lot of items using `maxResults=False`.

fixes #1393

The program was tested solely for our own use cases, which might differ from yours.

---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

Licensed under [BSD-2-Clause license](https://github.com/pycontribs/jira/blob/main/LICENSE)</sub>

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -473,9 +473,16 @@ class JIRA:
             proxies (Optional[Any]): Sets the proxies for the HTTP session.
             auth (Optional[Tuple[str,str]]): Set a cookie auth token if this is required.
             logging (bool): Determine whether or not logging should be enabled. (Default: True)
+            default_batch_sizes (Optional[Dict[ResourceType, Optional[int]]]): Manually specify the batch-sizes for the paginated retrieval of different item types.
+              `Resource` is used as a fallback for every item type not specified. If a item type is mapped to `None` no fallback occurs, instead the JIRA-Backend will use its default batch-size.
+              By default Issues will be queried in batches of 500, all other item types (`Resource`-key) will be queried in batches of 100.
+              Above standard configuration could be manually supplied by setting this parameter:
+                ``{Issue: 500, Resource: 100}``
+              (Default: None)
         """
         # force a copy of the tuple to be used in __del__() because
         # sys.version_info could have already been deleted in __del__()
+
         self.sys_version_info = tuple(sys.version_info)
         if options is None:
             options = {}
@@ -496,7 +503,7 @@ class JIRA:
         LOG.setLevel(_logging.INFO if logging else _logging.CRITICAL)
         self.log = LOG
 
-        self._options: Dict[str, Any] = copy.copy(JIRA.DEFAULT_OPTIONS)
+        self._options: Dict[str, Any] = copy.deepcopy(JIRA.DEFAULT_OPTIONS)
 
         if default_batch_sizes:
             self._options["default_batch_size"].update(default_batch_sizes)
@@ -831,12 +838,21 @@ class JIRA:
             raise KeyError(str(e) + " : " + json.dumps(resource))
 
     def _get_batch_size(self, item_type: Type[ResourceType]) -> Optional[int]:
+        """
+        Return the batch size for the given resource type from the options.
+
+        Check if specified item-type has a mapped batch-size, else try to fallback to batch-size assigned to `Resource`, else fallback to Backend-determined batch-size.
+
+        Returns:
+           Optional[int]: The batch size to use. When the configured batch size is None, the batch size should be determined by the JIRA-Backend.
+        """
         batch_sizes: Dict[Type[Resource], Optional[int]] = self._options[
             "default_batch_size"
         ]
         try:
             item_type_batch_size = batch_sizes[item_type]
         except KeyError:
+            # Cannot find Resource-key -> Fallback to letting JIRA-Backend determine batch-size (=None)
             item_type_batch_size = batch_sizes.get(Resource, None)
         return item_type_batch_size
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -397,7 +397,7 @@ class JIRA:
         proxies: Any = None,
         timeout: Optional[Union[Union[float, int], Tuple[float, float]]] = None,
         auth: Tuple[str, str] = None,
-        default_batch_sizes: Optional[Dict[Type[ResourceType], Optional[int]]] = None,
+        default_batch_sizes: Optional[Dict[Type[Resource], Optional[int]]] = None,
     ):
         """Construct a Jira client instance.
 
@@ -473,7 +473,7 @@ class JIRA:
             proxies (Optional[Any]): Sets the proxies for the HTTP session.
             auth (Optional[Tuple[str,str]]): Set a cookie auth token if this is required.
             logging (bool): Determine whether or not logging should be enabled. (Default: True)
-            default_batch_sizes (Optional[Dict[Type[ResourceType], Optional[int]]]): Manually specify the batch-sizes for the paginated retrieval of different item types.
+            default_batch_sizes (Optional[Dict[Type[Resource], Optional[int]]]): Manually specify the batch-sizes for the paginated retrieval of different item types.
               `Resource` is used as a fallback for every item type not specified. If a item type is mapped to `None` no fallback occurs, instead the JIRA-Backend will use its default batch-size.
               By default Issues will be queried in batches of 500, all other item types (`Resource`-key) will be queried in batches of 100.
               Above standard configuration could be manually supplied by setting this parameter to: ``{Issue: 500, Resource: 100}``
--- a/jira/client.py
+++ b/jira/client.py
@@ -474,9 +474,9 @@ class JIRA:
             logging (bool): Determine whether or not logging should be enabled. (Default: True)
             default_batch_sizes (Optional[Dict[Type[Resource], Optional[int]]]): Manually specify the batch-sizes for
               the paginated retrieval of different item types. `Resource` is used as a fallback for every item type not
-              specified. If a item type is mapped to `None` no fallback occurs, instead the JIRA-Backend will use its
+              specified. If an item type is mapped to `None` no fallback occurs, instead the JIRA-backend will use its
               default batch-size. By default all Resources will be queried in batches of 100. E.g., setting this to
-              ``{Issue: 500, Resources: None}`` will make :py:meth:`search_issues` query Issues in batches of 500, while
+              ``{Issue: 500, Resource: None}`` will make :py:meth:`search_issues` query Issues in batches of 500, while
               every other item type's batch-size will be controlled by the backend. (Default: None)
 
         """
--- a/jira/client.py
+++ b/jira/client.py
@@ -3143,6 +3143,11 @@ class JIRA:
         Returns:
             ResultList
         """
+        if not username and not query:
+            raise ValueError(
+                "Either 'username' or 'query' arguments must be specified."
+            )
+
         if username is not None:
             params = {"username": username}
         if query is not None:
@@ -3153,11 +3158,6 @@ class JIRA:
             params["issueKey"] = issueKey
         if expand is not None:
             params["expand"] = expand
-
-        if not username and not query:
-            raise ValueError(
-                "Either 'username' or 'query' arguments must be specified."
-            )
 
         return self._fetch_pages(
             User,
--- a/jira/client.py
+++ b/jira/client.py
@@ -681,6 +681,7 @@ class JIRA:
         request_path: str,
         startAt: int = 0,
         maxResults: int = 50,
+        batch_size: Optional[int] = None,
         params: Dict[str, Any] = None,
         base: str = JIRA_BASE_URL,
     ) -> ResultList[ResourceType]:
@@ -694,6 +695,9 @@ class JIRA:
             startAt (int): index of the first record to be fetched. (Default: 0)
             maxResults (int): Maximum number of items to return.
               If maxResults evaluates as False, it will try to get all items in batches. (Default:50)
+            batch_size (Optional[int]): Size of batches, only used if maxResults evaluates as False.
+              If not specified, the JIRA-instance's default size will be used. Strongly recommended to use a value over
+              100 for larger requests.
             params (Dict[str, Any]): Params to be used in all requests. Should not contain startAt and maxResults,
               as they will be added for each request created from this function.
             base (str): base URL to use for the requests.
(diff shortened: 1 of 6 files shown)


rynkk:
Required change was a lot less than anticipated: 1 LoC (+1 for fixing tests)

jira/client.py
```diff
@@ -1123,14 +1139,17 @@ def custom_field_option(self, id: str) -> CustomFieldOption:
     # Dashboards
 
     def dashboards(
-        self, filter=None, startAt=0, maxResults=20
+        self, filter=None, startAt=0, maxResults=20, batch_size: Optional[int] = None
```
rynkk, line 1142: I added the new argument to the end of parameters of the user-facing methods to keep everything backward compatible. Might be better to adjust it for a breaking release (this method is fine, others are not).

jira/client.py
```diff
@@ -771,6 +784,9 @@ def _fetch_pages(
                     and (total is None or page_start < total)
                     and len(next_items_page) == page_size
                 ):
+                    page_params = (
```
rynkk, line 787: If omitted the mock calls will point to the same page-params object, which causes retroactive changes to already fired mock calls

jira/client.py
```diff
@@ -771,6 +784,9 @@ def _fetch_pages(
...
+                        params.copy() if params else {}
+                    )  # Hack necessary for mock-calls to not change
```
rynkk, line 789: If omitted the mock calls will point to the same page-params object, which causes retroactive changes to already fired mock calls.

studioj:
I'm not sure what this entails ... it seems just like offering a different way of doing the same thing 🤔 
Could you update the examples maybe to explain the usecase a little bit?
I dont really like using maxResults and batchsize to set the maxResults in the json ...

rynkk:
>... it seems just like offering a different way of doing the same thing 🤔 
Maybe there is a misunderstanding, what exactly do you mean?

`maxResults`, if truthy, tries to pull X items, in one API-call. If you set maxResults to 8000, but the backend only returns at most 1000, you are getting exactly 1000 items (even though you might've expected 8000).
If `maxResults` is falsy, it tries to get ALL items in batches. However, as there is no maxResults (for the API-calls that is) specified for this case, the backend uses its default maxResults-size which is ~50.
See #1393 for details.

On a different note:
Maybe it would be more elegant to add the `batch_size` as an attribute to the JIRA-class. Avoids having to change all the `_fetch_pages`-dependant methods (e.g., `search_issues)`

studioj:
> 

Maybe setting a batch size for the whole jira class seems more logical instead of adding it for quite some methods? => 
Would there be a disadvantage for having a large batch size on shorter queries?
I see your improvement from 41s to 6s is pretty significant so its definitly worth continuing this train of thought.
I'm just trying to prevent tons of parameters being added to methods... It's often seen as a code smell if your parameter list is getting too large ... 

If you would be in dire need for a speedup quickly, probably a quick fix would be to increase the default batch size to 100 🤔?
although I like it to be configurable

adehad (changes requested):
> > 
> 
> Maybe setting a batch size for the whole jira class seems more logical instead of adding it for quite some methods? => Would there be a disadvantage for having a large batch size on shorter queries? I see your improvement from 41s to 6s is pretty significant so its definitly worth continuing this train of thought. I'm just trying to prevent tons of parameters being added to methods... It's often seen as a code smell if your parameter list is getting too large ...
> 
> If you would be in dire need for a speedup quickly, probably a quick fix would be to increase the default batch size to 100 🤔? although I like it to be configurable
> 
> @adehad @ssbarnea some architectural guidance is welcome :-)

@rynkk, thanks for this. Especially as this is one of the more convoluted areas in the code.
As suggested by @studioj I believe this batch size needs to be set at a class instance level. i.e. in here something like DEFAULT_BATCH_SIZE
https://github.com/pycontribs/jira/blob/678903805e34d91c3905d73d20fcc7a3f5c8036c/jira/client.py#L343

I completely understand where you are coming from with wanting to set this per function. It provides the most flexibility, especially as you will likely have more issues than dashboards and therefore would want to set it differently for different resources. Also scales differently depending on how fast the server is you are working with.

A solution I would prefer if you strongly would like to set this per endpoint might look like so:

```python
# within the DEFAULT_OPTIONS

# A dict of (Resource Type, Batch Size) 
DEFAULT_BATCH_SIZE: Dict[Resource, int] = {
    Issue : 500,
    Resource: 100,
}


...

# inside the JIRA.__init__

if not default_batch_sizes.get(Resource):
    _LOG.warning() # warn that the user will be defaulting to our default
    default_batch_sizes[Resource] = 50 # or some arbritrary
self.options["default_batch_size"] = default_batch_sizes  # or whatever the argument name is



...
# now inside fetch pages

        if maxResults:
            page_params["maxResults"] = maxResults
        else: # will need to set batch size to get all resources
            page_params["maxResults"] = self.options["default_batch_size"].get(item_type, self.options["default_batch_size"].get(Resource))
            _LOG.verbose(f"Batch size for {item_type.__name__} set to {page_params["maxResults"]}")
```

There

jira/client.py
adehad: hmm, the thing is this is already done at the top of the function where we actually use the batch size...  maybe it needs to be `page_params = page_params.copy()`

I'll need to step by step run your test through a debugger to make sure this is what we want

rynkk:
I agree, not having to specify the batch_size for every request is less error-prone and more user-friendly, while the dict-config allows the required configurability.

I implemented the requested changes with some addition:
It is possible to set a Resource's mapped batch-size to `None`, which means, that no batching on our side is done, but instead we let the JIRA-backend handle it.


---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

tests/tests.py
```diff
@@ -368,6 +371,25 @@ def test_fetch_pages(
         )
 
 
+@parameterized.expand(
```
adehad, line 374: can just pytest.mark.parametrize directly here (preferred)

jira/client.py
```diff
@@ -827,6 +830,16 @@ def _get_items_from_page(
...
             raise KeyError(str(e) + " : " + json.dumps(resource))
 
+    def _get_batch_size(self, item_type: Type[ResourceType]) -> Optional[int]:
+        batch_sizes: Dict[Type[Resource], Optional[int]] = self._options[
+            "default_batch_size"
+        ]
+        try:
+            item_type_batch_size = batch_sizes[item_type]
+        except KeyError:
+            item_type_batch_size = batch_sizes.get(Resource, None)
```
adehad, line 840: should be supplemented in the docstring of this function
```python
"""
Returns the batch size for the given resource type from the options.

Returns:
   Optional[int]: The batch size to use. When the configured batch size is None, the batch size is determined by the Jira backend.
"""
```


May be worth also a quick inline comment

```suggestion
            # None to let the Jira backend determine the batch size
            item_type_batch_size = batch_sizes.get(Resource, None)
```

jira/client.py
```diff
@@ -720,7 +723,7 @@ def _fetch_pages(
             page_params["startAt"] = startAt
         if maxResults:
             page_params["maxResults"] = maxResults
-        elif batch_size is not None:
+        elif batch_size := self._get_batch_size(item_type):
             page_params["maxResults"] = batch_size
```
adehad, line 727:
```suggestion
        else:
            page_params["maxResults"] = self._get_batch_size(item_type)
```

otherwise won't correctly apply when the batch size is configured to None

Probably a need a test for that

adehad (changes requested):
Looking great!

`fetch_pages()` and the way we handle `async` needs its own refactoring at some point so don't worry about it for now.

adehad:
@rynkk you should also rebase/merge the latest main if you haven't already, as we've fixed some build problems

jira/client.py
rynkk: @adehad, I tried before and apparently the API wont accept "maxResults=null". That's why I specifically only set it when its truthy, so we ignore `None`. Am I mistaken or missing something; for the old maxResults=False we did not set `maxResults` either?


---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

tests/tests.py
```diff
@@ -368,6 +371,25 @@ def test_fetch_pages(
...
+    ]
+)
+def test_get_batch_size(default_batch_sizes, item_type, expected):
+    class BatchSizeMock:
+        def __init__(self, batch_sizes):
+            self._options = {"default_batch_size": batch_sizes}
+
+    batch_size_mock = BatchSizeMock(default_batch_sizes)
+
+    assert JIRA._get_batch_size(batch_size_mock, item_type) == expected
```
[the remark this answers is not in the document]
rynkk: Good point, found a bug because of it, too. (copy.copy -> copy.deepcopy because we have a nested dict now)

rynkk:
Changes implemented, some modifications of my own, as well. Tests pass, docs do not. Some issue not being able to map `ResourceType`. Not sure what the cause is.


---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

tests/tests.py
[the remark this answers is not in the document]
adehad: Excellent

adehad (commented):
Good stuff, almost ready, let's see if we can fix the build

jira/client.py
adehad: hmm, not a huge fan of this, but it works and doesn't really do much harm so let's keep it in

jira/client.py
```diff
@@ -366,6 +366,10 @@ class JIRA:
             # 'Expires': 'Thu, 01 Jan 1970 00:00:00 GMT'
             "X-Atlassian-Token": "no-check",
         },
+        "default_batch_size": {
+            Resource: 100,
+            Issue: 500,
```
adehad, line 371:
```suggestion
```

Let's not make this the default for now, obviously feel free to increase this for your own system

adehad (changes requested):
Just 1 change and then I'm happy, @studioj any further thoughts?

tests/tests.py
```diff
@@ -296,40 +300,101 @@ def setUp(self):
...
+        (
+            {Resource: 1, Issue: None},
+            Issue,
+            None,
+        ),  # None -> backend determines batch-size
+        ({Resource: 1}, Dashboard, 1),  # fallback working correctly
+        ({}, Issue, 500),  # default value for Issue
+        ({}, Resource, 100),  # default value for everything else
+    ],
+)
```
adehad, line 391: (nitpick, non-blocking) rather than using comments to specify conditions the preferred way is using the [pytest ids=](https://docs.pytest.org/en/6.2.x/example/parametrize.html#different-options-for-test-ids)

```suggestion
    ids=["modify Issue default", "modify Resource default", "let backend decide for Issue", "fallback", "default for Issue", "default value for everything else"],
)
```

the idea being it improves the readability of the test report, less of a counting game when a test fails

rynkk: Didn't know of the `ids`-arg. Nice!

rynkk:
Default removed, tests & docs adjusted :-)

---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

jira/client.py
```diff
@@ -473,11 +472,12 @@ def __init__(
...
-              `Resource` is used as a fallback for every item type not specified. If a item type is mapped to `None` no fallback occurs, instead the JIRA-Backend will use its default batch-size.
-              By default Issues will be queried in batches of 500, all other item types (`Resource`-key) will be queried in batches of 100.
-              Above standard configuration could be manually supplied by setting this parameter to: ``{Issue: 500, Resource: 100}``
-              (Default: None)
+            default_batch_sizes (Optional[Dict[Type[Resource], Optional[int]]]): Manually specify the batch-sizes for
+              the paginated retrieval of different item types. `Resource` is used as a fallback for every item type not
+              specified. If a item type is mapped to `None` no fallback occurs, instead the JIRA-Backend will use its
+              default batch-size. By default all Resources will be queried in batches of 100. E.g., setting this to
+              ``{Issue: 500, Resources: None}`` will make :py:meth:`search_issues` query Issues in batches of 500, while
+              every other item type's batch-size will be controlled by the backend. (Default: None)
```
adehad, line 480: really excellent documentation here

adehad (dismissed):
Awesome, looks good to me, will wait a bit to see if @studioj has any comments.

Thanks for pushing through with this.

tests/tests.py
adehad: It is nice, I much prefer it compared to `parameterized.expand` way, where you can get away with creating an argument called "name" in the test function, or use a naming function, it feels a bit clunkier.

jira/client.py
```diff
@@ -716,6 +730,8 @@ def _fetch_pages(
             page_params["startAt"] = startAt
         if maxResults:
             page_params["maxResults"] = maxResults
+        elif batch_size := self._get_batch_size(item_type):
```
studioj, line 733: is this the first walrus operator we're using?
somebody has to go first :)
I must admit I'm not used to it yet

studioj (approved):
seems pretty thourough although I'm counting on @adehad for the details :P

jira/client.py
rynkk: It's an honor (for some). Just have to be very careful about braces when using it..

rynkk:
Found typos in the `default_batch_size`-docstring at the last glance

adehad:
Sorry @rynkk haven't had much free time, am having a bit more now so will try and push this through

PR #1396, opened 2022-06-05
making docstrings more uniform

studioj:
As stated in comment https://github.com/pycontribs/jira/pull/1387#discussion_r889697551

Quite a lot of rework could be done
- commits split up per file
- commits split up in 4 catagories
-- cleanup the old line length limited docstrings and comments
-- switch all docstrings explaining a bool so they explain the scenario in case of True
-- made all return statements in the docstrings the same spaces and whitelines included
-- set the typehint for issue correct accross all methods (it was a bit here and there and everywhere)
-- fixed some specific methods docstrings

I suggest reviewers to walk over it commit per commit ... that will be easier for context
Also I suggest to keep review time relatively short ... :-) merge conflicts are lurking behind the corner


studioj:
I just noticed my line lenght in my ide is at 140 => maybe i should again drop those changes  ... :( was soo happy not too many lines were added by the rework

jira/resources.py
```diff
@@ -315,15 +307,15 @@ def update(
...
-        Keyword arguments are marshalled into a dict before being sent. If this
-        resource doesn't support ``PUT``, a :py:exc:`.JIRAError` will be raised; subclasses that specialize this method
-        will only raise errors in case of user error.
+        Keyword arguments are marshalled into a dict before being sent. If this resource doesn't support ``PUT``, a :py:exc:`.JIRAError`
+        will be raised; subclasses that specialize this method will only raise errors in case of user error.
 
         Args:
             fields (Optional[Dict[str, Any]]): Fields which should be updated for the object.
-            async_ (bool): If true the request will be added to the queue so it can be executed later using async_run()
+            async_ (bool): The request will be added to the queue, so it can be executed later using async_run()
```
adehad, line 315: I would prefer all bool arguments to start with the word "True"

e.g.

```suggestion
            async_ (bool): True adds the request to async queue, so it can be executed later using async_run()
```

as I this ensures the meaning is crystal clear, prevents bad refactors

adehad (changes requested):
Regarding line length, I think it is probably fine for docstrings as we generate sphinx docs, but for inline comments wrapping to 72/88 is preferred. As when you are using a two column code layout it makes it easier.

jira/resources.py
studioj: makes sense in a way :D

studioj:
> Regarding line length, I think it is probably fine for docstrings as we generate sphinx docs, but for inline comments wrapping to 72/88 is preferred. As when you are using a two column code layout it makes it easier.

TBH keeping the line length consistent seems better for me ... i'll rework that part again for 88 columns

jira/client.py
```diff
@@ -176,8 +176,9 @@ def __iter__(self) -> Iterator[ResourceType]:
         return super().__iter__()
 
     # fmt: off
-    # The mypy error we ignore is about returning a contravariant type. As this class is a List of a generic 'Resource' class
-    # this is the right way to specify that the output is the same as which the class was initialized with.
+    # The mypy error we ignore is about returning a contravariant type. As this class
+    # is a List of a generic 'Resource' class this is the right way to specify that the
+    # output is the same as which the class was initialized with.
```
adehad, line 181:
```suggestion
    # The mypy error we ignore is about returning a contravariant type. 
    # As this class is a List of a generic 'Resource' class
    # this is the right way to specify that the output is the same
    # as which the class was initialized with.
```

adehad (changes requested):
These changes look like they are just for the sake of making the line lengths more consistent.

I would change the title to "make inline comments have a similar length"

The main problem I have with this, is that most of these inline comments are for developers only so breaking the reading flow should be avoided. Meaning we should try and break the flow of the comment when it makes sense to do so, not in the middle of a phrase.

I've put a few examples of better ways to break the flow of the comment. I would suggest try reading the comment out loud, a new line breaks the flow of the thought/sentence just like a comma/full-stop

jira/resources.py
```diff
@@ -138,8 +137,8 @@ class Resource:
         "closed",
     )
 
-    # A list of properties that should uniquely identify a Resource object
-    # Each of these properties should be hashable, usually strings
+    # A list of properties that should uniquely identify a Resource object Each of
+    # these properties should be hashable, usually strings
```
adehad, line 141: Why do we add the "Each of" to a new line here? if it made sense to do so we should continue the right case for a sentence, i.e. "each of"

jira/resources.py
```diff
@@ -781,15 +770,15 @@ def update(  # type: ignore[override]
 
         Args:
             fields (Optional[Dict[str, Any]]): DEPRECATED => a comment doesn't have fields
-            async_ (Optional[bool]): If True the request will be added to the queue, so it can be executed later using async_run()
+            async_ (Optional[bool]): to add the request to the queue, so it can be executed later using async_run()
```
adehad, line 773: prefer `True` definition first,

jira/resources.py
```diff
@@ -648,23 +641,21 @@ def update(  # type: ignore[override] # incompatible supertype ignored
...
-        will be ignored.
+        Each keyword argument (other than the predefined ones) is treated as a field name and the argument's value is treated as the
+        intended value for that field -- if the fields argument is used, all other keyword arguments will be ignored.
 
-        Jira projects may contain many different issue types. Some issue screens have different requirements for
-        fields in an issue. This information is available through the :py:meth:`.JIRA.editmeta` method. Further examples
-        are available here: https://developer.atlassian.com/display/JIRADEV/JIRA+REST+API+Example+-+Edit+issues
+        Jira projects may contain many different issue types. Some issue screens have different requirements for fields in an issue.
+        This information is available through the :py:meth:`.JIRA.editmeta` method.
+        Further examples are available here: https://developer.atlassian.com/display/JIRADEV/JIRA+REST+API+Example+-+Edit+issues
```
adehad, line 649: this is good

adehad (changes requested):
whoops forgot to add the more examples, I wouldn't be too fussy with the actual docstrings as they are parsed by sphinx anyway

looks like some places are good, I'll give this a closer look another time

jira/resources.py
studioj: ah forgot the optional bools

jira/client.py
studioj: lol isnt this nit picking alot :-D

I can also revert all changes in inline comments as i think all other changes are without remarks

jira/client.py
```diff
@@ -2910,26 +2892,24 @@ def search_issues(
...
             Union[Dict,ResultList]: Dict if ``json_result=True``
-
         """
         if isinstance(fields, str):
             fields = fields.split(",")
 
-        # this will translate JQL field names to REST API Name
-        # most people do know the JQL names so this will help them use the API easier
+        # this will translate JQL field names to REST API Name most people do know the
+        # JQL names so this will help them use the API easier
```
adehad, line 2912: I would prefer the original in this case

the first line reads like a summary "this will translate JQL field names to REST API Name"
The second line is then treated more as a way to add more details on why.

By combining them together like this it makes it harder to see that

jira/client.py
```diff
@@ -3090,13 +3076,11 @@ def search_assignable_users_for_projects(
         Args:
             username (str): A string to match usernames against
             projectKeys (str): Comma-separated list of project keys to check for issue assignment permissions
-            startAt (int): Index of the first user to return (Default: 0)
-            maxResults (int): Maximum number of users to return.
-              If maxResults evaluates as False, it will try to get all users in batches. (Default: 50)
+            startAt (int): Index of the first user to return (Default: ``0``)
+            maxResults (int): Maximum number of users to return. If maxResults evaluates as False, it will try to get all users in batches. (Default: ``50``)
```
adehad, line 3080: here is another example which follows my previous comment of summary and detail lines.
It also reaches 157 characters wide, which feels a bit long even for me.

But in this case I don't mind as these docstrings are mainly going to be read in the html sphinx docs.

adehad (commented):
sorry for taking a long time for the full review.

had a chance to look through client.py and also the autogenerated docs, looks good and nothing seems out of place.

adehad (dismissed):
minor error I spotted, but otherwise looks good
I hopefully will address the lint issue in a separate PR

adehad (approved):

PR #1401, opened 2022-06-15
fixed a case for search_assignable_users_for_issues where username an…

niteshthakur:
This change fixes an issue where `username` and `query` arguments are not passed for method `search_assignable_users_for_issues`

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3086,10 +3086,11 @@ class JIRA:
         Returns:
             ResultList
         """
+        params = {}
         if username is not None:
-            params = {"username": username}
+            params["username"] = username
         if query is not None:
-            params = {"query": query}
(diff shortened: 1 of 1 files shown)


niteshthakur:
need to remove check `if not username and not query:`

studioj:
looks legit , could you add a test which validates this?

adehad:
> need to remove check `if not username and not query:`

We still need this, this is what provides a useful error message to the user.

Technically this is also fixed by the changes in: https://github.com/pycontribs/jira/pull/1394 - where this check is moved to before the arguments are used

adehad:
Closing as #1394 is now merged.

@niteshthakur if you still have this problem do let me know and we can reopen this PR or create a new PR to address this

PR #1402, opened 2022-06-17
bug: config reading options editing JIRA.DEFAULT_OPTIONS

brezelman:
bug: don't edit JIRA.DEFAULT_OPTIONS, instead work with shallow copy.


rynkk:
Good find, though I'm curious how you encountered this problem. Did you instantiate one JIRA-instance using `get_jira` and another one using the non-config way in your program?

Also, once #1394 gets merged this might still cause problems, as the config is then a nested dict. Might want to change it to a ```copy.deepcopy```.
___
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

Licensed under [BSD-2-Clause license](https://github.com/pycontribs/jira/blob/main/LICENSE)</sub>

adehad:
#1394 is now merged, let's see if we can get a test case in while we are at it so we know we are fixing the right thing!

Let me know if you need any pointers, the tests can get a bit funky, my suggestion would be to put them in: https://github.com/pycontribs/jira/blob/main/tests/test_client.py

studioj:
@brezelman any update? Thank you for contributing

brezelman:
> @brezelman any update? Thank you for contributing

I'm sorry. I'm not able to provide a test implementation.

PR #1404, opened 2022-06-25
extend the examples with a more thorough example for getting attachments from a specific comment

studioj:


docs/examples.rst
```diff
@@ -336,6 +336,16 @@ Adding, editing and deleting comments is similarly straightforward::
     comment.update(body='updated comment body but no mail notification', notify=False)
     comment.delete()
 
+Get all images from a comment::
+
+    issue = jira.issue('JRA-1330')
+    regex_for_png = re.compile(r'\!(\S+?\.jpg)\|?\S*?\!')
```
adehad, line 342:
```suggestion
    regex_for_png = re.compile(r'\!(\S+?\.png)\|?\S*?\!')
```

? or maybe can change the variable name to jpg, whichever is fine with me

adehad (changes requested):

docs/examples.rst
studioj: validated on cloud with copy paste of a jpeg image ... I could add a few image types

PR #1405, opened 2022-06-25
improve explanation of exception when two sprints with the same name are found

studioj:


jira/client.py
```diff
@@ -4516,13 +4516,27 @@ def sprints(
             self.AGILE_BASE_URL,
         )
 
-    def sprints_by_name(self, id, extended=False):
+    def sprints_by_name(self, id, extended=False, state=None):
```
adehad, line 4519: shall we also add typehints here?

```suggestion
    def sprints_by_name(self, id: Union[str,int], extended: bool = False, state: str = None) -> Dict[str, Dict[str, Any]]:
```

the output type is a bit weird for this function (ie. could spit out the sprint object itself?)... but probably beyond this PR

jira/client.py
```diff
@@ -4516,13 +4516,27 @@ def sprints(
...
+        """Get a dictionary of sprint Resources where the name of the sprint is the key.
+
+        Args:
+            board_id (int): the board to get sprints from
+            extended (bool): Deprecated.
+            state (str): Filters results to sprints in specified states. Valid values: `future`, `active`, `closed`.
+              You can define multiple states separated by commas
```
adehad, line 4526: is this case sensitive?

tests/resources/test_sprint.py
```diff
@@ -97,3 +100,11 @@ def test_move_issue_to_backlog(self):
             # THEN: There is no longer the sprint assigned
             updated_issue_1 = self.jira.issue(self.issue_1)
             assert updated_issue_1.get_field(self._sprint_customfield()) is None
+
+    def test_two_sprints_with_the_same_name_raise_a_jirra_error_when_sprints_by_name_is_called(
```
adehad, line 104:
```suggestion
    def test_two_sprints_with_the_same_name_raise_a_jira_error_when_sprints_by_name_is_called(
```

jira/client.py
```diff
@@ -4516,13 +4516,27 @@ def sprints(
...
+
+        Returns:
+            dict: dictionary of sprints with the sprint name as key
```
adehad, line 4529:
```suggestion
            Dict[str, Dict[str, Any]]: dictionary of sprints with the sprint name as key
```

adehad (changes requested):

jira/client.py
studioj: no clue didnt validate this just copied from sprints()

adehad: beyond this PR then

Issue #1430, opened 2022-07-18
`search_issues` breaking change in 3.3.0


dtmistry:
### Bug summary

Prior to `3.3.0` `search_issues` allowed `fields=None`. E.g 

```python
fields=None
max_results=10
client.search_issues("search_string", maxResults=max_results, fields=fields)
```

With the `3.3.0` release, `fields` was defaulted to `*all`. But the `None` [check was removed](https://github.com/pycontribs/jira/pull/1360/files#diff-442d14b20d591afd0c3de830d2803e30d4816eedc708f9ad129316aa7e5a4a7cL2898), resulting in the below error -

```bash
File "/opt/python/python-3.8.5/lib/python3.8/site-packages/jira/client.py", line 2935, in search_issues
    for i, field in enumerate(fields):
TypeError: 'NoneType' object is not iterable
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

8.22.4

### jira-python version

3.3.0

### Python Interpreter version

3.8.5

### Which operating systems have you used?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA
# 2. When I call the function with argument x

fields=None
max_results=10
client.search_issues("search_string", maxResults=max_results, fields=fields)

# 3.
I get a `TypeError: 'NoneType' object is not iterable`
```

### Stack trace

```python
File "/opt/python/python-3.8.5/lib/python3.8/site-packages/jira/client.py", line 2935, in search_issues
    for i, field in enumerate(fields):
TypeError: 'NoneType' object is not iterable
```

### Expected behaviour

Search results are returned without any `TypeError`

studioj:
that was probably an unintended change... thx for reporting

studioj:
aha it isnt 
None fetched all .... 
@adehad I believe *all to be more correct
should we go for a deprecation warning and state that in 3.4 this will be disabled? or the behaviour will change to the behaviour of fields = []

adehad:
I think we should handle the `None` case explicitly and pass the fields as `*all`
Technically when passing `None` we could exclude the parameter it in the request, but then it would only include navugatable fields, but I think let's use all to make it easier for everyone 

We should make a patch release with this in.
I'm currently don't have machine to develop on, but happy to review any PR.

A test case with explicit None would also be ideal.

[...]

PR #1434, opened 2022-07-26
ensure `None` is handled correctly in `search_issues()`

adehad:
Fixes: #1430

FYI @dtmistry


adehad:
docs failing is due to:
https://github.com/sphinx-doc/sphinx/issues/10701

Probably worth us removing the tox doc building from our GHA as we have the dedicated readthedocs GHA running

PR #1436, opened 2022-07-27
update sphinx to fix builds and update RTD config due to deprecations

adehad:
See the deprecation notice:
https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version

Additionally used the constraints file for the rtd action

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,10 +7,12 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.8"
+  jobs:
+    pre_install:
+      - export PIP_CONSTRAINT=constraints.txt
 
 python:
   install:
-    # PIP_CONSTRAINT has been set as env variable in the RTD web interface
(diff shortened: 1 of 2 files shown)


adehad:
sphinx-rtd-theme seems to be slow at supporting docutils (https://github.com/readthedocs/sphinx_rtd_theme/issues/1302), so had to downgrade docutils
https://readthedocs.org/projects/jira/builds/17557640/

Evidence that this works is here: https://readthedocs.org/projects/jira/builds/17557640/
Where I had locked to the broken version 5.1.0 (latest was the fixed 5.1.1)

adehad:
no idea how the tests broke with these changes

adehad:
Some conversations about `PIP_CONSTRAINT` and the workaround can be found in the feature request in: https://github.com/readthedocs/readthedocs.org/issues/7258#issuecomment-1197786076

Issue #1437, opened 2022-07-27
jira.client.JIRA.add_attachment - 500 Internal Server Error


mtripoli:
### Bug summary

This problem occurs when the `add_attachment` method is called with a `str` value representing the file name assigned to the `attachment` parameter and no value assigned to the `filename` parameter.

Here is the salient message from the server log:
```
 Cannot create attachment without a filename - inline content? See http://jira.atlassian.com/browse/JRA-10825 (file=attachment-8292840129205496726.tmp).
```
(This also induces an NPE in the server)

diff that highlights the problem area:
```
diff --git a/jira/client.py b/jira/client.py
index de0a110..237cfca 100644
--- a/jira/client.py
+++ b/jira/client.py
@@ -1005,8 +1005,8 @@ class JIRA:
                 )

         fname = filename
-        if not fname and isinstance(attachment, BufferedReader):
-            fname = os.path.basename(attachment.name)
+        if not fname and isinstance(attachment_io, BufferedReader):
+            fname = os.path.basename(attachment_io.name)

         def generate_multipartencoded_request_args() -> Tuple[
             MultipartEncoder, CaseInsensitiveDict
```

The problem was introduced in 6e513952c854acf67f78bce8c281c0546489a346 but only recently merged and released.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

v8.13.22#813022

### jira-python version

3.3.0

### Python Interpreter version

3.8

### Which operating systems have you used?

- [X] Linux
- [X] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
client: jira.client.JIRA
issue: jira.resources.Issue
# 2. These calls throw JiraError the exceptions in the Stack trace section
client.add_attachment(issue, '/path/to/file.txt')
or this
client.add_attachment(issue, attachment='/path/to/file.txt')
# 3.
...
```

### Stack trace

```python
Traceback (most recent call last):
  File "/Users/mtripoli/.newt-cache/pyenv/versions/cpython-3.8.13/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/Users/mtripoli/src/NCO/ocjira/.venv/lib/python3.8/site-packages/jira/client.py", line 121, in wrapper
    result = func(*arg_list, **kwargs)
[...]

PR #1438, opened 2022-07-28
derive attachment fname from BufferedReader object (#1437) 

mtripoli:
Offering a fix for #1437.

### DIFF
--- a/tests/resources/test_attachment.py
+++ b/tests/resources/test_attachment.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 from tests.conftest import TEST_ATTACH_PATH, JiraTestCase
 
@@ -30,6 +31,14 @@ class AttachmentTests(JiraTestCase):
 
     def test_2_add_remove_attachment_using_filename(self):
         issue = self.jira.issue(self.issue_1)
+        attachment_no_filename_specified = self.jira.add_attachment(
+            issue,
+            TEST_ATTACH_PATH,
+        )
+        new_attachment = self.jira.attachment(attachment_no_filename_specified.id)
+        msg = f"attachment, no filename specified {new_attachment.__dict__} of issue {issue}"
+        self.assertEqual(new_attachment.filename, Path(TEST_ATTACH_PATH).name, msg=msg)
+        #
         attachment = self.jira.add_attachment(
             issue, TEST_ATTACH_PATH, "new test attachment"
         )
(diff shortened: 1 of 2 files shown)


adehad (changes requested):
Thanks @mtripoli, change looks good to me I think
test failure isn't related to this, but could I ask if you can add a test case here:
https://github.com/pycontribs/jira/blob/main/tests/resources/test_attachment.py

Can be a very simple one we we just include the first argument and if no exception is raised and the attachment is found we are all good. Hopefully the existing cases are sufficient to help you crack on, if not let me know and I can write up a test case

mtripoli:
@adehad , done. I totally missed the tests/resources dir otherwise I would have provided the coverage for the change on the first go.

tests/resources/test_attachment.py
```diff
@@ -30,6 +31,14 @@ def test_1_add_remove_attachment_using_filestream(self):
...
     def test_2_add_remove_attachment_using_filename(self):
         issue = self.jira.issue(self.issue_1)
+        attachment_no_filename_specified = self.jira.add_attachment(
+            issue,
+            TEST_ATTACH_PATH,
+        )
+        new_attachment = self.jira.attachment(attachment_no_filename_specified.id)
+        msg = f"attachment, no filename specified {new_attachment.__dict__} of issue {issue}"
+        self.assertEqual(new_attachment.filename, Path(TEST_ATTACH_PATH).name, msg=msg)
+        #
```
adehad, line 41: excellent test case, however, let's put this into a new method, the test name doesn't make sense here

We don't need to add a number to the tests so just something like:


```suggestion
    def test_add_attachment_without_filename(self):
        issue = self.jira.issue(self.issue_1)
        attachment_no_filename_specified = self.jira.add_attachment(
            issue=issue,
            attachment=TEST_ATTACH_PATH,
            filename=None,
        )
        new_attachment = self.jira.attachment(attachment_no_filename_specified.id)
        msg = f"attachment, no filename specified {new_attachment.__dict__} of issue {issue}"
        assert new_attachment.filename == Path(TEST_ATTACH_PATH).name, msg

    def test_2_add_remove_attachment_using_filename(self):
        issue = self.jira.issue(self.issue_1)
```

(I've also moved away from the self.assert methods, one day we will clean out the legacy unittest syntax)

adehad (changes requested):

adehad:
> @adehad , done. I totally missed the tests/resources dir otherwise I would have provided the coverage for the change on the first go.

No problem at all, the tests are definitely hard to understand for new contributors, its on my list to update the contributing guide to clarify things

adehad (approved):
Awesome, thanks for this !

PR #1439, opened 2022-07-31
Add support for IssueProperty resource

turran:
Add support for adding, removing and getting issue properties as described in https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-properties/#api-group-issue-properties

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -66,6 +66,7 @@ from jira.resources import (
     Issue,
     IssueLink,
     IssueLinkType,
+    IssueProperty,
     IssueSecurityLevelScheme,
     IssueType,
     IssueTypeScheme,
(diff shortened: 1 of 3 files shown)


jira/client.py
```diff
@@ -2396,6 +2397,49 @@ def add_worklog(
...
+    def issue_properties(self, issue: str) -> List[IssueProperty]:
+        """Get a list of issue property Resource from the server for an issue.
+
+        Args:
+            issue (str): ID or key of the issue to get properties from
+
+        Returns:
+            List[IssueProperty]
+        """
+        r_json = self._get_json("issue/" + str(issue) + "/properties")
```
adehad, line 2412:
```suggestion
        r_json = self._get_json(f"issue/{issue}/properties")
```

We prefer f-strings

jira/client.py
```diff
@@ -2396,6 +2397,49 @@ def add_worklog(
...
+
+        Args:
+            issue (str): ID or key of the issue to set the property to
+            key (str): Key of the property to set
+            data: The data to set for the property
+        Returns:
+            Response
+        """
+
+        url = self._get_url("issue/" + issue + "/properties/" + key)
```
adehad, line 2440:
```suggestion
        url = self._get_url(f"issue/{issue}/properties/{key}")
```

(f-strings)

jira/resources.py
```diff
@@ -290,6 +291,20 @@ def find(
...
             path = self._resource.format(id)
         url = self._get_url(path)
+        self._find_by_url(url, params)
+
+    def _find_by_url(
+        self,
+        url: str,
+        params: Optional[Dict[str, str]] = None,
+    ):
+        """Finds a resource on the specified url.
```
adehad, line 301: Could you elaborate in this docstring, as there is no return value, but it instead loads the url.
Perhaps the name itself can be changed.

I'm also not sure what `.find()`

jira/resources.py
```diff
@@ -290,6 +291,20 @@ def find(
...
+
+        Args:
+            url (str): url
+            params (Optional[Dict[str, str]]): params
+
```
adehad, line 306:
```suggestion
```

adehad (changes requested):
Excellent stuff here, few minor changes

jira/client.py
turran: Understood. Wasn't sure which one to use given that the code had both ways. Fixing it.

jira/resources.py
turran: Sure. The description is just a copy of the original _find(), I had to split it on two: _find() -> _find_by_url() in order to just overload one in the IssueProperty resource class. The motivation is that, despite other resources, IssueProperty JSON description does not have a "self" keyword and that has to be set by the passed in URL of the request.

jira/client.py
adehad: ahh, we tried to cleanse the code base using flynt or something, but maybe we didn't do a good enough job there, sorry

jira/resources.py
adehad: Cool, that is interesting. 

I'm a bit cautious of the additional request that needs to be made for this edge case. But tests are happy so probably will be alright, I'll give it a think but should be good

adehad (approved):

adehad:
will be trying to get some bugfixes in for a patch release, so will aim to merge this after that

adehad:
Thanks for the (very high quality) PR ! Now included in `3.4.0`

Sorry the original one got buried with the CI issues, we should be in a better place now for quicker reviews and deployments!

PR #1443, opened 2022-08-01
Remove setup.py and tidy build scripts

adehad:
## Change Summary
1. Remove `setup.py` (would have kept it if we were using `pyproject.toml`, [per the guidance from setuptools docs to preserve editable install behaviour across older pip's](https://github.com/pypa/setuptools/blob/5179e8fcd89657d0f9b3660e2a7ec6f6eec9ce36/docs/userguide/pyproject_config.rst))
2. Move tool config to `pyproject.toml` where possible, only flake8 remains in `setup.cfg`
3. Remove other old files incl. Makefile and other tools configs not used - can be added again if appropriate
4. For whatever reason had to update the logic around keyring password. The previous output captured used to be `No password provided\n assert ''\n`, but somehow became `No password provided\n`. I've moved this to a simple `ValueError` as assertion's aren't really a great way to surface these kinds of problems. (e.g. if someone runs with `-O` for whatever reason)
5. Consolidated project options in `setup.cfg` removing duplicates that made their way to `pyproject.toml` ( and the removed `setup.py`)

Testing the git archive can be performed using:
```
pip install git+https://github.com/pycontribs/jira.git@feature/pyproject
python -c "import jira; print(jira.__version__)"
```


ssbarnea:
@adehad That is one dangerous move as the full-pyproject support is very new in pip, likely will cause problems for many people.

ssbarnea:
On the other hand, I do fully support removal of `setup.py`.

adehad:
> @adehad That is one dangerous move as the full-pyproject support is very new in pip, likely will cause problems for many people.

My understanding is pretty limited, but I would have thought it wouldn't be a problem to most users as they would use the pypi wheels? Would be great to know a bit about what kinds of failures we can expect to see from your experience @ssbarnea 

I know asotille also has strong opinions on this topic and prefers not even having the presence of the pyproject file as pip completely changes its behaviour when it sees it....

ssbarnea:
Hehe... most of his opinions are strong. I will try to give a hand later this week, now I am quite busy dealing with too many tasks in parallel.

For the moment I would recommend matching what we do on https://github.com/ansible/ansible-lint/blob/main/tox.ini -- where we removed setup.py, but did not fully migrate to pyproject.toml, making full use of setuptools-scm.

ssbarnea (changes requested):
Do NOT move to pyproject.toml, editable installs still do not work. There is a setuptools PR that is open for fixing them.

adehad:
re-running to check if codecov will fix itsef

adehad:
@ssbarnea would be great if you have a chance for a second look on this

ssbarnea (approved):

Issue #1274, opened 2022-01-04
Unable to use JIRA.create_issue in client.py with integer `issuetype`


martinlocklear:
### Bug summary

At some point a bug was introduced which prevents us from using `def create_issue(` to create an issue of a specific `issuetype`, when you want to indicate the issue type using an int. This is especially problematic for issue types where the string `issuetype` doesn't unambiguously identify the necessary issue type. This causes def `issue_type_by_name(` to raise a KeyError.

There's currently a workaround (to use the bulk `def create_issues(`, which doesn't contain this bug.)

It looks like this issue was introduced originally in this commit: e829da6980980e0291c5787442524edc822a94fb

I created a [draft PR](https://github.com/pycontribs/jira/pull/1273) with a fix, but I'm unable to get the tests running locally so I can't meet the [requirements](https://github.com/pycontribs/jira/blob/09ece94f3cae7e6d0becfa87d77d6a05ce01cdf6/docs/contributing.rst) for submitting the pr (having test coverage for the issue)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

3.1.1

### Python Interpreter version

3.8

### Which operating systems have you used?

- [ ] Linux
- [X] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA
# 2. When I call the function with argument x
jira.create_issue(issuetype=10007)
# 3.  `create_issue` looks for the issue by int ID, but then ignores that and treats any `issuetype` parameter as a string.
```

### Stack trace

```python
NA
```

### Expected behaviour

I expect that the value passed with be treated as an int

PR #1445, opened 2022-08-03
Fix bug preventing use of an int for issuetype in def create_issue

rynkk:
Continuation of PR #1273 by adding tests

Apparent bug causes supplied issuetype for `create_issue` to always be treated as the name and not as the id, even if an integer was passed.

`name` should be used if a string was passed, e.g., `"1234"`,
`id` should be used, if an integer was passed, e.g., `1234`

See issue #1274

---
The program was tested solely for our own use cases, which might differ from yours.

Jannik Meinecke <jannik.meinecke@mercedes-benz.com> on behalf of MBition GmbH.
https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md
Licensed under MIT


jira/client.py
```diff
@@ -1502,7 +1502,7 @@ def create_issue(
         p = data["fields"]["issuetype"]
         if isinstance(p, int):
             data["fields"]["issuetype"] = {"id": p}
-        if isinstance(p, (str, int)):
+        if isinstance(p, str):
```
adehad, line 1505: should this be an elif?

adehad (commented):
thanks for this

jira/client.py
rynkk: I reckon it could be an elif. Thanks for the headsup.

---
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

adehad (approved):

PR #1448, opened 2022-08-08
Fix using str as issue type with create_issue when issue type with that name exists in many projects

### LINKED ISSUE #1444: Issue type '{name}' appears more than once on `create_issue`
### Bug summary

Updating from version 3.0 to 3.3.1, we started seeing `Issue type 'User Story' appears more than once` when calling `create_issue` with a project and `User Story` issue type.

Example call:
```python
jira.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": "User Story",
        "summary": "test",
    }
)
```

Looking at the code, a call is made to https://myorg.atlassian.net/rest/api/3/issuetype to retrieve all issue types. A solution would be to call https://myorg.atlassian.net/rest/api/3/project/PROJECTKEY/. I assume a project key or ID is provided since we cannot  create an issue outside of a project.

Something else I tried is:
```python
jira.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": 2,
        "summary": "test",
    }
)
```
but this fails with `Issue type '2' is unknown.`

This is probably caused by [this code in client.py](https://github.com/pycontribs/jira/blob/1f0a2ae74700f990778fb698f43953f1ad5aa4a7/jira/client.py#L1502)
```python
p = data["fields"]["issuetype"]
if isinstance(p, int):
    data["fields"]["issuetype"] = {"id": p}
if isinstance(p, (str, int)):
    data["fields"]["issuetype"] = {"id": self.issue_type_by_name(str(p)).id}
```

It seems to me that if we provide a int, e.g. 2, then we first change it to `{"id": 2}` and then still try to find an issue type named "2" which fails.

Is there a specific reason why that would be the case? Maybe we want to handle an issue type being named a number? (But then, users should still provide a str).

Lastly, a workaround is to it call like so (if you know the issue type id):
```python
jira.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": {"id": 2},
        "summary": "test",
    }
)
```

One can also list the issue types themself:
```python
project = jira_client.client.project("PROJECTKEY", expand="issueTypes")
issue_type_id = next(issue_type.id for issue_type in mun_project.issueTypes if issue_type.name == "User Story")
issue = jira_client.client.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": {"id": issue_type_id},
        "summary": "test",
    }
)
```

### Jira Instance type

### Jira instance version

_No response_

### jira-python version

3.3.1

3.8

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
# 0. Given a Jira instance with multiple projects with an issue type named `User Story`, but only one in the project `PROJECTKEY`
jira: JIRA
issue = jira.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": "User Story",
        "summary": "test",
    }
)
```

### Stack trace

```python
   File "test_jira.py", line 37, in test_jira_issue
     issue = jira_client.client.create_issue(
   File "/root/.cache/pypoetry/virtualenvs/something-BtA_kvLS-py3.8/lib/python3.8/site-packages/jira/client.py", line 1509, in create_issue
     data["fields"]["issuetype"] = {"id": self.issue_type_by_name(str(p)).id}
   File "/root/.cache/pypoetry/virtualenvs/something-BtA_kvLS-py3.8/lib/python3.8/site-packages/jira/client.py", line 2533, in issue_type_by_name
     raise KeyError(f"Issue type '{name}' appears more than once.")

KeyError: "Issue type 'User Story' appears more than once."
```

### Expected behaviour

`User Story` is looked up in the project `PROJECTKEY` and is unique

### Additional Context

_No response_

dblanchette:
Fixes #1444 

#1445 does solve the issue when providing the ID, but it is still not possible, as it was before, to provide a string like so:
```python
jira.create_issue(
    fields={
        "project": "PROJECTKEY",
        "issuetype": "User Story",
        "summary": "test",
    }
)
```

This is because [issue_type_by_name](https://github.com/pycontribs/jira/blob/00c37981511a923ee9cb41598118ceb04634fe0b/jira/client.py#L2521) does not take the project into consideration and thus will only work if the issuetype name exists only once across all projects in the Jira instance.

This situation is more frequent when using the team-managed projects that use their own issue types instead of sharing the standard ones.

I was not able to setup CI unfortunately, but I tested by calling my new code from another project which exhibited #1444

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1498,18 +1498,17 @@ class JIRA:
 
         p = data["fields"]["project"]
 
+        project_id = None
         if isinstance(p, (str, int)):
             project_id = self.project(str(p)).id
             data["fields"]["project"] = {"id": project_id}
-        else:
-            project_id = None
 
         p = data["fields"]["issuetype"]
         if isinstance(p, int):
             data["fields"]["issuetype"] = {"id": p}
         elif isinstance(p, str):
             data["fields"]["issuetype"] = {
-                "id": self.issue_type_by_name(str(p), project=str(project_id)).id
+                "id": self.issue_type_by_name(str(p), project=str(project_id) if project_id else None).id
             }
 
         url = self._get_url("issue")
--- a/jira/client.py
+++ b/jira/client.py
@@ -1508,7 +1508,9 @@ class JIRA:
             data["fields"]["issuetype"] = {"id": p}
         elif isinstance(p, str):
             data["fields"]["issuetype"] = {
-                "id": self.issue_type_by_name(str(p), project=str(project_id) if project_id else None).id
+                "id": self.issue_type_by_name(
+                    str(p), project=str(project_id) if project_id else None
+                ).id
             }
 
         url = self._get_url("issue")
--- a/jira/client.py
+++ b/jira/client.py
@@ -1499,13 +1499,18 @@ class JIRA:
         p = data["fields"]["project"]
 
         if isinstance(p, (str, int)):
-            data["fields"]["project"] = {"id": self.project(str(p)).id}
+            project_id = self.project(str(p)).id
+            data["fields"]["project"] = {"id": project_id}
+        else:
+            project_id = None
 
         p = data["fields"]["issuetype"]
         if isinstance(p, int):
             data["fields"]["issuetype"] = {"id": p}
         elif isinstance(p, str):
-            data["fields"]["issuetype"] = {"id": self.issue_type_by_name(str(p)).id}
+            data["fields"]["issuetype"] = {
+                "id": self.issue_type_by_name(str(p), project=str(project_id)).id
+            }
 
         url = self._get_url("issue")
         r = self._session.post(url, data=json.dumps(data))
@@ -2518,15 +2523,21 @@ class JIRA:
         """
         return self._find_for_resource(IssueType, id)
 
-    def issue_type_by_name(self, name: str) -> IssueType:
+    def issue_type_by_name(self, name: str, project: Optional[str] = None) -> IssueType:
         """
         Args:
             name (str): Name of the issue type
+            project (str): Key or ID of the project. If set, only issue types available for that project will be looked up
 
         Returns:
             IssueType
         """
-        matching_issue_types = [it for it in self.issue_types() if it.name == name]
+        if project:
+            issue_types = self.project(project, expand="issueTypes").issueTypes
+        else:
+            issue_types = self.issue_types()
+
+        matching_issue_types = [it for it in issue_types if it.name == name]
         if len(matching_issue_types) == 1:
             return matching_issue_types[0]
         elif len(matching_issue_types) == 0:
--- a/tests/resources/test_issue.py
+++ b/tests/resources/test_issue.py
@@ -268,6 +268,17 @@ class IssueTests(JiraTestCase):
         )
         self.assertEqual(issue.get_field("issuetype").name, dyn_it.name)
 
+    def test_create_issue_with_issue_type_name(self):
+        issue_types_resolved = self.jira.issue_types()
+        dyn_it = issue_types_resolved[0]
+
+        issue = self.jira.create_issue(
+            summary="Test issue created using a str issuetype",
+            project=self.project_b,
+            issuetype=dyn_it.name,
+        )
+        self.assertEqual(issue.get_field("issuetype").name, dyn_it.name)
+
     def test_update_with_fieldargs(self):
         issue = self.jira.create_issue(
             summary="Test issue for updating with fieldargs",


jira/client.py
```diff
@@ -1499,13 +1499,18 @@ def create_issue(
         p = data["fields"]["project"]
 
         if isinstance(p, (str, int)):
-            data["fields"]["project"] = {"id": self.project(str(p)).id}
+            project_id = self.project(str(p)).id
+            data["fields"]["project"] = {"id": project_id}
+        else:
+            project_id = None
```
adehad, line 1505: preference would be for this to be before the if statement. GitHub is a bit silly and doesn't let me suggest something as a "line is deleted" 

But it would look like:

```python
project_id = None
if condition:
    project_id = new_id
```

adehad (changes requested):
Awesome stuff, one minor comment

I'll also try and write up a test case for this before we merge (want to set a better standard in this project moving forward)

There is a devcontainer that might help you automagically get setup with CI in case you want to have a go too.

dblanchette:
I am sharing my experience in setting up the CI for your information/posterity below. I am on Ubuntu 20.04

It's great that you are working on improving the project quality. Thank you very much for the quick replies and work.

> I am running the container with `docker run -it -p 2990:2990 --name jira addono/jira-software-standalone` (Removed `-d` to see what was going on). The instance freezes up after a while.
> 
> As for running `tox`, I get weird errors
> 
> ```
> configuration error: `project` must contain ['name'] properties
> /home/dblanchette/repos/jira/.tox/py310/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py:125: _InvalidFile: The given `pyproject.toml` file is invalid and would be ignored.
>     !!
> 
> 
>     ############################
>     # Invalid `pyproject.toml` #
>     ############################
> 
>     Any configurations in `pyproject.toml` will be ignored.
>     Please note that future releases of setuptools will halt the build process
>     if an invalid file is given.
> 
>     To prevent setuptools from considering `pyproject.toml` please
>     DO NOT include the `[project]` or `[tool.setuptools]` tables in your file.
>     
> 
> !!
>     
>   if _skip_bad_config(project_table, orig_setuptools_table, dist)
> ``` 
> 
> Also:
> ```
> jira 3.3.3.dev3+g81d5127.d20220808 requires defusedxml, which is not installed.
> jira 3.3.3.dev3+g81d5127.d20220808 requires packaging, which is not installed.
> jira 3.3.3.dev3+g81d5127.d20220808 requires requests, which is not installed.
> jira 3.3.3.dev3+g81d5127.d20220808 requires requests-oauthlib, which is not installed.
> jira 3.3.3.dev3+g81d5127.d20220808 requires requests-toolbelt, which is not installed.
> jira 3.3.3.dev3+g81d5127.d20220808 requires typing-extensions, which is not installed.
> ```
> (I thought tox would install requirements by itself)
> 
> And then:
> ```
> ERROR: InvocationError for command /home/dblanchette/repos/jira/.tox/py310/bin/python -m pip check (exited with code 1)
> On branch fix/create-issue-issue-type-exists-several-projects
> Your branch is up to date with 'coveord/fix/create-issue-issue-type-exists-several-projects'.
> 
> Changes not staged for commit:
>   (use "git add <file>..." to update what will be committed)
>   (use "git restore <file>..." to discard changes in working directory)
>         modified:   tox.ini
> 
> no changes added to commit (use "git add" and/or "git commit -a")
> ERROR: Git reported dirty status. Git should never report dirty status at the end of testing, regardless if status is passed, failed or aborted.
> ```
> This seems weird to me as it is normal that I have modified files since I am working on the code. This happens even if I run tox after checking out the repo without changing anything else.
> 
> I was able to run the tests in the following manner:
> - `sudo apt install libkrb5-dev` (I was getting an error with krb5-config otherwise)
> - Commented out `git clean -xdf jira tests` and `python -m pip check` in tox.ini, under testenv.commands
> - Added `python -m pip install -r constraints.txt` after the lines I commented out above
> 
> After that, I was able to run the Jira instance container, wait for it to boot and then `tox -e py39 -- tests/resources/test_issue.py -k test_create_issue_with_issue_type_name`

dblanchette:
I don't know why CI is failing, it does not seem related to my changes

adehad (approved):
> I don't know why CI is failing, it does not seem related to my changes

yup don't worry about that, just a problem with github itself, we can't share secrets to pull requests so that fails

Test case looks good, ideally we would want a case where we have two issue types for different projects, but we don't have python functions for that yet, so just some technical debt we have for now

Thanks for sharing you troubles with the devcontainer, I really need to get to writing a page on using it, can I confirm if those dependencies needed to be installed inside the container?

dblanchette:
Thanks for merging!

> Thanks for sharing you troubles with the devcontainer, I really need to get to writing a page on using it, can I confirm if those dependencies needed to be installed inside the container?

Are we talking about the same thing? I just ran the container `jira addono/jira-software-standalone` manually on the side. The dependencies I installed were for the Python environment, tox did not seem to want to cooperate.

adehad:
Sorry my bad, I was thinking of: https://github.com/pycontribs/jira/blob/main/.devcontainer/Dockerfile

Issue #879, opened 2019-11-29
JIRA.my_permissions() needs a new required `permissions` parameter


cz-kialo:
**Describe the bug**
/mypermissions endpoint requires list of permissions in latest API version, see https://developer.atlassian.com/cloud/jira/platform/change-notice-get-my-permissions-requires-permissions-query-parameter/

**To Reproduce**

Call the `JIRA.my_permissions()` method

or

```
curl -D- -u $JIRA_USER:$JIRA_PWD -X GET https://$JIRA_DOMAIN.atlassian.net/rest/api/2/mypermissions
```

No Error for (if permission names exist)

    curl -D- -u $JIRA_USER:$JIRA_PWD -X GET https://$JIRA_DOMAIN.atlassian.net/rest/api/2/mypermissions?permissions=ADMINISTER,ADMINISTER_PROJECTS

Possible fix:

      def my_permissions(
        self, projectKey=None, projectId=None, issueKey=None, issueId=None
    ):

should get a new required parameter "permissions".

**Expected behavior**

Return permissions data

**Stack Trace**
```
  File "/Users/xxxxx/.virtualenvs/XXX/lib/python3.7/site-packages/jira/client.py", line 1854, in my_permissions
    return self._get_json('mypermissions', params=params)
  File "/Users/xxxxx/.virtualenvs/XXX/lib/python3.7/site-packages/jira/client.py", line 2549, in _get_json
    r = self._session.get(url, params=params)
  File "/Users/xxxxx/.virtualenvs/XXX/lib/python3.7/site-packages/jira/resilientsession.py", line 151, in get
    return self.__verb('GET', url, **kwargs)
  File "/Users/xxxxx/.virtualenvs/XXX/lib/python3.7/site-packages/jira/resilientsession.py", line 147, in __verb
    raise_on_error(response, verb=verb, **kwargs)
  File "/Users/xxxxx/.virtualenvs/XXX/lib/python3.7/site-packages/jira/resilientsession.py", line 57, in raise_on_error
    r.status_code, error, r.url, request=request, response=r, **kwargs)
jira.exceptions.JIRAError: JiraError HTTP 400 url: https://xxxxx.atlassian.net/rest/api/2/mypermissions
    text: The 'permissions' query parameter is required.
    
[...]

PR #1472, opened 2022-08-23
add required parameter field 'permissions' to my_permissions method

rena2damas:
As per Jira announcement:

_In order to guarantee stability and performance of Jira Cloud, we are constantly reviewing Jira REST APIs to make sure their design enables and encourages good practices. Starting 1 February 2019, we will require that you provide the permissions query parameter in all requests to the [Get my permissions](https://developer.atlassian.com/cloud/jira/platform/rest/#api-rest-api-2-mypermissions-get) resource._

Reference:
https://developer.atlassian.com/cloud/jira/platform/change-notice-get-my-permissions-requires-permissions-query-parameter/

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2630,6 +2630,7 @@ class JIRA:
         projectId: Optional[str] = None,
         issueKey: Optional[str] = None,
         issueId: Optional[str] = None,
+        permissions: Optional[str] = None,
     ) -> Dict[str, Dict[str, Dict[str, str]]]:
         """Get a dict of all available permissions on the server.
 
@@ -2638,6 +2639,7 @@ class JIRA:
             projectId (Optional[str]): limit returned permissions to the specified project
             issueKey (Optional[str]): limit returned permissions to the specified issue
             issueId (Optional[str]): limit returned permissions to the specified issue
+            permissions (Optional[str]): limit returned permissions to the specified csv permission keys (cloud required field)
 
         Returns:
             Dict[str, Dict[str, Dict[str, str]]]
@@ -2651,6 +2653,9 @@ class JIRA:
             params["issueKey"] = issueKey
         if issueId is not None:
             params["issueId"] = issueId
+        if permissions is not None:
+            params["permissions"] = permissions
+
         return self._get_json("mypermissions", params=params)
 
     # Priorities
--- a/jira/client.py
+++ b/jira/client.py
@@ -2634,6 +2634,10 @@ class JIRA:
     ) -> Dict[str, Dict[str, Dict[str, str]]]:
         """Get a dict of all available permissions on the server.
 
+        ``permissions`` is a comma-separated value list of permission keys that is
+        required in Jira Cloud. For possible and allowable permission values, see
+        https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-permission-schemes/#built-in-permissions
+
         Args:
             projectKey (Optional[str]): limit returned permissions to the specified project
             projectId (Optional[str]): limit returned permissions to the specified project
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -154,9 +154,11 @@ class FieldsTests(JiraTestCase):
         self.assertGreater(len(fields), 10)
 
 
-class MyPermissionsTests(JiraTestCase):
+class MyPermissionsServerTests(JiraTestCase):
     def setUp(self):
-        JiraTestCase.setUp(self)
+        super().setUp()
+        if self.jira._is_cloud:
+            self.skipTest("server only test class")
         self.issue_1 = self.test_manager.project_b_issue1
 
     def test_my_permissions(self):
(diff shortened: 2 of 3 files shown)


rena2damas:
this PR solves #879

tests/tests.py
```diff
@@ -154,9 +154,11 @@ def test_fields(self):
         self.assertGreater(len(fields), 10)
 
 
-class MyPermissionsTests(JiraTestCase):
+class MyPermissionsServerTests(JiraTestCase):
     def setUp(self):
-        JiraTestCase.setUp(self)
+        super().setUp()
+        if self.jira._is_cloud:
+            self.skipTest("server only test class")
```
adehad, line 161:
```suggestion
```

Not required as we use a pytest mark to determine what runs on the cloud, see below.

This opt-in method hopefully makes the tests easier to understand, let me know if that is not the case !

tests/tests.py
```diff
@@ -178,6 +180,44 @@ def test_my_permissions_by_issue(self):
         self.assertGreaterEqual(len(perms["permissions"]), 10)
 
 
+class MyPermissionsCloudTests(JiraTestCase):
```
adehad, line 183:
```suggestion
@allow_on_cloud
class MyPermissionsCloudTests(JiraTestCase):
```

This is imported from `tests.conftest`

The `skipTest` is fine here, if it doesn't work on server

jira/client.py
```diff
@@ -2638,6 +2639,7 @@ def my_permissions(
             projectId (Optional[str]): limit returned permissions to the specified project
             issueKey (Optional[str]): limit returned permissions to the specified issue
             issueId (Optional[str]): limit returned permissions to the specified issue
+            permissions (Optional[str]): limit returned permissions to the specified csv permission keys (cloud required field)
```
adehad, line 2642: Is there a link that we can reference to help users find out what keys are available?

adehad (changes requested):
Awesome, really appreciate you taking this on

I've made a separate branch to confirm that the tests run on the cloud instance (github has terrible secrets sharing to pull requests sorry), but happy for you to make further changes here and then merge this PR

jira/client.py
adehad: we can also raise an exception, e.g. `ValueError` if the user does not pass this variable but is running a cloud instance

tests/tests.py
rena2damas: You're right, for readability let's add the ```allow_on_cloud``` mark. We still need the ```skip``` since the permission API in the Jira cloud has diverged from the Jira server.

tests/tests.py
rena2damas: Without the ```skipTest```, the test class will run for the server and I think it fails with 400 bcs the server does not know what to do what the new ```permissions``` parameter added for cloud.

jira/client.py
rena2damas: [This](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-permission-schemes/#api-rest-api-3-permissionscheme-get ) page documents collections of permissions. I noticed also that the admin user created in ```make_local_jira_user.py``` contains 73 permissions which is much more than the ones documented in that page. What would be your suggestion in terms of improving docstring to include some Jira doc reference?

rena2damas: > we can also raise an exception, e.g. `ValueError` if the user does not pass this variable but is running a cloud instance

At the moment, if the user does not provide ```permissions``` field on a query for cloud Jira, it will raise a ```JiraError``` (with message ```"The 'permissions' query parameter is required."```). Is ```ValueError``` better than ```JiraError```? I personally prefer a ```JiraError``` but better be consistent in case other required fields are raising ```ValueError``` when not specified.

adehad: Ah nice, I think that existing `JiraError` is fine as is. It would make sense for us to handle it if the message was confusing, but that sounds very clear. But we should perhaps add a test for that case, ie we do not pass the parameter and we get the exception raised.

In terms of how to reference the docs, I would just add the full url, it looks like that is the standard elsewhere in other docstrings.

tests/tests.py
adehad: Yep but for this `MyPermissionsServerTests` we don't need the skip (as the cloud tests will never run it anyway as it does not have the marker)

tests/tests.py
rena2damas: Ok some update on my comment, more than readability, ```allow_on_cloud``` **must** be present to ensure the test runs on the cloud :)

jira/client.py
adehad: For now I think that we want to cover the Cloud case with the extra documentation due to the API change.

I think specifically we want to hyperlink:
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-permission-schemes/#built-in-permissions

tests/tests.py
rena2damas: Yep you're right ! I forgot that only tests marked with ```allow_on_cloud``` will run on the cloud. In that case that check is redundant .

tests/tests.py
adehad: yep exactly

https://github.com/pycontribs/jira/blob/009ec9fa50e482df153e695d67b11690ed0b5eea/.github/workflows/jira_cloud_ci.yml#L59

jira/client.py
rena2damas: Gotcha ! I added commit [7f82941](https://github.com/pycontribs/jira/pull/1472/commits/7f82941c8539196ad45f3a9f4e153539e2e79da2) that hopefully improves it but feel free to improve on it.

jira/client.py
```diff
@@ -2634,6 +2634,10 @@ def my_permissions(
     ) -> Dict[str, Dict[str, Dict[str, str]]]:
         """Get a dict of all available permissions on the server.
 
+        ``permissions`` is a comma-separated value list of permission keys that is
+        required in Jira Cloud. For possible and allowable permission values, see
+        https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-permission-schemes/#built-in-permissions
+
```
adehad, line 2640: Excellent

adehad (approved):
Perfect, I'll get this running in #1473 to verify the tests are happy and then merge this PR

PR #1473, opened 2022-08-23
bugfix/mypermissions args

adehad:
This is a complement to #1472 to prove that the cloud tests are running successfully


studioj (approved):

adehad:
#1473 now merged

PR #1483, opened 2022-08-31
let Jira cloud assign a default permissionScheme if none is specified

rena2damas:
issue: when a permissionScheme is specified , Jira cloud raises a 500 error

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4327,7 +4327,7 @@ class JIRA:
 
         ps_list: List[Dict[str, Any]]
 
-        if permissionScheme is None:
+        if permissionScheme is None and not self._is_cloud:
             ps_list = self.permissionschemes()
             for sec in ps_list:
                 if sec["name"] == "Default Permission Scheme":
@@ -4430,12 +4430,13 @@ class JIRA:
             "assigneeType": "PROJECT_LEAD",
             "description": "",
             # "avatarId": 13946,
-            "permissionScheme": int(permissionScheme),
             "notificationScheme": notificationScheme,
             "url": url,
         }
         if issueSecurityScheme:
             payload["issueSecurityScheme"] = int(issueSecurityScheme)
+        if permissionScheme is not None:
+            payload["permissionScheme"] = int(permissionScheme)
         if projectCategory:
             payload["categoryId"] = int(projectCategory)
 


rena2damas:
this closes #1483

adehad:
hmm this doesn't seem right, the Jira Cloud docs do say this is a correct parameter:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-projects/#api-rest-api-2-project-post

adehad:
we should add a test for this in our cloud tests

rena2damas:
I found this issue mentioned as well by some user: 

https://community.atlassian.com/t5/Jira-questions/Using-Default-Permission-Scheme-with-Create-Project-API-causes/qaq-p/1862489

It has almost a year and the issue still persists.

rena2damas:
The documentation says it's a valid field but it still throws a 500. We either embrace the error or we remove the field if not specified, which is what I am proposing.

The current test manager is suppressing that and other errors so the test suite does not cover that scenario:

```python
...
            create_attempts = 6
            for _ in range(create_attempts):
                try:
                    if self.jira_admin.create_project(project_key, project_name):
                        break
                except JIRAError as e:
                    if "A project with that name already exists" not in str(e):
                        raise e
```

adehad:
I don't think the test manager is suppressing this error, it only ignores if the error message is relating to the project already existing

adehad:
Are you able to confirm the integer value the original code returns for your instance?

jira/client.py
```diff
@@ -4327,7 +4327,7 @@ def create_project(
 
         ps_list: List[Dict[str, Any]]
 
-        if permissionScheme is None:
+        if permissionScheme is None and not self._is_cloud:
```
adehad, line 4330:
```suggestion
        if permissionScheme is None and self._is_cloud:
            # Let Jira decide the default field due to issues in handling the field
            # https://community.atlassian.com/t5/Jira-questions/Using-Default-Permission-Scheme-with-Create-Project-API-causes/qaq-p/1862489
        elif permissionScheme is None:
```

If we do want this change I would suggest doing it this way, so we can better document the reasoning for handling it in this way.

rena2damas:
My simplest test is the following:

```python
>>> j = JIRA(...)
>>> j._session.post(j._get_url("project"), data={"key": "X", "name": "XYZ", "leadAccountId": "xxx", "projectTypeKey": "software"})
<Response [201]>
>>> j._session.post(j._get_url("project"), data={"key": "X", "name": "XYZ", "leadAccountId": "xxx", "projectTypeKey": "software", "permissionScheme": 0}})
jira.exceptions.JIRAError: JiraError HTTP 500 url: https://xxx.atlassian.net/rest/api/2/project
        text: Internal server error
        
        response headers = {...}
        response text = {"errorMessages":["Internal server error"],"errors":{}}
```

Having the field ```permissionScheme``` on raises a ```500```.

rena2damas:
> Are you able to confirm the integer value the original code returns for your instance?

I am running it against Jira Cloud Platform and the ```permissionScheme ``` is forced to 0 if I don't specify it - if that was the question asked.

rena2damas:
> I don't think the test manager is suppressing this error, it only ignores if the error message is relating to the project already existing

You're right , it doesn't look like it. I need more time to understand what is suppressing the 500.

rena2damas:
Ok so I ran the Jira Cloud CI on my local env and it fails:

```python
...
  File "/tmp/jira/tests/conftest.py", line 255, in _create_project
    if self.jira_admin.create_project(project_key, project_name):
```

which is the following line:

```python
...
if self.jira_admin.create_project(project_key, project_name):
```

It begs the question of how is this not failing in this project?

I ran the test suite against API v2:

```bash
https://xxx.atlassian.net/rest/api/2
```

rena2damas:
Yes full stack trace below:

```bash
ERROR    tests.conftest:conftest.py:63 JiraError HTTP 500 url: https://xxx.atlassian.net/rest/api/2/project
        text: Internal server error
        
        response headers = {...}
        response text = {"errorMessages":["Internal server error"],"errors":{}}
  File "/tmp/jira/tests/conftest.py", line 58, in setUp
    self.test_manager = JiraTestManager()
  File "/tmp/jira/tests/conftest.py", line 156, in __init__
    self.create_some_data()
  File "/tmp/jira/tests/conftest.py", line 287, in create_some_data
    self.project_a_id = self._create_project(self.project_a, self.project_a_name)
  File "/tmp/jira/tests/conftest.py", line 259, in _create_project
    raise e
  File "/tmp/jira/tests/conftest.py", line 255, in _create_project
    if self.jira_admin.create_project(project_key, project_name):
  File "/tmp/jira/jira/client.py", line 4449, in create_project
    r = self._session.post(url, data=json.dumps(payload))
  File "/tmp/jira/.tox/py39/lib/python3.9/site-packages/requests/sessions.py", line 635, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/tmp/jira/jira/resilientsession.py", line 223, in request
    elif raise_on_error(response, **processed_kwargs):
  File "/tmp/jira/jira/resilientsession.py", line 69, in raise_on_error
    raise JIRAError(
jira.exceptions.JIRAError: JiraError HTTP 500 url: https://xxx.atlassian.net/rest/api/2/project
        text: Internal server error
        
        response headers = {...}
        response text = {"errorMessages":["Internal server error"],"errors":{}}

```

rena2damas:
It fails with ```500``` for API ```3``` as well. 

The simplest of tests is this:

```python
jira_admin=JIRA(...)
jira_admin.create_project("X", "XYZ")
```

adehad:
Very strange, if you are able to get logs from the Jira instance itself that might help us further, but unfortunately I am not sure why we are seeing this behaviour.

rena2damas:
> Very strange, if you are able to get logs from the Jira instance itself that might help us further, but unfortunately I am not sure why we are seeing this behaviour.

Thing is, this is the Cloud . If it was a local server I could check the logs and we could compare server versions . But being a cloud instance, it should all be the same version but possibly isn't . I created my cloud environment less than a month ago. And bcs yours is older it could be linked to a different version. But I'm just guessing here since none of this is documented and I'm assuming your cloud environment does not raise 500.

From here I can create an (extra) Jira ticket to report the situation .

rena2damas:
Ok as opposed to what I said yesterday, it does seem the cloud has a version linked to it. My cloud instance has the following properties:

```python
{'baseUrl': 'https://xxx.atlassian.net',
 'buildDate': '2022-09-05T12:15:56.000+0000',
 'buildNumber': 100206,
 'defaultLocale': {'locale': 'en_US'},
 'deploymentType': 'Cloud',
 'scmInfo': '40f823cc9831469ea3c6decd06fddaf0ad67f8e2',
 'serverTime': '2022-09-06T08:03:48.153+0000',
 'serverTitle': 'Jira',
 'version': '1001.0.0-SNAPSHOT',
 'versionNumbers': [1001, 0, 0]}
```

Please confirm that you are using an older version than mine.

adehad:
Sorry @rena2damas , have been swamped with work, just letting you know I haven't forgotten to look into this further

rena2damas:
@adehad no problem I myself been also busy but thanks nonetheless

studioj (changes requested):
Thank you for your contribution, could you add a test to verify this?

PR #1498, opened 2022-09-08
Add method to get issues completed within a sprint

lukebarklimore:
I've added the ability to get a list of issues completed within a particular sprint. It works the same way `removed_issues()` does. I found `completedIssues` in the JSON which is used in the Sprint Report's 'Completed Issues' section.

While I was here, I fixed the misleading docstring which I stumbled across whilst trying to find the method to return the list of items I wanted (completed ones).

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4735,6 +4735,19 @@ class JIRA:
 
         return issues
 
+    def completed_issues(self, board_id: str, sprint_id: str):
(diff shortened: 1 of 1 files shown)


studioj (changes requested):
awesome @lukebarklimore , thank you for contributing. It would be cool if you could add a testcase which validates this behaviour. Sorry for taking so long for a review :-s

PR #1502, opened 2022-09-16
Use proper SPDX identifier

Shortfinga:


adehad (approved):
Thanks !

(link for my future reference: https://spdx.org/licenses/BSD-2-Clause.html)

PR #1518, opened 2022-10-07
bug: request_type_by_name requires limit parameter in larger instances

AlexanderNorris:
Due to default limit parameters it is possible that the `request_type_by_name()` function does not find request types successfully. By adding the ability to pass a limit this can be extended to allow users with more than the Default 100, or less dependent on settings in your organisation, requestTypes to work within this library.


studioj:
please add a test to validate this

AlexanderNorris:
Dumb question, can you please provide me a link to where the other tests for the `create_customer_request()` function are?

PR #1521, opened 2022-10-21
Modified the type hinting for the attachment in the add_attachment fu…

rpop0:
…nction to properly support BytesIo and StringIO

This change was done due to the fact that using an object of type BytesIO or StringIO resulted in IDE warnings regarding mismatched types.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -21,7 +21,7 @@ import warnings
 from collections import OrderedDict
 from collections.abc import Iterable
 from functools import lru_cache, wraps
-from io import BufferedReader
+from io import BufferedReader, BytesIO, StringIO
 from numbers import Number
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


studioj (approved):

studioj:
CI fails here
`  jira/client.py:1001: error: Incompatible types in assignment (expression has type "Union[BufferedReader, BytesIO, StringIO]", variable has type "BufferedReader")  [assignment]
`
please fix

studioj:
@rpop0 there were a lot of fixes to docstrings and types merged in another PR. Could you check if yours is still an issue?

rpop0:
@studioj Sorry for the late response but the incorrect type hinting still seems to be an issue in the latest version. I'll update this when I get the time.

cperrin88:
Wouldn't it be enough to just change the type to [`BufferedIOBase`](https://docs.python.org/3/library/io.html#io.BufferedIOBase)?

PR #1526, opened 2022-11-03
Properly parse errorMessage key in error response

GeyseR:
After [recent refactoring](https://github.com/pycontribs/jira/commit/c6d59a1b2fce536b9cb11139a38c474ce23ca416) of the resilientsession module, we started to get empty text field in the JIRAError instances for the "Site temporary unavailable" errors.

This PR adds proper processing of such error responses and also adds consistency in errors processing in several parts of the lib

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union,
 from requests import Response
 from requests.structures import CaseInsensitiveDict
 
-from jira.resilientsession import ResilientSession
+from jira.resilientsession import ResilientSession, parse_errors
 from jira.utils import json_loads, threaded_requests
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


jira/resources.py
```diff
@@ -358,7 +329,8 @@ def update(
         r = self._session.put(self.self + querystring, data=json.dumps(data))
         if "autofix" in self._options and r.status_code == 400:
             user = None
-            error_list = get_error_list(r)
+            raise_on_error(r)
+            error_list = parse_errors(r)
             logging.error(error_list)
```
adehad, line 334: would we want to log the error and afterwards raise?

adehad (changes requested):
thanks for catching this.

Could you add some tests for this? The `tests/test_resilientsession.py` would be the best place

jira/resources.py
GeyseR: it looks like raising the error should be removed completely as it will always raise for 400 HTTP response
previously it was raised in the `json_loads` and caught by `get_error_list` code

GeyseR:
I've fixed the issue with unexpected error raising and added a test for errors parsing logic

adehad (approved):
Excellent stuff

adehad:
Thanks for the great quality PR !

Issue #1462, opened 2022-08-16
Issue createmeta will fail from Jira 9+


Ry-DS:
### Problem trying to solve

If anyone tries to use this library to access a Jira server running v9+, the function will fail. Check out https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html for more details.

### Possible solution(s)

You can see an example solution here: https://github.com/atlassian-api/atlassian-python-api/pull/1031

### Alternatives

You will need to write your own api requests if you want to use the `createmeta` endpoint, as it will fail if you use the latest version of Jira.

adehad:
In addition to the deprecation warning can raise a sensible exception by checking the jira version if run on a Jira 9+

dshvedchenko:
@adehad please review #1503, I'm open to discussion and  implementation of all required suggestion

pmilosev:
I've opened and alternative pull request, please review: https://github.com/pycontribs/jira/pull/1527

PR #1527, opened 2022-11-08
Support for the new 'createmeta' API in Jira 8.4+

pmilosev:
Fix for https://github.com/pycontribs/jira/issues/1462

This is an alternative to https://github.com/pycontribs/jira/pull/1503, introducing new API functions and not trying to reconstruct the old API. Even when filtering is used in the old API to get metadata for single project and issue type, the results are still not identical (differ in structure).

I'm not sure if any tests can be written testing against different versions of JIRA. At the moment I have tested manually against JIRA 7.13.8, 8.22.6 and 9.3.1.


pmilosev:
Am I supposed to add a label to the PR or the admins ?
I see a failing check for it.

dshvedchenko:
you are right, results a different. i will work on make them similar.
but in your code you also need to pay attention that _get_json is not paging aware, and will return only first part of metadata.

this is why I'm using _fetch_pages , coz it gets all metadata

pmilosev:
Thanks for reminding me about the paging ... I remember writing a TODO and later not looking into it.
This should not be a problem for the new API though as it only returns results (the fields) for one issue type, so only if there are over 50 fields in a issue type the paging will be an issue.

I still want to discourage you from trying to make the old API behave like the new one:
- you need complicated code that iterates through all projects (all filtered projects) and all (filtered) issue types, composing the result in the same structure as earlier; but then you are going against the JIRA API decisions
- if you don't do that and only reconstruct the results from the first project and first issue type (I remeber seeing that in the PR) then you are implicitely introducing filters that the callers are not expecting; so not really a benefit, but a source for potential bugs
- currently Jira 9 is not supported (already breaks), so there are no depndant projects that work and will break if we deprecate the old API and provide a new one; Since Atlassian decided to change the architecture of their APIs I think the best is to explicitely follow those decisions

Anyway these are conceptual decisions that need to be made by the maintainers.

pmilosev:
It appears the original `createmeta` function is also using `_get_json` so I will not change this unless it causes an issue:
- The new API is filtered by design so if paging was not an issue before it will not be now
- This whole technology stack is new for me so it's more likely I will introduce bugs
- I have limited setup / capacity for testing, which combined with previous point makes me want to introduce as few changes as possible

jira/client.py
```diff
@@ -1750,6 +1792,15 @@ def createmeta(
             Dict[str, Any]
 
         """
+        if self._version >= (9, 0, 0):
+            raise JIRAError(f"Unsupported JIRA version: {self._version}")
```
studioj, line 1796: also add the info from below in the exception ... => please use createmeta_issuetypes and createmeta_fieldtypes

jira/client.py
```diff
@@ -1722,6 +1722,48 @@ def create_customer_request(
...
+        For details see: https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html
+
+        Args:
+            projectIdOrKey (Union[str, int]): id or key of the project for which to get the metadata.
+            issueTypeId (Union[str, int]): id of the issue type for which to get the metadata.
+        Returns:
+            Dict[str, Any]
+        """
+        if self._version < (8, 4, 0):
+            raise JIRAError(f"Unsupported JIRA version: {self._version}")
```
studioj, line 1761: also add the info from below in the exception ... => please use createmeta_issuetypes and createmeta_fieldtypes

jira/client.py
```diff
@@ -1722,6 +1722,48 @@ def create_customer_request(
...
+        This API was introduced in JIRA 8.4 as a replacement for the more general purpose API 'createmeta'.
+        For details see: https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html
+
+        Args:
+            projectIdOrKey (Union[str, int]): id or key of the project for which to get the metadata.
+        Returns:
+            Dict[str, Any]
+        """
+        if self._version < (8, 4, 0):
+            raise JIRAError(f"Unsupported JIRA version: {self._version}")
```
studioj, line 1740: also add the info from below in the exception ... => please use createmeta_issuetypes and createmeta_fieldtypes

studioj (changes requested):
it would be nice if you would help the user more in the exception message

adehad:
> I'm not sure if any tests can be written testing against different versions of JIRA. At the moment I have tested manually against JIRA 7.13.8, 8.22.6 and 9.3.1.

The docker image allows specifying a version so we should be able to configure a test case to run on  the higher version, example:
https://github.com/pycontribs/jira/blob/1c36bcdbd3ebdbebdb089f411d463e4ea2840d44/.github/workflows/jira_server_ci.yml#L20

jira/client.py
```diff
@@ -1750,6 +1792,15 @@ def createmeta(
...
 
         """
+        if self._version >= (9, 0, 0):
+            raise JIRAError(f"Unsupported JIRA version: {self._version}")
+        elif self._version >= (8, 4, 0):
+            warnings.warn(
+                "This API have been deprecated in JIRA 8.4 and is removed in JIRA 9.0."
+                "Use 'createmeta_issuetypes' and 'createmeta_fieldtypes' instead.",
+                DeprecationWarning,
+            )
```
adehad, line 1802: we probably want to change the stack level here. Unfortunately the python default isn't super helpful. 
We can check this is a mocked test

pmilosev: Sorry, I don't have a lot of experience with Python, so I don't understand what does _stack level_ in this context mean.
Should I do something like:
```python
 warn(
            "This function will fail from Jira 9+. "
            "Use issue_createmeta_issuetypes or issue_createmeta_fieldtypes instead.",
            DeprecationWarning,
            stacklevel=2,
        )
```
(taken from https://github.com/atlassian-api/atlassian-python-api/pull/1031/files)

pmilosev:
I just discovered that cloud JIRA API is independent from Jira DC/Server API and any clients using it must account for the potential differences: https://community.atlassian.com/t5/Jira-questions/How-can-I-tell-actual-Jira-version-of-a-cloud-instance-for/qaq-p/1483730

According to the cloud Jira API documentation the old `createmeta` endpoint is still used, so I'll have to account for this in my PR.

jira/client.py
adehad: yep exactly right, here's a useful video explaining this situation: https://www.youtube.com/watch?v=CtFdXBEwYfk

adehad:
> I just discovered that cloud JIRA API is independent from Jira DC/Server API and any clients using it must account for the potential differences: https://community.atlassian.com/t5/Jira-questions/How-can-I-tell-actual-Jira-version-of-a-cloud-instance-for/qaq-p/1483730
> 
> According to the cloud Jira API documentation the old `createmeta` endpoint is still used, so I'll have to account for this in my PR.

@pmilosev yes it can get quite tricky.
We have `JIRA._is_cloud` that can be used to check whether we are on server or cloud too.

We've compiled a list of the API docs here in case you haven't found it already:
https://github.com/pycontribs/jira#jira-rest-api-reference-links

pmilosev:
I've updated the code according to the review comments above and accounted for the difference in Jira Cloud API.
Manual re-test against Jira DC 7.13.8, 8.22.6 and 9.3.1, as well as Jira Cloud 1001 (build number 100210).

pmilosev:
@adehad is there something else blocking this PR from being merged ?
I would appreciate if we can get this done while I still have some of my attention to it.

adehad:
Thanks for the reminder @pmilosev .

Could I ask you to see if we can get some test cases in?

For testing the versions we don't have access to at the moment I would suggest mocking the value of `self._cloud` and `self._version`.

Let me know if you need a hand with this style of test. Off the top of my head I believe you would want to patch the the test method, as that is probably the best way to patch them before the `setUp()` runs. Or otherwise feel free to use a pytest style test.

```python
@mock.patch(...)
def test_createmeta_test_name():

```

Roooodie:
any update on this? issues still happened if we're using Jira 9+.
sorry if I'm bothering people here, I am just curious about the failed checking changes above.
@studioj @pmilosev

frost9i:
Can checks be forced to restart?
It looks like the issue was fixed, but wasn't merged because automated testing failed.

studioj (approved):

studioj:
> Can checks be forced to restart? It looks like the issue was fixed, but wasn't merged because automated testing failed.

the server CI part is a bit flakey and more failing than passing :s 
@adehad can you merge?

pmilosev:
I wanted to add some tests but am too busy to do so, sorry. I think tests are indeed a good thing and if anyone is able to write them I would really appreciate it. For me the issue is that I'm not experienced with the stack and writing unit tests would require me to spend a decent amount of time learning.

That being said the failing checks predate my changes.
I have tested the change manually against multiple versions of Jira (see previous comments for details) and the change is already used in another product (see DefectDojo links above).

I hope the PR can be merged as it is.

adehad:
Thank you for your contributions @pmilosev and @dshvedchenko. 

I've merged it as is for now and raised a separate issue for adding tests. 

Hopefully can get it into a release soon

PR #1528, opened 2022-11-11
Add overload to JIRA.search_issues() for better return type hinting

seanboothm:


jira/client.py
```diff
@@ -2963,6 +2963,32 @@ def resolution(self, id: str) -> Resolution:
...
+    ) -> ResultList[Issue]:
+        ...
+
+    @overload
+    def search_issues(
+        self,
+        jql_str: str,
+        startAt: int,
+        maxResults: int,
+        validate_query: bool,
```
adehad, line 2985: Let me know if you figure out how to do this so mypy is happy.

I believe it was this issue I was running into when I first started adding typehints for these chunkier functions
https://github.com/python/mypy/issues/6580

PR #1551, opened 2022-11-16
Updating installation instructions

denilsonsa:
`pip -E` was removed in pip 1.1 (2012-02-16). https://pip.pypa.io/en/stable/news/#v1-1

Since the previous paragraph mentions running the client standalone, let's install `jira[cli]` instead of the plain `jira`.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -7,13 +7,20 @@ The easiest (and best) way to install jira-python is through `pip <https://pip.p
 
 This will handle installation of the client itself as well as the requirements.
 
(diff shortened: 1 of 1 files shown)


denilsonsa:
One CI job is unhappy because this PR doesn't have any labels. And I don't think I can add labels myself.

The other CI job(s) are failing due to Python tests failing, which is completely unrelated to this PR that only changes a couple of lines of documentation.

adehad (approved):

adehad:
Thanks for the contribution!

Issue #1452, opened 2022-08-12
parameter template_name is ignored in create_project method


KLaMaCH:
### Bug summary

The HTTP request payload use the template_key variable which is not a parameter:
https://github.com/pycontribs/jira/blob/fdebf2e1ff006413c1ef2ee73d7252ccf4d9a4c9/jira/client.py#L4384

The template_key variable will be defined only if template_name is none:
https://github.com/pycontribs/jira/blob/fdebf2e1ff006413c1ef2ee73d7252ccf4d9a4c9/jira/client.py#L4317-L4323

Thus if I call create_project() method with any template_name the project will be created without a template.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

3.3.2

### Python Interpreter version

3.10

### Which operating systems have you used?

- [X] Linux
- [x] macOS
- [X] Windows

### Reproduction steps

```python
jira.create_project(
            'PK',
            template_name='com.pyxis.greenhopper.jira:gh-simplified-scrum-classic'
)
```

### Stack trace

```python
payload = {'name': 'PK', 'key': 'PK', 'projectTypeKey': 'software', 'projectTemplateKey': None, 'leadAccountId': '5ad724c3e378d62b3bb41132', 'assigneeType': 'PROJECT_LEAD', 'description': '', 'permissionScheme': 0, 'notificationScheme': 10000, 'url': '', 'issueSecurityScheme': 10000}
```

### Expected behaviour

Project created with the particular template

Davisned:
@adehad Neither your draft, nor @geraudloups solution have been pulled into master... any reason for that?

adehad:
I've been meaning to add a test, as this would be important in making sure we don't break the functionality again. Unfortunately my availability for open source isn't very high at the moment, am happy to accept contributions that expand upon those solutions

PR #1553, opened 2022-11-19
use `template_name` for `create_project()`

adehad:
Looks like @geraudloup fixed this

#1452

## TODO:
1. Add a test


sebasmiranda1:
This is a very important fix, I was going to suggest it myself but checked the PR's.

Right now the create_project function creates only one type of project, it's not using the template_name provided by the user.

PR #1574, opened 2022-12-07
[feature] Add support for requests timeout

Victor-D:
This PR adds the timeout parameter when performing HTTP request and fix some comments/typing related to the timeout parameter. 

Jira is based on the python requests library to perform HTTP requests so we can use its timeout parameter as explained here : https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

I think this library planed to do so, but it was not implemented.

The timeout parameter still has a None default value so no change is expected on applications which don't set a timeout value. For other applications who though they were using it then it will start to work.


bboe (approved):
I look forward to seeing this PR merged since the functionality was accidentally lost here:
https://github.com/pycontribs/jira/pull/1366/files#diff-0605333df4b31ca7f97b2591e3bcba9c71ea201f4a88e383316e7f6c1c478e06L177

studioj (dismissed):

studioj:
Looks cool and thank you alot for adding a few tests. The only issue I foresee here @Victor-D is that this "bug-fix" might break some systems which have defined a super low timeout which unwittingly worked.
what's your opinion @adehad

adehad:
I didn't realise this was a regression, we should merge this in that case. We should be able to debug any issues for timeout errors if they come up in issues or discussions so let's try it for now

PR #1578, opened 2022-12-18
Upgrade to tox v4

ssbarnea:


ssbarnea:
@adehad ping me on `#pycontribs` channel on matrix/irc, I want to chat a little bit about use of GHA in jira package.

adehad:
Summary of discussion:

We should avoid using `lint` and `docs` in the same github action as the tests. Likely we can simply remove, as we are using pre-commit and the readthedocs actions separately anyway.

Discussed using, which we will look into in the coming days:
https://github.com/coactions/dynamic-matrix

PR #1580, opened 2022-12-28
Update the type for issue in add_comment & transition_issue for better type hints

dvaerum:
I have found that these 2 function works with both the type "str" and "ISSUE" because whatever one adds as the issue argument is converted to a string, and when the class Issue is converted to a string it returns the issue-key.
So, this should give better type hinting, and I am sure there are more place in the code this scenario exists 😄

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1936,7 +1936,7 @@ class JIRA:
     @translate_resource_args
     def add_comment(
         self,
-        issue: str | int,
+        issue: str | int | Issue,
         body: str,
         visibility: dict[str, str] | None = None,
         is_internal: bool = False,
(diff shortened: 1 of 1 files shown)


adehad:
Thanks @dvaerum, could you rebase and see if this fixes the any issues. 

Also a reminder to update the docstrings !

dvaerum:
> Thanks @dvaerum, could you rebase and see if this fixes the any issues.
> 
> Also a reminder to update the docstrings !

Done, but I am not certain why ci/cloud/py3.8 fails

adehad (approved):
Yep don't worry about the cloud tests, there are some github limitations that prevent pull requests accessing the required env vars.

Thanks for sorting these out!

PR #1586, opened 2023-01-08
adding sprint goal

mohit2152sharma:
The method `update_sprint` doesn't allow to update the sprint goal, as it is not included in method's parameter. Simply adding `goal` into function's argument, just like other arguments (e.g. `name`, `endDate`, `startDate`) fixes that.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4698,7 +4698,9 @@ class JIRA:
                 )
         return sprints
 
-    def update_sprint(self, id, name=None, startDate=None, endDate=None, state=None, goal=None):
+    def update_sprint(
+        self, id, name=None, startDate=None, endDate=None, state=None, goal=None
+    ):
         payload = {}
         if name:
             payload["name"] = name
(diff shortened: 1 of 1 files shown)


studioj:
awesome @mohit2152sharma  could you also add a test which verifies that functionality

studioj (changes requested):
Thanks for submitting this.
please add a testcase validating this change

mohit2152sharma:
I couldn't find test cases for `update_sprint()` function, I was thinking to modify that to include this scenario, could you direct me to the respective section if it exists?

studioj:
> I couldn't find test cases for `update_sprint()` function, I was thinking to modify that to include this scenario, could you direct me to the respective section if it exists?

maybe have a look here
https://github.com/pycontribs/jira/blob/main/tests/resources/test_sprint.py

adehad:
Thanks for starting this PR @mohit2152sharma ! We have now merged in a PR that took inspiration from your work #1806

PR #1596, opened 2023-01-28
Add union issue

Sineaggi:
There are other functions in the `client.py` file that take issue typed as a `str` that then call `str(issue)`.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2069,7 +2069,9 @@ class JIRA:
 
     # non-resource
     @translate_resource_args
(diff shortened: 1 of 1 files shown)


studioj:
could it be this is solved by https://github.com/pycontribs/jira/pull/1396 ?
please review and approve

Sineaggi:
> could it be this is solved by #1396 ? please review and approve

Looks like that solves the issues I was running into and more.

studioj:
@adehad care to merge the https://github.com/pycontribs/jira/pull/1396. I'll close this one as its duplicate. Thx @Sineaggi for contributing ❤

PR #1626, opened 2023-03-14
Provide page-aware access to createmeta issuetypes

ddelabru:
Version 3.5.0 of the client library introduced the createmeta_issuetypes() and createmeta_fieldtypes() client member functions to replace the deprecated form of the createmeta Jira endpoint. However, these functions return the raw JSON of a single response, and do not handle pagination that may be applied to the endpoints, such as when an issue type within a project has more than 50 associated fields. I recently encountered a Jira deployment where this case occurred, rendering
createmeta_fieldtypes() unuseful.

Because the functions added in 3.5.0 have a different return type than these new functions, instead of changing the behavior of those functions this commit creates two new client member functions: project_issue_types()  and project_issue_fields().

I opened this commit after I had to query project issue type fields in a case where there were more than 50.

I am open to constructive criticism if anyone can think of a more elegant solution than creating a new `Resource` subclass for something that cannot be individually queried in the Jira API.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -60,6 +60,7 @@ from jira.resources import (
     Customer,
     CustomFieldOption,
     Dashboard,
+    Field,
     Filter,
     Group,
     Issue,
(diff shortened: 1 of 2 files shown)


ddelabru:
I can see the coverage check failed. I will try to add unit tests 15 March.

studioj:
Thank you for the PR.
would be great if next to the tests you would be adding, you could also extend the function docstrings with some extra explanation while you're at it... we seem to be lacking there :)

adehad:
Happy to provide any support you may need for this, looking good so far!

SharryXu:
Hi @ddelabru ,
Thanks for your PR 👍 
I think we should use your version instead of https://github.com/pycontribs/jira/pull/1729.

adehad:
Thanks for this great quality PR @ddelabru, I've merged it in with another contributor's tests and should get it merged in soon.

Issue #1573, opened 2022-12-06
 SSL Certificate Not Sent


Minweiwangaaaa:
### Bug summary

error
```
  File "/Library/Python/3.9/site-packages/jira/resilientsession.py", line 69, in raise_on_error
    raise JIRAError(
jira.exceptions.JIRAError: JiraError HTTP 400 url: https://jira.xxxx.com/rest/api/2/serverInfo
	text: <html>
<head>
<title>400 Required SSL Certificate Not Sent</title>

```

My code:

 ```
   passaa = "xxxxxx=="
    uPass = base64.b64decode(passaa).decode('UTF-8')

    jira_options = {
        'server': 'https://jira.xxxxxx.com/',
        'verify': './akamai.cer',
        'basic_auth':('myname', uPass)
    }

    jira = JIRA(jira_options)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

Atlassian Jira Project Management Software (v8.8.1#808001-sha1:e5cdcf9)

### jira-python version

3.4.1

### Python Interpreter version

3.8

### Which operating systems have you used?

- [ ] Linux
- [X] macOS
- [ ] Windows

### Reproduction steps

```python
My code:

 
   passaa = "xxxxxx=="
    uPass = base64.b64decode(passaa).decode('UTF-8')

    jira_options = {
        'server': 'https://jira.xxxxxx.com/',
        'verify': './akamai.cer',
        'basic_auth':('myname', uPass)
    }

    jira = JIRA(jira_options)
```
```


### Stack trace

```python
Traceback (most recent call last):
  File "/Users/xxxx/PycharmProjects/pythonProject/fortify.py", line 378, in <module>
    postToJira('aaaa')
  File "/Users/xxxxx/PycharmProjects/pythonProject/fortify.py", line 371, in postToJira
    jira = JIRA(jira_options)
  File "/Library/Python/3.9/site-packages/jira/client.py", line 574, in __init__
    si = self.server_info()
  File "/Library/Python/3.9/site-packages/jira/client.py", line 3059, in server_info
    j = self._get_json("serverInfo")
  File "/Library/Python/3.9/site-packages/jira/client.py", line 3764, in _get_json
    r = self._session.get(url, params=params)
  File "/Library/Python/3.9/site-packages/requests/sessions.py", line 600, in get
    return self.request("GET", url, **kwargs)
[...]

PR #1631, opened 2023-03-27
fix verify not sent with request (#1573)

jongracecox:
This commit fixes an issue where the `verify` option is not passed along with requests issued by the ResilientSession class. This commit adds the `verify` option into the request args inside the `_jira_prepare` function.

The commit includes a unit test that verifies that `session.verify` is correclty handed over to requests during subsequent Jira API calls.

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -184,6 +184,9 @@ class ResilientSession(Session):
             # but for people subclassing we should preserve old behaviour
             prepared_kwargs["data"] = json.dumps(data)
 
+        if "verify" not in prepared_kwargs:
+            prepared_kwargs["verify"] = self.verify
+
         return prepared_kwargs
 
     def request(  # type: ignore[override] # An intentionally different override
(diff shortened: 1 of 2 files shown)


studioj:
thanks for the PR

studioj (dismissed):

jongracecox:
It's not clear to me why the `ci / cloud / py3.8` tests are failing, but I notice the same errors on other PRs (#1528), so wonder whether this is not specific to this PR.

jongracecox:
> It's not clear to me why the `ci / cloud / py3.8` tests are failing, but I notice the same errors on other PRs (#1528), so wonder whether this is not specific to this PR.

I assume it because my PR doesn't have access to the project secrets, so doesn't need anything fixed from my side.

adehad:
> > It's not clear to me why the `ci / cloud / py3.8` tests are failing, but I notice the same errors on other PRs (#1528), so wonder whether this is not specific to this PR.
> 
> I assume it because my PR doesn't have access to the project secrets, so doesn't need anything fixed from my side.

Exactly the case, sorry for the delay. Thanks for the great quality PR!

adehad (approved):

PR #1643, opened 2023-04-02
Switch from SIGNATURE_RSA to SIGNATURE_HMAC_SHA1

traylenator:
RHEL 9 has a policy of deprecating SHA1 and has removed it from python-oauthlib via the addition of locally added patch.

Switching to SIGNATURE_RSA_HMAC_SHA1 allows this module to be used on RHEL9 when the system package python3-oauthlib-3.1.1-2.el9.noarch.rpm is used.

Here is the patch form RHEL9 python-oauthlib.
[0001-Rip-out-RSA-SHA1.patch.txt](https://github.com/pycontribs/jira/files/11132574/0001-Rip-out-RSA-SHA1.patch.txt)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3702,13 +3702,13 @@ class JIRA:
     def _create_oauth_session(
         self, oauth, timeout: float | int | tuple[float, float] | None
     ):
-        from oauthlib.oauth1 import SIGNATURE_RSA
+        from oauthlib.oauth1 import SIGNATURE_HMAC_SHA1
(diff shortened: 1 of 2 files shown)


adehad (approved):
Awesome, appreciate it!

xdgc:
This breaks existing software. Against Jira Server `v8.20.10#820010-sha1:ace47f9` and using python-jira 3.5.1, I get `oauth_problem=signature_method_rejected`. Pinning to 3.5.0, code still works.

PR #1648, opened 2023-04-19
[enhancement] Add get_form to allow access to form data in jira issues (proform api)

damnmso:
unfortunately i have no docker running and limited in resources.
I create this PR now and try to get the docker instances running at my home workstation to test if i can get it working on my private machine (more resources)

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4384,16 +4384,23 @@ class JIRA:
     def get_form(self, issue: str, formid: str) -> list[dict]:
         """For the specified issue, returns the form data as key/value list. (Admin required).
 
-            Args:
-                issue (str): The issue id.
-                formid (str): the ID of the form, default is 1
+        Args:
+            issue (str): The issue id.
+            formid (str): the ID of the form, default is 1
 
-            Returns:
-                List[dict]: A dictionary containing the "answers" as value and the label as key.
+        Returns:
+            List[dict]: A dictionary containing the "answers" as value and the label as key.
         """
-        response = self._session.get(self.server_url + "/rest/proforma/1/issue/" + issue + "/form/" + formid + "/answers")
+        response = self._session.get(
+            self.server_url
+            + "/rest/proforma/1/issue/"
+            + issue
+            + "/form/"
+            + formid
+            + "/answers"
+        )
         return json_loads(response)
-    
+
     def get_issue_type_scheme_associations(self, id: str) -> list[Project]:
         """For the specified issue type scheme, returns all of the associated projects. (Admin required).
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -4381,6 +4381,19 @@ class JIRA:
         self.permissionschemes.cache_clear()
         return data
 
+    def get_form(self, issue: str, formid: str) -> list[dict]:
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -4381,6 +4381,26 @@ def delete_permissionscheme(self, id: str):
...
+        """For the specified issue, returns the form data as key/value list. (Admin required).
+
+        Args:
+            issue (str): The issue id.
+            formid (str): the ID of the form, default is 1
+
+        Returns:
+            List[dict]: A dictionary containing the "answers" as value and the label as key.
+        """
+        response = self._session.get(
```
adehad, line 4394: have a look at thee function below and try and copy the style.
We want to use the `_get_url` function with f-strings based on the user input

jira/client.py
```diff
@@ -4381,6 +4381,26 @@ def delete_permissionscheme(self, id: str):
         self.permissionschemes.cache_clear()
         return data
 
+    def get_form(self, issue: str, formid: str) -> list[dict]:
```
adehad, line 4384:
```suggestion
    def get_form(self, issue: str, form: str) -> list[dict]:
```

and change docstring.

jira/client.py
```diff
@@ -4381,6 +4381,26 @@ def delete_permissionscheme(self, id: str):
...
+        """For the specified issue, returns the form data as key/value list. (Admin required).
+
+        Args:
+            issue (str): The issue id.
+            formid (str): the ID of the form, default is 1
```
adehad, line 4389: how do we enforce this default? It doesn't seem to be set in the function signature.

adehad (changes requested):
good start, but needs a bit of attention

jira/client.py
damnmso: why enforce? maybe it should instead be renamed to "starts with 1". i am no fan of implicit magic that happens somehow. if someone wants the form data, he should know the ID. if he has multiple forms, then it might be 2, 3 or something else.

jira/client.py
damnmso: its an ID, not a form object, so i would not change that.

jira/client.py
adehad: That would be better yes, if we are saying that form IDs at 1 indexed that is different from saying the "default" is 1, as that has specific connotations in function documentations.

jira/client.py
adehad: My preference would be to use the generic term `form`, and use typehints to specify the types we allow, we then can handle a Form object in the future by simply extending the typehints, rather than adding a new argument to the function.

For example here, we might want to allow `str | int` and then in the function cast it to the string

GeotabDDag:
Hello, is there any expected release timeline for this get forms method? I would like to leverage this functionality.

robert-at-accenture:
Hi, I would also really appreciate if you would release this feature 😊

GvandeSteeg:
Hi, yes this would be a great feature if it could be completed @damnmso @adehad 
For the time being I've simply implemented the request manually, but it would be a fantastic feature if implemented accordingly

PR #1664, opened 2023-06-01
Allow the user to override the default signature method

Governa:
PR #1643 changed the default signature_method from SIGNATURE_RSA to SIGNATURE_HMAC_SHA1. This was done to keep compatibility with RHEL.

The problem is that some Jira servers don't accept SIGNATURE_HMAC_SHA1 and there's currently no way for the user to choose which method to use.

This PR adds a new, optional, field that can be passed to the oauth parameter and allows the user to choose which signature_method they want.


adehad (approved):
Thanks for the contribution!

Issue #1629, opened 2023-03-27
create_filter() to support permissions


sathishsms:
### Bug summary

Pls add support for this https://docs.atlassian.com/software/jira/docs/api/REST/9.3.1/#api/2/filter-addSharePermission

```
permissions = [
        {
            "type": "group",
            "groupname": "Engineering",
            "view": True,
            "edit": True,

response = requests.post(
            f"{url}/rest/api/2/filter/{filter_id}/permission",
            auth=auth,
            headers=headers,
            json=permission,
        )

        # Check the status code of the response
        if response.status_code == 201:
            current_app.logger.info(
                f"Filter permission updated successfully!
                  Status code: {response.status_code}, text: {response.text}"
            )
        else:
            current_app.logger.error(
                f"Error updating filter permission. Status
                  code: {response.status_code}, text: {response.text}"
            )
```

Need support for adding permission for the filters being created

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### Jira instance version

8.16.1

### jira-python version

main

### Python Interpreter version

3.10

### Which operating systems have you used?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
response = requests.post(
            f"{url}/rest/api/2/filter/{filter_id}/permission",
            auth=auth,
            headers=headers,
            json=permission,
        )
```

### Stack trace

```python
NA
```

### Expected behaviour

NA

PR #1679, opened 2023-06-07
Add 'add_filter_share_permissions' method and support in create_filter()

ABizzinotto:
Addresses issue #1629 

This PR adds a `add_filter_share_permissions` method to the client, supporting the `filter/{id}/permission` endpoint (https://docs.atlassian.com/software/jira/docs/api/REST/9.3.1/#api/2/filter-addSharePermission).

Also add support for it in the `create_filter()` method, allowing users to pass on permissions to add when creating a new filters.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -35,6 +35,7 @@ from typing import (
     Literal,
     SupportsIndex,
     TypeVar,
+    Union,
     no_type_check,
     overload,
 )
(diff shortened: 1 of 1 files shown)


Mazorius:
I look forward to this solution as I need this as well :)

Any idea what is open? Cannot see the logs? Will got that released.

Issue #1557, opened 2022-11-28
Move away from using the imghdr module as it is deprecated & will be removed in 3.13 https://docs.python.org/3/library/imghdr.html


dbaxa:
### Bug summary

Hi, as per away https://docs.python.org/3/library/imghdr.html the imghdr module is deprecated as of python 3.11 & will be removed in python 3.13. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

main

### Python Interpreter version

3.11

### Which operating systems have you used?

- [ ] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
1. Install jira in python 3.11
2. Import it
3. <path>jira/client.py:11: DeprecationWarning: 'imghdr' is deprecated and slated for removal in Python 3.13
    import imghdr
```
```


### Stack trace

```python
<path>jira/client.py:11: DeprecationWarning: 'imghdr' is deprecated and slated for removal in Python 3.13
    import imghdr
```
```


### Expected behaviour

No deprecation warning is raised.

### Additional Context

_No response_

tibu:
Do you know when this change could be part of a release?

NasaGeek:
Looks like 3.6.0, released last week, addressed this issue. I expect this can be closed.

PR #1680, opened 2023-06-07
Replace imghdr with Pillow's PIL

ABizzinotto:
Addresses issue #1557 

The `imghdr` module will be deprecated in Python 3.13. Currently, it throws a deprecation warning.

This PR replaces `imghdr` with the `PIL` module.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -99,6 +99,8 @@ parso==0.8.3
     # via jedi
 pickleshare==0.7.5
     # via ipython
+pillow>=2.1.0
+    # via jira (setup.cfg)
 pluggy==1.0.0
     # via pytest
 prompt-toolkit==3.0.38
--- a/constraints.txt
+++ b/constraints.txt
@@ -99,7 +99,7 @@ parso==0.8.3
     # via jedi
 pickleshare==0.7.5
     # via ipython
-pillow>=2.1.0
+pillow==9.5.0
     # via jira (setup.cfg)
 pluggy==1.0.0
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -51,6 +51,7 @@ zip_safe = False
 install_requires =
     defusedxml
     packaging
+    Pillow>=9.5.0
```
adehad, line 54: Could we loosen this requirement?

Ideally we want a minimum version, and then if we can add an entry to our `constraints.txt` with this version, that would be perfect.

adehad (changes requested):
Thanks for this, minor, but blocking, comment!

setup.cfg
ABizzinotto: @adehad done, I set the requirement to the minimum version required for this to work (2.1.0). Thanks!

constraints.txt
```diff
@@ -99,6 +99,8 @@ parso==0.8.3
     # via jedi
 pickleshare==0.7.5
     # via ipython
+pillow>=2.1.0
```
adehad, line 102:
```suggestion
pillow==9.5.0
```

setup.cfg
adehad: Awesome, made a small correction to the `constraints.txt` - we use a `==` locked/pinned version in that file.

Will merge once the CI is happy, thanks again!

adehad (approved):
Thanks for the contribution !

Issue #1646, opened 2023-04-13
how to set description with rich text style


woshiyushengzhou:
### Problem trying to solve

I want rich text in the description

studioj:
@woshiyushengzhou 
does this comment help?
https://github.com/pycontribs/jira/issues/1556#issuecomment-1332033220
If we're missing documentation you can always open a PR to update our examples 
https://jira.readthedocs.io/examples.html#
https://github.com/pycontribs/jira/tree/main/examples

PR #1681, opened 2023-06-07
Add rich text example to docs

ABizzinotto:
Addresses #1646 

This PR adds an example of how to use rich text to add comments to the docs.


adehad (approved):
Awesome

Issue #1431, opened 2022-07-20
add_watcher doesn´t work, returns 200, instead 204


JonasArroyo:
### Bug summary

**Using Jira Python API: doesn´t work:**

```
jira = JIRA(server='http://XXX.atlassian.net', basic_auth=('YYYY', jira_token))
response = jira.add_watcher('XYZ-4', 'User A' )
watchers = jira.watchers('XYZ-4')
```

response returns an answer HTTP 200, instead 204 and user is not added as watcher
watcher variable confirms that user is not added

**Using a POST query it works:**

```
import requests
from requests.auth import HTTPBasicAuth
import json

url = "https://XXX.atlassian.net/rest/api/2/issue/XYZ-4/watchers"

auth = HTTPBasicAuth(user, jira_token)

headers = {
   "Accept": "application/json",
   "Content-Type": "application/json"
}
test = requests.post(url, headers=headers, auth=auth)
print(test.status_code)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### Jira instance version

8.16.1

### jira-python version

main

### Python Interpreter version

3.10

### Which operating systems have you used?

- [ ] Linux
- [ ] macOS
- [X] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA

# 2. With an existing case already in Jira XYZ-4

# 3. Call the method: jira.add_watcher(case, user)

# 4. No error return, but user was not added

# 5. Tried with JIRA API v2 and it works.
```

### Stack trace

```python
No, all OK
```

### Expected behaviour

I would expect that user will be added or error thrown.

adehad:
@JonasArroyo, just checking that you have also tried using the accountId of the user instead of the name?

I wonder if this is one of those Jira Cloud GDPR related issues.

PR #1682, opened 2023-06-07
Use user_id when adding watcher

ABizzinotto:
I believe this is related to #1431 

The `add_watcher` method is failing when using a user name. I tried with v2 and v3 of the Cloud API. It worked with the user ID though. I believe this is related to Jira's GDPR issues.

Added a simple fix to retrieve the ID before adding the watcher. This is already done in the `delete_watcher` method.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2342,7 +2342,9 @@ class JIRA:
             Response
         """
         url = self._get_url("issue/" + str(issue) + "/watchers")
-        return self._session.post(url, data=json.dumps(watcher))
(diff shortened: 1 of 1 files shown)


adehad (approved):
Double checked with the docs and it looks good, the endpoint accepts a single string as the body with the user id / username etc,

supermitch:
This was a breaking change, and shouldn't this be backwards compatible? Adding a fix to search for user ID using email adds an API call when I already have the user IDs. Can't we verify the format first?

PR #1695, opened 2023-07-11
Fix issue_type_scheme method not working/returning correct type

DEADSEC-SECURITY:
This PR fixes a bug with the method `issue_type_scheme` where it was raising an error due to the Json Object not containing `schemes` which was expected to have the `List[IssueTypeSchemes]` object list.

Now the method properly converts the json data into the list of objects intended.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -4287,17 +4287,25 @@ class JIRA:
 
     @lru_cache(maxsize=None)
     def issue_type_schemes(self) -> list[IssueTypeScheme]:
-        """Get all issue type schemes defined (Admin required).
+        """
+        Get all issue type schemes defined.
 
-        Returns:
-            List[IssueTypeScheme]: All the Issue Type Schemes available to the currently logged in user.
+        @warning: Admin is required to perform this action
+
+        @return: List[IssueTypeScheme]: All the Issue Type Schemes available to the currently logged in user
         """
         url = self._get_url("issuetypescheme")
 
         r = self._session.get(url)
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -4285,18 +4285,38 @@ def permissionschemes(self):
...
+    def issue_type_schemes(
+        self, expand_projects: bool = False, expand_issue_types: bool = False
+    ) -> list[IssueTypeScheme]:
         """Get all issue type schemes defined (Admin required).
 
-        Returns:
-            List[IssueTypeScheme]: All the Issue Type Schemes available to the currently logged in user.
+        @param expand_projects: For each issue type schemes, returns information about the projects the issue type scheme is assigned to
+        @param expand_issue_types: For each issue type schemes, returns information about the issueTypes the issue type scheme have.
+        @return: List[IssueTypeScheme]: All the Issue Type Schemes available to the currently logged in user.
```
adehad, line 4295: google style docstrings

jira/client.py
```diff
@@ -4285,18 +4285,38 @@ def permissionschemes(self):
...
+            expand.append("projects")
+        if expand_issue_types:
+            expand.append("issueTypes")
+
+        params = {}
+        if expand:
+            params["expand"] = ",".join(expand)
+
         url = self._get_url("issuetypescheme")
+        print(url)
```
adehad, line 4308:
```suggestion
```

adehad (changes requested):
Look forward to the addressed comments!

Issue #1435, opened 2022-07-26
Error "ModuleNotFoundError: No module named 'keyring'" when trying to use jirashell


joegiglio:
### Bug summary

The jira package itself works fine, but if I try to run jirashell from the terminal, I get:

**ModuleNotFoundError: No module named 'keyring'**

Here is my current list of packages (`pip freeze`) from a venv I created yesterday:

certifi==2022.6.15
charset-normalizer==2.1.0
defusedxml==0.7.1
idna==3.3
jira==3.3.0
oauthlib==3.2.0
packaging==21.3
pyparsing==3.0.9
requests==2.28.1
requests-oauthlib==1.3.1
requests-toolbelt==0.9.1
typing_extensions==4.3.0
urllib3==1.26.11

If I manually install keying using pip and try again, I get "No module named 'IPython'"

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

3.3.0

### Python Interpreter version

3.9.10

### Which operating systems have you used?

- [ ] Linux
- [X] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Type jirashell from a terminal
```

### Stack trace

```python
Traceback (most recent call last):
  File "/Users/joe/dev/mine/jira/venv/bin/jirashell", line 5, in <module>
    from jira.jirashell import main
  File "/Users/joe/dev/mine/jira/venv/lib/python3.9/site-packages/jira/jirashell.py", line 15, in <module>
    import keyring
ModuleNotFoundError: No module named 'keyring'
```

### Expected behaviour

Jirashell should run without error

adehad:
Gotcha, I think this might be a documentation problem, as to use jirashell you need to 
`pip install jira[cli]`

joegiglio:
I tried that suggestion and I get:

```
mine/jira % pip install jira[cli]
zsh: no matches found: jira[cli]
```

I found a `jira-cli `and tried to install that but then got this error:

```
Collecting jira-cli
  Downloading jira_cli-3.0-py3-none-any.whl (27 kB)
Requirement already satisfied: jira in ./venv/lib/python3.9/site-packages (from jira-cli) (3.3.0)
Requirement already satisfied: requests in ./venv/lib/python3.9/site-packages (from jira-cli) (2.28.1)
Collecting suds-jurko
  Downloading suds-jurko-0.6.zip (255 kB)
[...]

PR #1709, opened 2023-08-04
#1435 - recommend installing jira[cli] to get jirashell deps

jefft:
`pip install jira` used to result in a working `jirashell`, but nowadays there is a missing dependency (see #1435), and `pip install 'jira[cli]'` is necessary. The [installation doc](https://github.com/pycontribs/jira/blob/main/docs/installation.rst) has been updated, but is inconsistent -- the first reference says `pip install jira`, and next paragraph says `pip install 'jira[cli]'`.

The attached patch fixes the inconsistency to always recommend `pip install 'jira[cli]'`, and adds a sentence explaining what `[cli]` does.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -3,9 +3,10 @@ Installation
 
 The easiest (and best) way to install jira-python is through `pip <https://pip.pypa.io/>`_::
 
-    pip install jira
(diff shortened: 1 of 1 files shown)


adehad (approved):
Thanks, good spot

Issue #1694, opened 2023-07-10
Search User does not return inactive users


nathan-gilbert:
### Bug summary

The following only returns active users: `search_users(maxResults=False, includeInactive=True, query="%")`

This generates a request like: `GET /rest/api/2/user/search?query=%25&includeActive=True&includeInactive=True&maxResults=100`

Which does not return inactive users, only active. 

In order to get all users (inactive + active) the following request needs to be generated (as verified with Postman): `/rest/api/2/users/search?includeInactive=true&query=%25&maxResults=100`

A workaround is to do: `_fetch_pages(User, None, "users/search", 0, 100, {"includeInactive": "true", "query": "%"})` but that isn't the intended usage of this library. 

This was tested with `Python 3.11.4`.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

3.5.2

### Python Interpreter version

3.11

### Which operating systems have you used?

- [ ] Linux
- [ ] macOS
- [X] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA
# 2. When I call the function with argument x
jira.the_function(x)
# 3.
...
```

### Stack trace

```python
None
```

### Expected behaviour

All users, inactive and active, are returned.

adehad:
The params need to run through `json.dumps()` in order to be in the correct format. We should probably do a quick audit of all instances where we pass in `params=` and ensure that all booleans are handles correctly.

Proposed work around for now:

```python
search_users(maxResults=False, includeInactive="true", query="%")
```

PR #1710, opened 2023-08-09
#1694 Fix Search User does not return active users

tusharrana786:
Passed the params through json.dumps() in order to generate query in correct format.


tusharrana786:
which inputs are you talking about exactly? to the search user function?

adehad (approved):

adehad:
Refer to #1780

PR #1711, opened 2023-08-10
Added raw api_get, api_put, api_post methods

dja852:
Added 3 new methods to the JIRA class (get, put, and post) to enable a user to send a raw API call leveraging the established JIRA session and return a Response object for further handling.

This is useful to interact with an API endpoint not yet supported by this package.


adehad:
I'm hesitant to add this, we have the existing `self._session` which would give more flexibility to users who do want to create their own methods.

I think this example works best for the GET requests but for the others where there are other keyword arguments that could be added it is unfortunately limiting.

Perhaps we can add some documentation on how to add custom methods, eg via subclassing to highlight the recommended way to handle unsupported endpoints?

Thank you for the effort of this contribution, sorry it didn't work out this time, but look forward to more!

PR #1712, opened 2023-08-15
Add skipUserResolution for assign_issue to decide whether to skip user resolution or not

yaooqinn:
Fix the issue I described at Apache/spark https://github.com/apache/spark/pull/42496. 

The assignee from the spark merge PR script is trustworthy, while calling assign_issue here may use a different user as we only search 20 candidates.


yaooqinn:
This request will be closed because we have not received any response from the admin.

PR #1713, opened 2023-08-15
Add handling of 503 Service Unavailable retries

gmainguet:
Hi,

This is a follow up of #925 and associated PR #1364 

As mentioned in https://developer.atlassian.com/cloud/jira/platform/rate-limiting/ , retries should not be limited to 429 responses.

503 responses are slightly different but should follow the same retry logic.

This PR aims at implementing Atlassian's pseudo code written in [that page](https://developer.atlassian.com/cloud/jira/platform/rate-limiting/#response-handling-pseudo-code) while keeping the already existing behaviour for 429s.

Unit tests have been updated accordingly.

Note: As a matter of fact, 503 responses were received from Jira Cloud (08-2023) and the library wasn't able to handle these. Hence this PR.


jira/resilientsession.py
```diff
@@ -290,11 +291,14 @@ def __recoverable(
         Returns:
             bool: True if the request should be retried.
         """
-        is_recoverable = False  # Controls return value AND whether we delay or not, Not-recoverable by default
+        suggested_delay = (
+            -1
+        )  # Controls return value AND whether we delay or not, Not-recoverable by default
```
mjurbanski-reef, line 296: right now this comment doesn't quite make sense

renaming this var to `recovery_delay` would make it much more readable

jira/resilientsession.py
```diff
@@ -304,45 +308,27 @@ def __recoverable(
...
+        elif isinstance(response, Response):
+            recoverable_error_codes = [
+                HTTPStatus.TOO_MANY_REQUESTS,
+                HTTPStatus.SERVICE_UNAVAILABLE,
+            ]
 
-                LOG.warning(warning_msg)
+            if response.status_code in recoverable_error_codes:
+                retry_after = response.headers.get("Retry-After")
+                if retry_after:
```
mjurbanski-reef, line 319: this looks like regression: 429 should be still always retried

gmainguet: 429s are retried, just with a different delay as per previous behaviour -- see line 322

jira/resilientsession.py
gmainguet: what would you change the comment to? (it makes sense to me but I'm ok to add more meaning to it)

I kind of agree with `recovery_delay`, but then it won't eventually be the recovery delay because of the max limit.

mjurbanski-reef: then you go with -1 right? anyhow, this was just a minor suggestion, don't worry about it

mjurbanski-reef (approved):

adehad (approved):
Thank you for this high quality contribution

Issue #883, opened 2019-12-10
Support filebased cookiejar


valentijnscholten:
**Is your feature request related to a problem? Please describe.**
The current cookie auth implementation does remember the session during the lifetime of the python script. Cookies are not stored on disk for the next run of the script. This entices users to put the password in plaintext in the sourcecode or a config file.

**Describe the solution you'd like**
Support filebased cookiejars to store the cookies on disk. This would allow for the cookies to be stored. The cookie only contains a JSESSION ID. This is still sensitive, but a lot less sensitive compared to a plaintext password, especially in corporate environment with ldap or other sso solutions.
The current library "Requests" seems to support FileCookieJar (and other externally provided cookiejars): https://kite.com/python/docs/requests.cookies.RequestsCookieJar

PR #1720, opened 2023-08-20
Add option to persist cookies

dosy4ev:
* Change `auth` param from tuple to dict to support optional `cookie_jar` param.
* Add option to provide callable for lazy password fetching.


dosy4ev:
fixes #883

adehad (changes requested):
(as mentioned please look into supporting existing use case too)

Issue #1721, opened 2023-08-25
resources.py __getattr__ causing stack overflow


sparrowt:
### Bug summary

In certain cirumstances `jira.resources.Resource.__getattr__` enters an infinite loop resulting in stack overflow (see output below).

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

9.4.8

### jira-python version

3.5.2

### Python Interpreter version

3.8

### Which operating systems have you used?

- [ ] Linux
- [ ] macOS
- [X] Windows

### Reproduction steps

```python
# 1. Set up a virtualenv and install the following
pip install jira==3.5.2 pytest==7.4.0

# 2. Run this test

from unittest import mock

import jira
from requests.models import Response


def test_stuff():
    # Simulate a broken response
    duff_response = Response()
    duff_response.status_code = 200
    duff_response.reason = 'foo'
    duff_response._content = b'{"something":"blah"}'

    with mock.patch('jira.resilientsession.ResilientSession.get', return_value=duff_response):
        # Try to log in - this fails with stack overflow
        jira.JIRA(
            'https://jira.example.net',
            token_auth='thisisatoken',
            validate=True,
        )
```
like this: `py.test test_stuff.py`
# 3. Observe stack overflow
See below
```


### Stack trace

```python
Fatal Python error: Cannot recover from stack overflow.
Python runtime state: initialized

Current thread 0x0000a32c (most recent call first):
  File "path\to\venv\lib\site-packages\jira\resources.py", line 222 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
  File "path\to\venv\lib\site-packages\jira\resources.py", line 224 in __getattr__
[...]

PR #1722, opened 2023-08-25
Avoid infinite recursion in Resource.__getattr__

sparrowt:
Otherwise the `self.raw` access within the except handler below comes back through and if the originally requested item isn't actually in self.raw then it loops forever.

Fixes: https://github.com/pycontribs/jira/issues/1721

I haven't added a test as it seems pretty niche but there is a repro test case on the issue linked above which we can incorporate if desired?


adehad:
@sparrowt would be great if you can add that test case into: https://github.com/pycontribs/jira/blob/main/tests/test_client.py

PR #1723, opened 2023-08-26
feat: add an option to use POST for search_issues

BlankTiger:
Addresses #701 and #548. Gives user an option to use POST endpoint instead of GET endpoint for fetching issues.


adehad (approved):
Awesome, appreciate the quality of this PR. It sounds like we may want to consider updating the docs to help people recognize when they might want to use this flag too.

For now merging !

PR #1776, opened 2023-12-12
migrate setup.cfg to pyproject.toml

deronnax:
Just a work in progress for now.


pyproject.toml
```diff
@@ -84,3 +84,112 @@ required-imports = ["from __future__ import annotations"]
 [tool.ruff.pydocstyle]
 # Use Google-style docstrings.
 convention = "google"
+
+[project]
```
adehad, line 88: can we put all the project specific config to the top of the file and all the tools at the bottom

adehad (changes requested):
yep no issues on the cloud build failure, small re-ordering but otherwise all good.

Thanks

pyproject.toml
deronnax: there you go.

adehad (approved):

adehad:
Thanks for the contribution!

PR #1778, opened 2023-12-18
Fix #1774 by updating the Accept value to be valid

### LINKED ISSUE #1774: Client incorrectly sends *.* for Accept header, instead of */*
### Bug summary

The client can send the HTTP Accept header as `application/json,*.*;q=0.9` which is invalid. This should be `application/json,*/*;q=0.9`. Note `*/*` in place of the incorrect `*.*`. A quick search shows that this comes from resilientsession.py line 165. Amazingly this does not currently fail when calling Jira Cloud, but the incorrect Accept header is (increasingly) likely to return a 406 error.

### Jira Instance type

### Jira instance version

latest

### jira-python version

any

any

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
from jira import JIRA

jira = JIRA('https://jira.atlassian.com/')

jira.server_info()
```

### Stack trace

```python
None yet.
```

### Expected behaviour

Expect not to get 406 error.

### Additional Context

_No response_

dbaxa:
Fix #1774 by updating the Accept value to be valid

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -162,7 +162,7 @@ class ResilientSession(Session):
         super().__init__()
 
         # Indicate our preference for JSON to avoid https://bitbucket.org/bspeakmon/jira-python/issue/46 and https://jira.atlassian.com/browse/JRA-38551
-        self.headers.update({"Accept": "application/json,*.*;q=0.9"})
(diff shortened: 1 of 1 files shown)


dbaxa:
cc @ssbarnea

adehad (approved):

PR #1786, opened 2023-12-26
fix: small fix for dev docker

studioj:
bumping to focal (hirsute is not maintained anymore)
and libkrb5 was needed otherwise there were weird install errors after running tox

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,8 @@
-# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.231.5/containers/ubuntu/.devcontainer/base.Dockerfile
+# See here for image contents: https://github.com/devcontainers/images/blob/main/src/base-ubuntu/.devcontainer/Dockerfile
 
-# [Choice] Ubuntu version (use hirsuite or bionic on local arm64/Apple Silicon): hirsute, focal, bionic
-ARG VARIANT="focal"
-FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+# [Choice] Ubuntu version: https://github.com/devcontainers/images/tree/main/src/base-ubuntu
+ARG VARIANT="ubuntu-22.04"
+FROM mcr.microsoft.com/vscode/devcontainers/base:${VARIANT}
 
 # [Optional] Uncomment this section to install additional OS packages.
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,7 @@
 # See here for image contents: https://github.com/devcontainers/images/blob/main/src/base-ubuntu/.devcontainer/Dockerfile
 
 # [Choice] Ubuntu version: https://github.com/devcontainers/images/tree/main/src/base-ubuntu
-ARG VARIANT="jammy"
+ARG VARIANT="ubuntu-22.04"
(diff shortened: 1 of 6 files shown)


adehad:
I think the key is to change the `.devcontainer.json` file as that overrode the variable in the Dockerfile.

Ran a test now where I built the devcontainer from that branch and it seems to have worked as expected:
image

It successfully used the jammy version and installed the project

adehad:
Regarding the test failing, I think it is now how when deleting a project the project does not get permanently deleted, but exists in a 'trash', so it is still 'found' by some endpoints but not others. A bit weird I thought I fixed with the right new parameter, but it seems not?

.devcontainer/Dockerfile
```diff
@@ -1,9 +1,9 @@
-# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.231.5/containers/ubuntu/.devcontainer/base.Dockerfile
+# See here for image contents: https://github.com/devcontainers/images/blob/main/src/base-ubuntu/.devcontainer/Dockerfile
 
-# [Choice] Ubuntu version (use hirsuite or bionic on local arm64/Apple Silicon): hirsute, focal, bionic
-ARG VARIANT="hirsute"
-FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+# [Choice] Ubuntu version: https://github.com/devcontainers/images/tree/main/src/base-ubuntu
+ARG VARIANT="jammy"
```
adehad, line 4:
```suggestion
ARG VARIANT="ubuntu-22.04"
```

Shall we use the numeric or name based reference here? I personally find the numeric easier to compare at a glance if we are on the latest or not

studioj: ah you pushed here ... i didnt see :D

tests/test_client.py
```diff
@@ -37,7 +37,7 @@ def slug(request, cl_admin):
 
     def remove_by_slug():
         try:
-            cl_admin.delete_project(slug)
+            cl_admin.delete_project(slug, enable_undo=False)
```
studioj, line 40: oho is it that?

adehad: I *think* so, I also had to manually restore the project from the 'trash'. I'm not sure why. We shall see if you tests pass too.

adehad: what is strange is how it is the same project key for all the commits.

adehad: hmm it still not using the GH prefix, but  I think this is because it is not a default env var

adehad: trying once more to see if I can get it to pick up the env vars, otherwise I think we can say we have fixed the tests for now

adehad: So now it is correctly picking up the env vars:
see the CI prefixed proj name: https://github.com/pycontribs/jira/actions/runs/7338468278/job/19981037988?pr=1786#step:7:457

But note how we are getting a warning regarding the project key being >10 chars
```
  E           	response text = {"errorMessages":[],"errors":{"projectKey":"The project key must not exceed 10 characters in length."}}
```

adehad: so looks like this has been fixed now, using the correct prefix, but the projects are not fully deleted from the Cloud instance. Might be worth looking into that as I guess the instance will need periodic purging otherwise.

adehad: I've implemented a specific garbage collection handler to truly clear out any dangling projects.

studioj:
> sorry for making this more than a 'small fix'

was intending to do a small one but then i saw CI  failing :D

PR #1787, opened 2023-12-27
Added warning log if issue link type not present

SageTheWizard:
This error correction created problems for me, resulting in numerous issues to have inverted links (i.e. Blocks / Is Blocked By was inverted). 
Debugging the issue was difficult and required to look into this code to determine what actually happened, if a log message was present, it would have simplified my debugging, so I've added it!

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2524,7 +2524,7 @@ class JIRA:
         issue_link_types = self.issue_link_types()
 
         if type not in issue_link_types:
-            LOG.warning(
+            self.log.warning(
                 "Warning: Specified issue link type is not present in the list of link types"
             )
             for lt in issue_link_types:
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -2524,6 +2524,9 @@ def create_issue_link(
         issue_link_types = self.issue_link_types()
 
         if type not in issue_link_types:
+            LOG.warning(
```
adehad, line 2527:
```suggestion
            self.log.warning(
```

We use this logger object !

adehad (changes requested):
Small change, glad you got it sorted.

studioj:
master seems ok hopefully lets rebase and see

studioj (approved):

studioj:
> Small change, glad you got it sorted.

changes?

SageTheWizard:
I see there's a failure deploying to the cloud, if this is something caused by my changes please let me know so that I can change as needed.  (I'm doubtful as it's just a log print though :) )

adehad (approved):
Thanks for this contribution!

Issue #1761, opened 2023-11-17
Add new JIRA Teams API


Maxim-Durand:
### Problem trying to solve

I think it's currently not possible to fetch JIRA's team members using this client.

### Possible solution(s)

Jira launched a new rest-api at https://developer.atlassian.com/platform/teams/rest/v1/api-group-teams-members-public-api/#api-group-teams-members-public-api to support this workflow.
I think it's only about adding this into the current client for a new release.

### Alternatives

Instead of using https://github.com/pycontribs/jira, the user can use the REST API
For instance here is code snippet to fetch the members of a team  (you'll need to fill the `jira_.*` variables with your own values).
```
from requests.auth import HTTPBasicAuth

url = f"{jira_server_url}/gateway/api/public/teams/v1/org/{jira_org_id}/teams/{jira_team_id}/members"
headers = {
    "Accept": "*/*",
    "Content-Type": "application/json"
}
payload = json.dumps({
    "first": 40
})
response = requests.request(
    "POST",
    url,
    auth=HTTPBasicAuth(jira_api_id, jira_api_token),
    data=payload,
    headers=headers
)
print(response.text)
```

Maxim-Durand:
I'll try to create a PR to add this feature.

Sam-Hepworth:
I use "<xxx>.atlassian.net/gateway/api/v3/teams/<teamId>/members".

PR #1794, opened 2024-01-07
Adding teams API

Maxim-Durand:
https://github.com/pycontribs/jira/issues/1761

> [!IMPORTANT]
> This PR isn't finished as I'm currently stuck not being able to test my changes. 
Indeed the jira server container used to test is too old and doesn't support newest API like the [Jira Team API](https://developer.atlassian.com/platform/teams/rest/v1/api-group-teams-members-public-api/#api-group-teams-members-public-api) or the [Jira Org API](https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-organization/#api-group-organization)
> Hence, I'm asking the maintainer of this project what is the strategy to keep this project up-to-date with Jira newest APIs ? Do we have a cloud server we could use to test ?


adehad:
Hi @Maxim-Durand, we do have a Jira Cloud instance. However, the credentials for this are not shared with pull requests due to a limitation within GitHub so the process would likely involve the maintainers running in separately to confirm functionality.

jira/client.py
```diff
@@ -1309,6 +1310,188 @@ def update_filter(
...
+            org_id (str): organization identifier
+            team_id (str): team identifier
+            site_id (Optional[str])
+
+        Returns:
+            Team
+        """
+        url = f"gateway/api/public/teams/v1/org/{org_id}/teams/{team_id}"
+        if site_id is not None:
+            url += f"?siteId={site_id}"
```
adehad, line 1360: url parameters should be passed in with `params=`

jira/client.py
```diff
@@ -1309,6 +1310,188 @@ def update_filter(
...
+            team_id (str): team identifier
+
+        Returns:
+            bool
+        """
+        url = f"gateway/api/public/teams/v1/org/{org_id}/teams/{team_id}"
+        r = self._session.delete(url)
+        if r.status_code == 204:
+            return True
+        return False
```
adehad, line 1383:
```suggestion
        return r.ok
```

Would be the preferred way if you just wanted a boolean

jira/resources.py
```diff
@@ -1234,6 +1235,25 @@ def __init__(
...
+class Team(Resource):
+    """A Jira team."""
+
+    def __init__(
+        self,
+        options: dict[str, str],
+        session: ResilientSession,
+        raw: dict[str, Any] = None,
+    ):
+        TEAM_API_BASE_URL = "{server}/gateway/api/public/teams/v1/"  # rest/{rest_path}/{rest_api_version}/{path}"
```
adehad, line 1247: The all caps naming convention does not need to be used for variables inside the init method, look at the `AgileResource` example which is more similar to what you intend to do here.

tests/tests.py
```diff
@@ -619,6 +623,27 @@ def test_add_and_remove_user(self):
         result = self.jira.delete_user(self.test_username)
         assert result, True
 
+    def test_create_team(self):
```
adehad, line 626: I would look to create any tests in a new `tests/resources/test_teams.py`

I would look at `test_filter`, the way it uses a context manager may also be helpful for you

tests/tests.py
```diff
@@ -619,6 +623,27 @@ def test_add_and_remove_user(self):
...
+        sleep(2)  # avoid 500 errors
+        result_team = self.jira.create_team(
+            self.org_id,
+            self.test_team_description,
+            self.test_team_name,
+            self.test_team_type,
+        )
+        self.assertEqual(
+            self.test_team_name,
+            result_team.__getattr__("displayName"),
```
adehad, line 643:
```suggestion
            result_team["displayName"],
```

tests/tests.py
```diff
@@ -619,6 +623,27 @@ def test_add_and_remove_user(self):
         result = self.jira.delete_user(self.test_username)
         assert result, True
 
+    def test_create_team(self):
+        if self._should_skip_for_pycontribs_instance():
+            self._skip_pycontribs_instance()
```
adehad, line 628: We use a test decorator to enable the tests running on the Jira Cloud instance, this is denoted by `@allow_on_cloud`, an example is in `test_group`

jira/client.py
```diff
@@ -1309,6 +1310,188 @@ def update_filter(
...
+
+        Args:
+            team_id (str): Id of the team.
+            org_id (str): Id of the org.
+
+        Returns:
+            list[str]
+        """
+        url = f"/gateway/api/public/teams/v1/org/{org_id}/teams/{team_id}/members"
+        payload = {"first": 50}
```
adehad, line 1428: this pagination implementation should be moved into a helper method, if is is specific to the Teams API I would make it a static method of the Teams Resource

Then you would call it like:

`Teams._fetch_pages(session=session, url=url, ...)`

I would aim to match as many of the parameters to the function `JIRA._fetch_pages()` if possible

adehad (changes requested):
lots of effort into this, great start to your first PR

jira/client.py
Maxim-Durand: In theory it's not specific to Teams API, but isn't exactly like `JIRA._fetch_pages()` neither as it's using different identifiers for Jira's latest API (why they didn't stick to the same variable names between API idk :roll_eyes: )

For now I simply created it as a helper function but if it it's used outside of Teams it's bound to change.

tests/resources/test_teams.py
```diff
@@ -0,0 +1,80 @@
...
+from tests.conftest import JiraTestCase, allow_on_cloud
+
+
+@allow_on_cloud
+class TeamsTests(JiraTestCase):
+    def setUp(self):
+        JiraTestCase.setUp(self)
+        self.test_team_name = f"testTeamFor_{self.test_manager.project_a}"
+        self.test_team_type = "OPEN"
+        self.org_id = os.environ["CI_JIRA_ORG_ID"]
```
Maxim-Durand, line 15: I'm not sure how to get the `org_id` of the testing organisation :thinking:

One solution would be to implement the organisation API and to create a new one before running the tests ?

Maxim-Durand:
> However, the credentials for this are not shared with pull requests due to a limitation within GitHub so the process would likely involve the maintainers running in separately to confirm functionality.

Although I understand the reasoning this seems like a very slow and inefficient process.
Is there no way to simply share an API token with limited permissions and for a limited time-frame so I can actually test this without spamming maintainers ?

tests/resources/test_teams.py
Maxim-Durand: I created a new PR (#1803) based on this one but also adding Organisations API so that I'm able to test the Team API.

Maxim-Durand:
https://github.com/pycontribs/jira/pull/1794#discussion_r1460508561

I think I'll close this PR in favor of #1803

Maxim-Durand:
> Typically I would expect contributors to have tried their API with their own instances (or other public instances, as shown in https://github.com/pycontribs/jira/blob/main/examples/basic_use.py). At very least the "get" based methods.

Fair enough, I'll do that.

PR #1798, opened 2024-01-12
client: Cache raw field information

lhh:
The field JSON from the JIRA server contains important information, such as schemas, which can be used by applications building on this library. Since it's generally preferred, whenever possible, to reduce unnecessary calls to JIRA API endpoints, we can keep this information in the client object.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -672,6 +672,14 @@ class JIRA:
             JIRA.checked_version = True
 
         self._fields_cache_value: dict[str, str] = {}  # access via self._fields_cache
+        self._fields_cache_value_raw: list[dict[str, Any]] = []  # access via self._fields_cache_raw
+
+    @property
+    def _fields_cache_raw(self) -> list[dict[str, Any]]:
+        """Cached raw dictionary of of /field endpoint. Lazy loaded."""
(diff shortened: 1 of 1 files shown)


lhh:
There doesn't seem to be a way for me apply the label after submitting the PR, but this would fall under 'enhancement'.

lhh:
The py38 failure seems to be unrelated to the change here.

adehad:
Have you considered the possibility for using a `functools.lru_cache` decorator on the method?
That would also give us some helper methods, like the ability to clear the cache

(as you say the cloud test failings are unrelated, so no need to worry)

lhh:
It would be possible to do that.

The rate of change of the information in `/rest/api/latest/field` is not typically very fast, but it is certainly true that the information does get updated, and that re-fetching it periodically is a good idea in long-running processes.

Presently, the value in `_fields_cache_value` is not done that way - once it is populated, it is not updated from the JIRA server. This MR follows that precedent.

If we do decide use `functools.lru_cache` decorator with `_fields_cache_value_raw`, I think we need to be careful to ensure that the value in `_field_cache_value` is updated whenever the value of `_fields_cache_value_raw` is updated; these data sets should be consistent.

PR #1803, opened 2024-01-20
Adding support for Teams and Organisations API

Maxim-Durand:
Resolves the issue https://github.com/pycontribs/jira/issues/1761

#1794 Started the needed work to add the Teams API but couldn't get around testing it properly as it needed to create an organisation for testing.

Hence this PR is based on #1794 but also adds support for Organisations management.


Maxim-Durand:
@adehad I'd be very grateful if you could share with me the test results on the cloud instance :pray:

PR #1806, opened 2024-02-01
Add goal field to update/create sprint

zbarahal:
https://github.com/pycontribs/jira/pull/1586 seems like it stalled out after you asked for tests so I created a separate PR.

This adds goal as a field to create/update sprint and adds test cases. There weren't any existing test cases for update_sprint so I extended that.

I'm still working on getting the local dev setup but I figure I can see how the tests run in CI

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -5073,7 +5073,7 @@ class JIRA:
         if goal:
             payload["goal"] = goal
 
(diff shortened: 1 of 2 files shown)


adehad (approved):
changes look good to me! Thanks for this effort

PR #1820, opened 2024-02-12
Update createmeta warning with new method names

mesozoic:
The warning message that comes up when calling `createmeta()` on Jira Server 9.x refers to methods that have since been renamed. This branch fixes the warning message.


adehad (approved):
Great spot, thank you

Issue #1805, opened 2024-01-26
Regression of 429 response handling in 3.6.0


matthias-bach-by:
### Bug summary

While the improved handling of the retry-after header that got introduced in 3.6.0 via #1713 is an important improvement, it sadly introduced two  regressions in the behaviour when handling 429 responses.
1. While the retry-after value is finally evaluated and not just printed, it is sadly used as an upper bound for the backoff time. Thus, we are almost guaranteed to run into a second 429 response when handling the first one.
2. The code assumes requests that have a retry-after value of 0 must not be retried. However, Jira seems to frequently provide a retry-after value of 0 on 429 responses. 

With the previous plain backoff you had a good chance of backing off enough to happen to evade the rate limiting as long as you weren't running parallel requests. With the new behaviour we are a lot more aggressive and even with a single client easily run into the case of not even retrying.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

9.12.2

### jira-python version

3.6.0

### Python Interpreter version

3.12

### Which operating systems have you used?

- [X] Linux
- [X] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA
# 2. Running a random cheap operation often enough
for _ in range(100):
    jira.issue('SOME-1')
```

### Stack trace

```python
E           jira.exceptions.JIRAError: JiraError HTTP 429 url: https://jira.example.org/rest/api/2/search?jql=issuekey+in+%28SOME-1&startAt=0&validateQuery=True&fields=issuetype&fields=resolution&fields=status&maxResults=100
E           	text: Rate limit exceeded.
E
E           	response headers = {'Content-Type': 'application/json;charset=UTF-8', …}
E           	response text = {"message":"Rate limit exceeded."}
```

### Expected behaviour

I'd expect the value of the retry-after header to be interpreted as a lower bound for the backup. I.e., doing something along the following:

    delay = suggested_delay + self.max_retry_delay * random.random()

Furthermore, we should also retry requests that have a `suggested_delay` of 0. If that would indicate an error for 503 requests, we might need to make the decision depend on the return code, too.

PR #1825, opened 2024-02-26
Improve handling of Jira's retry-after handling

matthias-bach-by:
As reported in  #1805, while 3.6.0 brought a beneficial change in evaluating Jira's retry-after header, in some scenarios it actually degraded the success rate of requests when hitting rate-limits due to two issues:
1. Jira intends the retry-after header to be a minimum backoff but the code treats it as a maximum backoff, and
2. Jira sometimes sends a retry-after value of 0 seconds which the existing code implicitly treated as a failed request.

To address this, this PR changes the back-off behaviour to treat the retry-after time as a minimum backoff, at least as long as our maximum back-off time does allow this, and clips the retry-after header, if specified, to a minimum of one second.

I have extended the existing tests for the retry logic to assert this behaviour. If you would prefer a separate set of tests or see other changes to the approach, let me know and I'll adjust the PR.

### DIFF
--- a/jira/resilientsession.py
+++ b/jira/resilientsession.py
@@ -315,7 +315,7 @@ class ResilientSession(Session):
             if response.status_code in recoverable_error_codes:
                 retry_after = response.headers.get("Retry-After")
                 if retry_after:
-                    suggested_delay = int(retry_after)  # Do as told
+                    suggested_delay = 2 * int(retry_after)  # Do as told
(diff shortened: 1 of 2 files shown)


matthias-bach-by:
The cloud tests are failing in set-up and this seems to be a wider issue unrelated to my changes as I am also seeing that on other PRs. Happy to rebase once that is fixed on main.

adehad:
Thanks for the excellent PR and sorry for the delay in getting in merged and released!

PR #1834, opened 2024-03-05
Add issue.remove_field_value()

nickbroon:
Complements the existing add_field_value()

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -893,25 +893,25 @@ class Issue(Resource):
         else:
             return getattr(self.fields, field_name)
 
-    def add_field_value(self, field: str, value: str):
+    def add_field_value(self, field: str, value: Any):
         """Add a value to a field that supports multiple values, without resetting the existing values.
 
         This should work with: labels, multiple checkbox lists, multiple select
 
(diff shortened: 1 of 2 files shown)


adehad:
@nickbroon thanks for this! Do you think we can add a test (to: tests\resources\test_issue.py) where perhaps we 

1. add the value (perhaps to the label field)
2. check the value is added
3. remove the value
4. check the value is removed
?

nickbroon:
> @nickbroon thanks for this! Do you think we can add a test (to: tests\resources\test_issue.py) where perhaps we
> 
> 1. add the value (perhaps to the label field)
> 2. check the value is added
> 3. remove the value
> 4. check the value is removed
>    ?

I added a test that exercises both `remove_field_value()` and `add_field_value()`

nickbroon:
The UT failures look unrelated to the changes in the PR.

nickbroon:
Rebased to latest main.

I don't have permission to update PR labels. "enhancement" needs changed to "feat"

nickbroon:
@adehad I think this is ready for review and consideration of merging.

Issue #1836, opened 2024-03-12
Flesh out Dashboard Endpoints


jpavlav:
### Problem trying to solve

I have a project at work where we are creating and updating dashboards dynamically. This is the library we primarily use for interacting with Jira in our `python` code, so I figured it makes sense to expose those endpoints if possible.

### Possible solution(s)

The solution I came up with is to create the appropriate calls in the `client` and the corresponding `resource`s as well to flesh out this functionality. Per @studioj, I've been conscious of the fact that these calls are both only available in the `cloud` and many are considered `experimental`.

jpavlav:
If someone with the power to can, please assign this to me 😄 @studioj

PR #1837, opened 2024-03-12
ISSUE-1836: Add `Dashboard` Support

jpavlav:
Addresses ISSUE-1836
#1836 
The impetus for these changes is a project I'm working on at work. We want to create a dynamically updating dashboard that helps prioritize maintenance operations. I realize this is a large PR, and one large commit, but I feel like the work in it's entirety makes sense when looked at together. Additionally, there are some changes that were just formatting, in the case where spacing or anything along those lines have changed, that was auto-formatting and accepted by the linting `tox -e lint`.

`jira/client.py`
----------------
* Imported `DashboardItemProperty`, `DashboardItemPropertyKey`, and `Gadget` resources to client for use in new methods.
* Updated the `dashboards` method to include the `gadgets` that exist on a given dashboard. This is a logical association that makes sense, but isn't directly exposed in the API.
* Added `create_dashboard` method. It creates a dashboard via the API and returns a `Dashbord` object.
* Added `copy_dashboard` method.
* Added `update_dashboard_automatic_refresh_seconds` method. This calls the `internal` API, which is why it's decorated with `experimental` and `cloud`. This might change in the future, but it really is a handy thing to have, otherwise, the user has to configure this in the web interface.

---
* Added `dashboard_item_property` method. This is available on both `cloud` and `server`.
* Added `dashboard_item_property_keys` method. This is available on both `cloud` and `server`.
* Added `set_dashboard_item_property` method. This is available on both `cloud` and `server`.
---
^^ These methods all provide a means of adding arbitrary metadata to `dashboard_items` (`gadgets`) and/or configure them via specific keys.

* Added `dashboard_gadgets` method. This returns the gadgets associated with a given dashboard. It also iterates over the `keys` for this `gadget`'s properties, generating a list of `DashboardItemProperty` objects that are associated with each gadget. This makes it really easy for the user to associate which configuration/metadata goes with which gadget.
* Added `gadgets` method. This returns a list of from `jira` of all the `gadgets` that are available to add to any dashboard.
* Added `add_gadget_to_dashboard` method. This allows the user to add gadgets to a specified dashboard.
* Added the protected method `_get_internal_url`. This is very similar to `get_url` or `get_latest` url, where `options` are updated to allow for easy resolution of paths that are on the `internal` `jira` api.
* Updated the `_find_for_resource` type hint on `ids` because it is possible that a resource requires more than `2` ids to resolve it's url.

jira/resources.py
-----------------
* Added the new resources `DashboardItemProperty`, `DashboardItemPropertyKey`, and `Gadget` to the `__all__` list so they are represented.
* Updated the typehint on the `find` method for the same reason as the `_find_for_resource` method.
* Added a `gadgets` attribute to the `Dashboard` resource to house `gadget` references.
* Added `DashbordItemPropertyKey` resource.
* Added `DashboardItemProperty` resource. The `update` and `delete` methods are overridden here because it does not have a `self` attribute. This is kind of in an in between space as far as being considered a resource, but for ease of use as an interface, it makes sense for it to be considered.
* Added `Gadget` resource. It too has overridden `update` and `delete` methods for the aforementioned reasons.

jira/utils/__init__.py
----------------------
* Added `cloud` convenience decorator for client methods that make calls that are only available on the `cloud` api. It checks the `client` instance to see if it `_is_cloud`. If not, it logs a warning and returns `None`. This was the convention seen on other endpoints on the `client`.
* Added `experimental` convenience decorator for client methods that make calls that are experimental. It attempts to run the client method, if a `JIRAError` is raised that has a response object, the response is checked for a status code in `[404, 405]` indicating either the path no longer accepts the HTTP verb or no longer exists, and then logs a warning and returns `None`. Otherwise it reraises the error

^^ Regarding these convenience decorators, it makes it really apparent to developers which calls are for the cloud and/or experimental.

* Added `remove_empty_attributes` convenience method. I found myself having to remove empty attributes or add a lot of branching in order to accommodate optional payload parameters or path parameters. This function made that easier.

Finally I wrote tests to cover these changes. These tests worked for me locally against my docker `jira server` instance. However, most of them didn't run, because the test checks the environment to find out where it is running.

For the cloud tests, I signed up for a trial account for `jira cloud` and ran them against my instance. They all panned out okay and cleaned up after themselves appropriately.


jira/client.py
```diff
@@ -1203,7 +1213,240 @@ def dashboard(self, id: str) -> Dashboard:
...
+        url = f"{url}/{id}/copy"
+        r = self._session.post(url, json=data)
+
+        raw_dashboard_json: dict[str, Any] = json_loads(r)
+        return Dashboard(self._options, self._session, raw=raw_dashboard_json)
+
+    @cloud
+    @experimental
+    def update_dashboard_automatic_refresh_seconds(
+        self, id: str, minutes: int
```
studioj, line 1297: update the name of the method or parameter to allign ... try to use the wording of the jira webui

jpavlav: Thanks @studioj! I'll take a look again at the UI for this option and respond with a bit more context.

jpavlav: Oh yeah, I see sorry about that! I'll change it to `refresh_minutes`:

image

☝️ that's what it looks like in the UI. I'll push up the change momentarily.

jpavlav: Fixed [here](https://github.com/pycontribs/jira/pull/1837/files#diff-442d14b20d591afd0c3de830d2803e30d4816eedc708f9ad129316aa7e5a4a7cR1296).

studioj (dismissed):

jira/resources.py
```diff
@@ -552,8 +556,157 @@ def __init__(
...
+        self.item_properties: list[DashboardItemProperty] = []
         self.raw: dict[str, Any] = cast(Dict[str, Any], self.raw)
 
+    def update(  # type: ignore[override] # incompatible supertype ignored
+        self,
+        dashboard_id: str,
+        color: str | None = None,
+        position: dict[str, Any] | None = None,
+        title: str | None = None,
+    ) -> Resource:
```
adehad, line 659:
```suggestion
    ) -> DashboardGadget:
```

jira/resources.py
```diff
@@ -552,8 +556,157 @@ def __init__(
...
+        Resource.__init__(
+            self, "dashboard/{0}/items/{1}/properties/{2}", options, session
+        )
+        if raw:
+            self._parse_raw(raw)
+        self.raw: dict[str, Any] = cast(Dict[str, Any], self.raw)
+
+    def update(  # type: ignore[override] # incompatible supertype ignored
+        self, dashboard_id: str, item_id: str, value: dict[str, Any]
+    ) -> Resource:
```
adehad, line 596:
```suggestion
    ) -> DashboardItemProperty:
```

jira/utils/__init__.py
```diff
@@ -79,3 +82,79 @@ def json_loads(resp: Response | None) -> Any:
...
+    Args:
+      client_method: The method that is being called by the client.
+
+    Returns:
+      Either the result of the wrapped function or None.
+    """
+
+    @functools.wraps(client_method)
+    def check_if_cloud(*args, **kwargs):
+        # NOTE(jpavlav): The first argument of any class instance is a `self`
```
adehad, line 103: On this point, you may want to check if the first argument is an instance of the JIRA class. As otherwise this decorator might be used incorrectly

jira/utils/__init__.py
```diff
@@ -79,3 +82,79 @@ def json_loads(resp: Response | None) -> Any:
         if not resp.text:
             return {}
         raise
+
+
+def cloud(client_method: Callable) -> Callable:
```
adehad, line 87: Part of me is nervous about such a simple name, but seeing it in use it does look fine. But I wonder if there are other names we can use

jira/utils/__init__.py
```diff
@@ -79,3 +82,79 @@ def json_loads(resp: Response | None) -> Any:
...
+        instance.log.warning(
+            "This functionality is not available on Jira Data Center (Server) version."
+        )
+        return None
+
+    return check_if_cloud
+
+
+def experimental(client_method: Callable) -> Callable:
+    """A convenience decorator to inform if a client method is experimental.
```
adehad, line 118: It might make sense to highlight experimental from which side ie ours or Atlassian here

tests/resources/test_dashboard.py
```diff
@@ -29,3 +52,364 @@ def test_dashboard(self):
         dashboard = self.jira.dashboard(expected_ds.id)
         self.assertEqual(dashboard.id, expected_ds.id)
         self.assertEqual(dashboard.name, expected_ds.name)
+
+    @pytest.mark.skipif(
+        os.environ.get("CI_JIRA_TYPE", "Server").upper() != "CLOUD",
+        reason="Functionality only available on Jira Cloud",
+    )
```
adehad, line 59: Let's make this into a similar style as ``@allow_on_cloud`` maybe the name can be only_run_on_cloud

tests/resources/test_dashboard.py
```diff
@@ -29,3 +52,364 @@ def test_dashboard(self):
...
+    @allow_on_cloud
+    def test_copy_dashboard(self):
+        original_dashboard = self.jira.create_dashboard(
+            name=rndstr(), share_permissions=[{"type": "authenticated"}]
+        )
+        self.dashboards_to_delete.append(original_dashboard)
+        # NOTE(jpavlav): Add something to the dashboard so we can test the copy worked
```
adehad, line 129: As these comments are actually describing the test flow, it is better to use the

Given, When, Then

Syntax for the test comments.

Typically you associate your name with TODOs more than notes like this, as the git history can always tell us who specifically.
There are quite a few instances throughout this PR and it would be good to identify which should be turned into normal comments vs. which were your notes while you were developing. The latter which we should not be leaving in the code

adehad (changes requested):
An excellent quality PR, still a few things that I think should be addressed before merging, but a promising start!

jpavlav:
Thanks for taking a look! I'll make another commit with some updates addressing your feedback.

jira/client.py
```diff
@@ -1203,7 +1213,240 @@ def dashboard(self, id: str) -> Dashboard:
...
+        return dashboard
+
+    @cloud
+    @experimental
+    def create_dashboard(
+        self,
+        name: str,
+        description: str | None = None,
+        edit_permissions: list[dict] | list = [],
+        share_permissions: list[dict] | list = [],
```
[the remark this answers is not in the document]
jpavlav: Ah yes! Mutable function arguments. These only get interpreted once so it can lead to unforeseen behavior. I'll take care of that.

jira/client.py
```diff
@@ -1203,7 +1213,240 @@ def dashboard(self, id: str) -> Dashboard:
...
+        return self._fetch_pages(
+            DashboardItemPropertyKey,
+            "keys",
+            f"dashboard/{dashboard_id}/items/{item_id}/properties",
+        )
+
+    def dashboard_item_property(
+        self, dashboard_id: str, item_id: str, property_key: str
+    ) -> DashboardItemProperty:
+        """Get a gadget Resource from a specific dashboard the server.
```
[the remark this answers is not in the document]
jpavlav: ah, sorry about this. I normally don't do such large PR's for exactly this reason, was just an oversight. Getting to that now.

jira/resources.py
jpavlav: Ah yes, I think this one and the one after it are probably leftover copy - paste from the parent class' `update` method. Good catch.

jira/utils/__init__.py
jpavlav: Yeah, I considered that, but thought that the utilization was pretty clear, and that if there was a failure, it would be readily apparent why. Having said that, I can go ahead and make that adjustment. I think the only other concern I had was circular import dependency. Since `jira.client` depends on `jira.utils`, but I think it would only matter if something from `jira.utils` was actually a member of one of the classes in `client`. Either way, will see what I can do.

jpavlav: Yeah, what I was afraid of here is the case. I can't check if the `self` argument `isinstance` of the `JIRA` class because that would necessitate importing it and causes a circular dependency:

```python
ImportError: cannot import name 'JIRA' from partially initialized module 'jira.client' (most likely due to a circular import) (/Users/jpalmer/Repositories/github.com/Personal/jira/jira/client.py)
```

I guess I'll refactor this a bit. I'll put these decorators in the `client`. I don't think it's unreasonable for them to exist there, they pertain to the client specifically so it should be fine.

jira/utils/__init__.py
jpavlav: Changed this to `cloud_api`. I think that's a little clearer hopefully.

jira/client.py
```diff
@@ -3875,6 +4120,24 @@ def _set_avatar(self, params, url, avatar):
         data = {"id": avatar}
         return self._session.put(url, params=params, data=json.dumps(data))
 
+    def _get_internal_url(self, path: str, base: str = JIRA_BASE_URL) -> str:
+        """Returns the full url based on Jira base url and the path provided.
+
+        Using the API version specified during the __init__.
```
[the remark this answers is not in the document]
adehad: We are overriding the version with `latest` so this line should be reviewed and removed if necessary

jira/resources.py
```diff
@@ -593,7 +593,7 @@ def __init__(
 
     def update(  # type: ignore[override] # incompatible supertype ignored
         self, dashboard_id: str, item_id: str, value: dict[str, Any]
-    ) -> Resource:
+    ) -> DashboardItemProperty:
```
adehad, line 596: recall we need to update the corresponding docstring return type

adehad (changes requested):
Nice, there is one pending thing we need to look at here.

And that is the `resource_class_map` in the `resources.py` file: https://github.com/pycontribs/jira/blob/c51c8a3cc4ac4ad65c6870eb8c60bf84ee7ed6ea/jira/resources.py#L1647

What we are looking at is a regex mapping of the URLs to the resource types.
From what I can see from the REST API reference we are dealing with:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-dashboards/#api-rest-api-2-dashboard-get

```
# DashboardItemProperty
/rest/api/2/dashboard/{dashboardId}/items/{itemId}/properties
# Example regex pattern for the map:
# https://regex101.com/r/1Nw7U6/1
# r"dashboard/[^/]+/items/[^/]+/properties$"

# DashboardItemKey
/rest/api/2/dashboard/{dashboardId}/items/{itemId}/properties/{propertyKey}

# DashboardGadget
/rest/api/2/dashboard/{dashboardId}/gadget/{gadgetId}
```

Once this is in we can add to your existing tests an `isinstance()` check on the class type

jira/resources.py
jpavlav: Ah! I thought I got all of those.

jpavlav:
@adehad I think I've got it now. I added the paths to the new resources. Added some more data points for the parameterized `test_generic_resource` that cover the new additions to that resource map dict. I also peppered the dashboard tests themselves with `isinstance` assertions where I thought it made sense. The cloud tests are still passing locally for me so figured 👍 . Let me know if there is anything else I need to address please 😄

adehad (approved):
excellent. I'll need to wait till later to actually run the cloud tests to merge (@studioj if you are able to confirm feel free to squash merge)

PR #1784, opened 2023-12-23
Issue Field and Issue Type object based methods

adehad:
Supersedes #1626.

Original credit goes to: @ddelabru

# Summary of Further Changes to original PR
1. Migrates previous Field and Issue Type resource test
2. Update docstring to highlight we are using the createmeta API for anyone searching through.

I also used this as a way to sort through the Jira Cloud test failures.

PR #1838, opened 2024-03-15
add backward compatibility for createmeta_issuetypes & createmeta_fieldtypes

paminov:
pr #1784 replaced `createmeta_issuetypes` & `createmeta_fieldtypes` with `project_issue_types` & `project_issue_fields`. This will break applications that use the library and utilize the `createmeta_*` methods.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -1763,6 +1763,7 @@ class JIRA:
             maxResults (int): Maximum number of issues to return.
               Total number of results is available in the ``total`` attribute of the returned :class:`ResultList`.
               If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)
+
         Returns:
             Dict[str, Any]
         """
@@ -1792,6 +1793,7 @@ class JIRA:
             maxResults (int): Maximum number of issues to return.
               Total number of results is available in the ``total`` attribute of the returned :class:`ResultList`.
               If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)
+
         Returns:
             Dict[str, Any]
         """
--- a/jira/client.py
+++ b/jira/client.py
@@ -2126,7 +2126,7 @@ class JIRA:
 
         .. deprecated:: 3.6.0
             Use :func:`project_issue_fields` instead.
-        
+
         This API was introduced in JIRA Server / DC 8.4 as a replacement for the more general purpose API 'createmeta'.
         For details see: https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -1748,6 +1748,62 @@ class JIRA:
                 "Use 'createmeta' instead."
             )
 
+    def createmeta_issuetypes(
+        self,
+        projectIdOrKey: str | int,
+        startAt: int = 0,
+        maxResults: int = 50,
+    ) -> dict[str, Any]:
+        """Get the issue types metadata for a given project, required to create issues.
+        This API was introduced in JIRA Server / DC 8.4 as a replacement for the more general purpose API 'createmeta'.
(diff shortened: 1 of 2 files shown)


adehad:
thanks for this, could you add a deprecation warning to the functions before calling the replacement method.

paminov:
> thanks for this, could you add a deprecation warning to the functions before calling the replacement method.
@adehad  thanks for pointing this out. I've added deprecation warnings to the methods.

jira/client.py
```diff
@@ -2076,6 +2076,76 @@ def _check_createmeta_issuetypes(self) -> None:
...
+        Returns:
+            Dict[str, Any]
+        """
+        warnings.warn(
+            "'createmeta_issuetypes' is deprected and will be removed in future releases."
+            "Use 'project_issue_types' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        self._check_createmeta_issuetypes()
```
adehad, line 2104:
```suggestion
        """Get the issue types metadata for a given project, required to create issues.

        .. deprecated:: 3.6.0
            Use :func:`project_issue_types` instead.

        This API was introduced in JIRA Server / DC 8.4 as a replacement for the more general purpose API 'createmeta'.
        For details see: https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html

        Args:
            projectIdOrKey (Union[str, int]): id or key of the project for which to get the metadata.
            startAt (int): Index of the first issue to return. (Default: ``0``)
            maxResults (int): Maximum number of issues to return.
              Total number of results is available in the ``total`` attribute of the returned :class:`ResultList`.
              If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)

        Returns:
            Dict[str, Any]
        """
        warnings.warn(
            "'createmeta_issuetypes' is deprecated and will be removed in future releases."
            "Use 'project_issue_types' instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        self._check_createmeta_issuetypes()
```

jira/client.py
```diff
@@ -2076,6 +2076,76 @@ def _check_createmeta_issuetypes(self) -> None:
...
+        Returns:
+            Dict[str, Any]
+        """
+        warnings.warn(
+            "'createmeta_fieldtypes' is deprected and will be removed in future releases."
+            "Use 'project_issue_fields' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        self._check_createmeta_issuetypes()
```
adehad, line 2140:
```suggestion
        """Get the field metadata for a given project and issue type, required to create issues.

        .. deprecated:: 3.6.0
            Use :func:`project_issue_fields` instead.
        
        This API was introduced in JIRA Server / DC 8.4 as a replacement for the more general purpose API 'createmeta'.
        For details see: https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html

        Args:
            projectIdOrKey (Union[str, int]): id or key of the project for which to get the metadata.
            issueTypeId (Union[str, int]): id of the issue type for which to get the metadata.
            startAt (int): Index of the first issue to return. (Default: ``0``)
            maxResults (int): Maximum number of issues to return.
              Total number of results is available in the ``total`` attribute of the returned :class:`ResultList`.
              If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)

        Returns:
            Dict[str, Any]
        """
        warnings.warn(
            "'createmeta_fieldtypes' is deprecated and will be removed in future releases."
            "Use 'project_issue_fields' instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        self._check_createmeta_issuetypes()
```

adehad (changes requested):

adehad (approved):
thanks again, just did a small fix for the linter and docs to be happiest
image

PR #1845, opened 2024-03-25
Revert "Bump codecov/codecov-action from 3.1.4 to 4.1.0"

adehad:
This reverts commit eb0ec90e08ae24823e266b0128b852022d212982.

Refer to: https://github.com/codecov/codecov-action/issues/1292


adehad:
if we add the codecov token I think we'll be good, but I don't think it would work for PR's, which is kinda annoying so probably best to downgrade?

PR #1847, opened 2024-03-31
Replace pillow dependency with filetype

### LINKED ISSUE #1801: Make dependency on pillow optional
### Bug summary

Please provide a way to install and use your package without `pillow` which is really a nightmare to package in pyinstaller or using shiv. You can use `extra` to have this dependencies optional and try/except the import only when really needed. 

thanks

### Jira Instance type

### Jira instance version

_No response_

### jira-python version

main

3.8

- [ ] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
pillow is hard to package in shiv (zipapp).
```

### Stack trace

```python
n/a
```

### Expected behaviour

jira can be installed without pillow. only to use the image related feature this dependency is needed

### Additional Context

_No response_

cilki:
Fixes #1801

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -105,7 +105,7 @@ parso==0.8.3
     # via jedi
 pickleshare==0.7.5
     # via ipython
-pillow==10.2.0
+filetype==1.2.0
     # via jira (setup.cfg)
 pluggy==1.4.0
     # via pytest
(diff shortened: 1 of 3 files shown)


adehad (commented):
I'm concerned by the current maintained state of `filetype` but it appears that for images at least we should be fine. I am concerned that we don't have a test for this method yet, so if we can add one in that would give me confidence in using this package

dimitarOnGithub:
@adehad just in case you've missed it, there's some general conversation about the use of this functionality going on in #1867 around this, seems like that was the PR that got traction, even if it was filed after this one.

PR #1861, opened 2024-05-22
Provide overload type hints for search_issues variants

### LINKED ISSUE #1608: `client#search_issues`: `Incompatible types in assignment (expression has type "Union[str, Issue]", variable has type "Issue")`
### Bug summary

Using

```py
issues = jira.search_issues(query, maxResults=False)
```

Gives the aforementioned result:

```py
 error: Incompatible types in assignment (expression has type "Union[str, Issue]", variable has type "Issue")  [assignment]
```

The culprit seems to be:

```
Returns:
            Union[Dict,ResultList]: Dict if ``json_result=True``
```

which should be quickly solveable with a little bit of https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading

### Jira Instance type

### Jira instance version

_No response_

### jira-python version

3.4.1

3.9

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
jira: JIRA
jira.the_function(x)
# 3.
...
```

### Stack trace

```python
-
```

### Expected behaviour

No typing issues

### Additional Context

_No response_

Stealthii:
search_issues can return either a dictionary (if json_result) or a list of Issues under default behavior.

This change provides separate hints using overload so that type checked code doesn't need to cast unnecessarily when dealing with output.

Tests have been added, and cases where casting was used have been removed.

Fixes #1608.

### DIFF
--- a/examples/auth.py
+++ b/examples/auth.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 from collections import Counter
-from typing import cast
 
 from jira import JIRA
 from jira.client import ResultList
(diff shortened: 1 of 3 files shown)


Stealthii:
The [cloud / py3.8](https://github.com/pycontribs/jira/actions/runs/9193983041/job/25287168593?pr=1861#logs) job seems to be failing for all PRs, unrelated to the changes provided here.

filbranden:
Looks awesome! Would be nice if this was merged, I run into issues having to cast this every time I use it

adehad (approved):
Thanks for this

Stealthii:
Thank you for getting this merged!

PR #1863, opened 2024-05-25
adding warning when requests_futures isnt installed

RichieCahill:
This is a super small pr. I dont think it requires any new tests because of its nature.
Please let me know if I did anything wrong.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -798,6 +798,11 @@ class JIRA:
 
                 async_class = FuturesSession
             except ImportError:
+                msg = (
+                    "async option requires requests-futures to be installed. "
+                    "falling back to synchronous implementation."
+                )
+                warnings.warn(msg)
                 pass
             async_workers = self._options.get("async_workers")
 


dimitarOnGithub:
@adehad / @studioj this one looks good to me as-is, but I wonder if the constructor would be a better place for this warning considering it wouldn't be visible until the code attempts to execute the `_fetch_pages` method? I think it'd be better if the import, along with the assignment of the `async_workers` and `async_class`, happens in the constructor itself, that way you get an instant warning that a requirement is missing.

jira/client.py
```diff
@@ -573,7 +573,22 @@ def __init__(
...
+                    from requests_futures.sessions import FuturesSession
+
+                    options["async_class"] = FuturesSession
+                except ImportError:
+                    msg = (
+                        "async option requires requests-futures to be installed. "
+                        "falling back to synchronous implementation."
+                    )
+                    warnings.warn(msg)
+                    pass
```
adehad, line 590: My preference would be to:

1. do the import check outside the scope of this class, then either set `FutureSession` as `None` if it is not imported or have a boolean variable, eg. `_has_async_support`. 
2. This warning can be emitted in init as you have done if the async option is passed incorrectly. 
Furthermore I would augment this warning message with a suggestion of how to install this extra requirement, as we do have this as an extra, it may be appropriate to say, `this can be installed using the 'async' extra, e.g. pip install jira[async]`
3. I would be tempted to modify the option itself to be False and then no further code changes need to be made in the dependency functions.
4. Code that needs the async may check for the boolean or that the imported class is not None.

jira/client.py
```diff
@@ -847,14 +852,16 @@ def json_params() -> dict[str, Any]:
...
                 if (
                     async_class is not None
                     and not is_last
                     and (total is not None and len(items) < total)
                 ):
                     async_fetches = []
                     future_session = async_class(
-                        session=self._session, max_workers=async_workers
+                        session=self._session,
+                        max_workers=self._options["async_workers"],
```
adehad, line 864: Just to preserve the original code I would suggest keep this as

```suggestion
                        max_workers=self._options.get("async_workers"),
```

adehad (changes requested):
Thanks for the initial review @dimitarOnGithub and for the PR @RichieCahill. I think we should tackle this in a more obvious way if possible, I've added some suggestions in the review

jira/client.py
RichieCahill: what value dose preserve the original code add in this case?

jira/client.py
RichieCahill: ive updated the code based on your suggestion. Im unclear on what 3 means.

RichieCahill:
should we insted of warning if async=true and requests_futures isnt installed just raise an exception

RichieCahill:
@adehad please take a look at this pr i have updated parts of it

PR #1864, opened 2024-05-28
fix: Use field translation in Jira.issue method

Wojtini:
Fields param in issue are untranslated so if you use custom_field's name instead of id Jira API won't return it. Those changes are supposed to fix that.

Changes:
1. Extracted translation part to seperate methods
2. Changed variable names to be less abraviated
3. Reused those methods in Issue method
4. Missing tests for limiting fields on Issue method
5. Seperate tests for translation (Added duedate field to test issue)


jira/client.py
```diff
@@ -3564,16 +3561,63 @@ def search_issues(
...
+        issues: list[Issue], translation_map: dict[str, str]
+    ) -> None:
+        """Untranslate issues fields names from JIRA's REST API field id.
+
+        Args:
+            issues (list[Issue]): list of issues to be untranslated
+            translation_map (dict[str, str]): translation dict containing original field name and field id
+        """
+        if not translation_map:
+            return
```
adehad, line 3612: do we want to move this out of a staticmethod and instead have a cache of the the translation_map inside the class instance? ie similar to what is done with the `self._field_cache` ?

PR #1870, opened 2024-07-01
remove dead code in _fetch_pages

RichieCahill:


adehad (approved):
@studioj I remember looking at this during our first fixing of the pipeline and wondering why it existed, I think there should be no issue removing

PR #1871, opened 2024-07-01
removing BountySource from readme

RichieCahill:


adehad (approved):
Thanks for this, for any future readers, some discussions in their repo: https://github.com/bountysource/core/issues/1586

Issue #1855, opened 2024-04-03
add_watcher should accept a user ID, as per Jira API


supermitch:
### Problem trying to solve

I noticed that in version 3.5.x, the new Jira cloud (and this package) required a user ID to add watchers.
See https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-watchers/#api-rest-api-3-issue-issueidorkey-watchers-post

However, in https://github.com/pycontribs/jira/pull/1682, someone added the (very useful) ability to once again add watchers by email addresses. However, they:

1. Added an extra API call
2. Broke consistency with the API docs

### Possible solution(s)

Please update the method and documentation so we can pass either an email address or an `accountId`, if we already have it.

### Alternatives

- Alternatively, I would revert the functionality to be more in-line with the REST API.
- Alternatively, update the documentation to be more explicit, stating that a user search is being conducted internally.

### Additional Context

Source: https://jira.readthedocs.io/_modules/jira/client.html#JIRA.add_watcher

dimitarOnGithub:
I agree with the sentiment here, and also described that the internal user lookup might be resulting in unexpected behaviour in [this comment](https://github.com/pycontribs/jira/pull/1301#issuecomment-2191864680) on #1301.

@adehad I'll be happy to look into this since I believe it'd be a good way to solve both this issue, as well as #1284.

adehad:
@dimitarOnGithub happy for you to look further into a solution, thanks for this

Issue #1284, opened 2022-01-27
Cannot properly assign to person


ZionDeng:
### Bug summary

I am using ` jira==3.0.1` python package, but I cannot properly assign my created ticket to the person I want. 
My server is: https://issues.teslamotors.com/
This is the ticket and the person I want to assign
![assign](https://user-images.githubusercontent.com/47084312/151302232-a43a4f0b-4282-4553-b8eb-575077d4dd41.png)
But when I am trying to assign this ticket to him, I got the error
![error](https://user-images.githubusercontent.com/47084312/151302613-5ca27bbf-a3de-4e0f-9389-42879d0b90ff.png)

I am sure that the assignee name is xins, but jira wants to assignee to xinhe

Similar issue goes to `soyou`， `jian`， `xins`, but `leidxu`, `lifachen`, `fjin`, `elliu` can be properly assigned.

I think there are some bug in the assign logic, the add watcher function has no similar bugs. I can properly add watchers. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

3.0.1

### Python Interpreter version

3.8

### Which operating systems have you used?

- [ ] Linux
- [ ] macOS
- [X] Windows

### Reproduction steps

```python
jira : JIRA = JIRA('https://issues.teslamotors.com/', {user}, {password})
issue: Issue = jira.create_issue(
    project = "ACE",
    summary = "VVV.EEE TEST Jira Ticket #COA",
    description = "TEST Jira Ticket for Vendor VVV engineer EEE",
    issuetype = {"name": "Task"},
    components = task["components"]
    components = [],
)
jira.assign_issue(issue.id, 'xins')
```

### Stack trace

```python
![error](https://user-images.githubusercontent.com/47084312/151302613-5ca27bbf-a3de-4e0f-9389-42879d0b90ff.png)
```

### Expected behaviour

Properly assign to the right person
![assign](https://user-images.githubusercontent.com/47084312/

GabitovRustam:
To fix this issue need to change 
line [1780 in client.py](https://github.com/pycontribs/jira/blob/main/jira/client.py#L1780)
 `       payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}`
to lines 
```
        user_key = self._get_user_key(assignee)
        payload = {"accountId": user_id} if self._is_cloud else {"name": user_key}
```
[...]

PR #1874, opened 2024-07-12
Remove internal user lookup

dimitarOnGithub:
This PR removes the internal user lookup functionality performed by the `assign_issue`, `add_watcher` and `remove_watcher` methods.

By doing so, issues reported in #1855 and #1284 should be resolved. Additionally, this fixes an actual bug that the internal lookup logic introduced, namely the fact that while `_get_user_id` method would work for 100% exact usernames, it'd still return the first result if it fails to do so, as seen below:

### Cloud
```python
>>> jira.search_users(query='dummy')
[<JIRA User: displayName='dummy', accountId='...5c6e8'>, <JIRA User: displayName='dummy', accountId='...39292'>]
>>> jira._get_user_id('dummy')
'...5c6e8'
```

### Server
```python
>>> jira.search_users('bot')
[<JIRA User: displayName='CDE Bot', key='cde.bot', name='cde.bot'>, <JIRA User: displayName='Ops Bot', key='opsbot', name='opsbot'>, <JIRA User: displayName='Dev CO Bot', key='devco.bot', name='devco.bot'>]
>>> jira._get_user_id('bot')
'cde.bot'
```

It also makes the `_get_user_identifier` method public and provides the ability to pass a `User` resource object to the methods.

Going forward, users will have to provide the exact username (for hosted) or accountId (for cloud) in order to perform any of those operations.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -185,9 +185,15 @@ If you only want a few specific fields, save time by asking for them explicitly:
     issue = jira.issue('JRA-1330', fields='summary,comment')
 
 Reassign an issue::
+    Reassigning an issue can be done either by providing the username/accountId (hosted/cloud) or the User Resource object itself::
 
-    # requires issue assign permission, which is different from issue editing permission!
-    jira.assign_issue(issue, 'newassignee')
+        # requires issue assign permission, which is different from issue editing permission!
+        # via username, for hosted instances
+        jira.assign_issue(issue, 'newassignee')
+        # or accountId, for cloud instances
+        jira.assign_issue(issue, 'gweg3:5fr23r23r-3041-4342-23f2-2g3g232c2e1234:8008sdg3-441a-12f2-9sg1-erbwer3q3r3')
+        # or via the User retrieved from search_users()
+        jira.assign_issue(issue, user_resource)
 
 If you want to unassign it again, just do::
 
@@ -421,15 +427,17 @@ Watchers are objects, represented by :class:`jira.resources.Watchers`::
         # watcher is instance of jira.resources.User:
         print(watcher.emailAddress)
 
-You can add users to watchers by their name::
+You can add users to watchers by their name (hosted) / accountId (cloud) or the User resource itself::
 
     jira.add_watcher(issue, 'username')
     jira.add_watcher(issue, user_resource.name)
+    jira.add_watcher(issue, user_resource)
 
 And of course you can remove users from watcher::
 
     jira.remove_watcher(issue, 'username')
     jira.remove_watcher(issue, user_resource.name)
+    jira.remove_watcher(issue, user_resource)
 
 Attachments
 -----------
--- a/jira/client.py
+++ b/jira/client.py
@@ -2263,19 +2263,21 @@ class JIRA:
 
     # non-resource
     @translate_resource_args
-    def assign_issue(self, issue: int | str, assignee: str | None) -> bool:
+    def assign_issue(self, issue: int | str, assignee: str | None | User) -> bool:
         """Assign an issue to a user.
 
         Args:
             issue (Union[int, str]): the issue ID or key to assign
-            assignee (str): the user to assign the issue to. None will set it to unassigned. -1 will set it to Automatic.
+            assignee (Union[str, User]): username (for hosted) or account ID (for cloud) of the user to add to the
+                watchers list. Alternatively, you can provide the User object itself.
 
         Returns:
             bool
         """
         url = self._get_latest_url(f"issue/{issue}/assignee")
-        user_id = self._get_user_id(assignee)
-        payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}
+        if isinstance(assignee, User):
+            assignee = self.get_user_identifier(assignee)
+        payload = {"accountId": assignee} if self._is_cloud else {"name": assignee}
         self._session.put(url, data=json.dumps(payload))
         return True
 
--- a/jira/client.py
+++ b/jira/client.py
@@ -2209,7 +2209,7 @@ class JIRA:
             params["expand"] = expand
         return self._get_json("issue/createmeta", params)
 
-    def _get_user_identifier(self, user: User) -> str:
+    def get_user_identifier(self, user: User) -> str:
         """Get the unique identifier depending on the deployment type.
 
         - Cloud: 'accountId'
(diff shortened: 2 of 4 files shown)


dimitarOnGithub:
@adehad this is the PR for what I mentioned on #1855. I did some manual tests for both server and cloud on the instances I have access to, but I didn't quite manage to set up my local environment for the automated tests, I need to spend some more time on that as it seems harder than the JIRA library itself haha

jira/client.py
```diff
@@ -2223,59 +2223,23 @@ def _get_user_identifier(self, user: User) -> str:
...
     @translate_resource_args
-    def assign_issue(self, issue: int | str, assignee: str | None) -> bool:
+    def assign_issue(self, issue: int | str, assignee: str | None | User) -> bool:
         """Assign an issue to a user.
 
         Args:
             issue (Union[int, str]): the issue ID or key to assign
-            assignee (str): the user to assign the issue to. None will set it to unassigned. -1 will set it to Automatic.
+            assignee (Union[str, User]): username (for hosted) or account ID (for cloud) of the user to add to the
+                watchers list. Alternatively, you can provide the User object itself.
```
adehad, line 2234: we should keep the detail regarding `-1` and `None`

jira/client.py
```diff
@@ -2719,36 +2683,37 @@ def watchers(self, issue: str | int) -> Watchers:
...
 
     @translate_resource_args
-    def add_watcher(self, issue: str | int, watcher: str) -> Response:
+    def add_watcher(self, issue: str | int, watcher: str | User) -> Response:
         """Add a user to an issue's watchers list.
 
         Args:
             issue (Union[str, int]): ID or key of the issue affected
-            watcher (str): name of the user to add to the watchers list
+            watcher (str): username (for hosted) or account ID (for cloud) of the user to add to the watchers list
```
adehad, line 2691:
```suggestion
            watcher (str | User): username (for hosted) or account ID (for cloud) of the user to add to the watchers list
```

adehad (changes requested):
In general I like the idea of the merge request and it is the direction I want the library to go moving forward, I don't want to be passing strings or integers around, I much rather they be objects.

I think what we are missing in the orignal implementation was checking if the entered str/int was a valid user. In a sense could we use `JIRA.user(id_or_accountId)` to first check if the user provided a valid complete ID, if so then use that, otherwise fallback to the search and maybe add a warning to use a more specific search

jira/client.py
```diff
@@ -2223,59 +2223,23 @@ def _get_user_identifier(self, user: User) -> str:
         """
         return user.accountId if self._is_cloud else user.name
 
-    def _get_user_id(self, user: str | None) -> str | None:
```
adehad, line 2226: a benefit of keeping this as a method is that it allows users of the library to override its behaviour should they not like its implementation. I think it would be best to keep it in, but have it also accept `User` objects and additionally also first check if the `JIRA.user()` method returns a valid user before attempting the search with a warning to use the `User` object or a specific string

jira/client.py
```diff
@@ -2223,59 +2223,23 @@ def _get_user_identifier(self, user: User) -> str:
...
 
         Returns:
             bool
         """
         url = self._get_latest_url(f"issue/{issue}/assignee")
-        user_id = self._get_user_id(assignee)
-        payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}
+        if isinstance(assignee, User):
+            assignee = self.get_user_identifier(assignee)
+        payload = {"accountId": assignee} if self._is_cloud else {"name": assignee}
```
adehad, line 2242:
```suggestion
        if isinstance(assignee, User):
            user_id = self.get_user_identifier(assignee)
        payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}
```

(non-blocking) I generally have a preference to avoid overriding function arguments values inside the function. The above is an example of what I would prefer

adehad: also applies to `watcher` below

docs/examples.rst
```diff
@@ -185,9 +185,15 @@ If you only want a few specific fields, save time by asking for them explicitly:
     issue = jira.issue('JRA-1330', fields='summary,comment')
 
 Reassign an issue::
+    Reassigning an issue can be done either by providing the username/accountId (hosted/cloud) or the User Resource object itself::
```
adehad, line 188: image

Remove the `::` after `Reassign an issue`

and also there should be no indentation for the first line, and therefore codeblock underneath also should get indented back

```suggestion
Reassigning an issue can be done either by providing the username/accountId (hosted/cloud) or the User Resource object itself::
```

dimitarOnGithub:
Sorry about the mess with the docs, for some reason my local working copy was showing the document entirely borked, hence why I did the silly thing with the indentations. b0b552ec52ba83ab621f0c90e46190fb067eeb8c should fix that.

jira/client.py
dimitarOnGithub: I'm not opposed to that idea, the only reason why I opted for value overriding is because `assignee` might also be a string or None and in that case it's being passed directly into the payload dict as-is. I think in order to maintain the value of the variable untouched, the code would have to change to something like:
```suggestion
        if isinstance(assignee, User):
            user_id = self.get_user_identifier(assignee)
        else:
            user_id = assignee
        payload = {"accountId": user_id} if self._is_cloud else {"name": user_id}
```
If that's okay with you, I'll go ahead and change it, I just figured the current design might be a bit tidier, but I'm not dead set on it.

jira/client.py
dimitarOnGithub: On one hand, I agree with that (what's said both here and https://github.com/pycontribs/jira/pull/1874#pullrequestreview-2237715012), it is a nice QoL feature to have, however on the other, I think it doesn't solve the problem in #1284 and it also is not consistent with the JIRA API, as pointed out in #1855; 

In my eyes, the benefit of removing the internal lookup entirely is that:

- it's consistent with the JIRA API, which returns a 4xx error code if no user is found with the provided data - the library will raise a `JIRAError` for the user to handle;
- it solves the ambiguity of the internal search; I think that the operations (add/remove watchers & assign issue) should be clear cut - either the data provided is either valid and the operation succeeds, or it is not and it will raise an error. Having the library do an internal search takes the decision of how to proceed away from the user.

An elegant solution that offers the best of both worlds would be, IMO, adding an additional boolean flag to each function that allows the users to disable the search if they'd like, ie:
```python
    def assign_issue(self,
                     issue: int | str,
                     assignee: str | None | User,
                     with_lookup: bool = False) -> bool:
        
```

jira/client.py
adehad: this is a good point, let's leave it as is for now

adehad (changes requested):
just something I noticed that would be pretty critical to his working is removing this decorator that is trying to be helpful but removes the type safety unfortunately.

I thought we had a code coverage action but it isn't seeing the stats from this PR, so I couldn't exactly confirm it wasn't reaching the new code branch

jira/client.py
dimitarOnGithub: I love that idea, will get around to looking into this by the end of the week hopefully!

jira/client.py
```diff
@@ -2223,59 +2223,24 @@ def _get_user_identifier(self, user: User) -> str:
...
-            if len(users) > 1:
-                matches = [u for u in users if self._get_user_identifier(u) == user]
-            user_obj = matches[0] if matches else users[0]
-
-        except Exception as e:
-            raise JIRAError(str(e))
-        return self._get_user_identifier(user_obj)
-
     # non-resource
     @translate_resource_args
```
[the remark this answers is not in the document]
dimitarOnGithub: I don't think the decorator touches on the `User` objects, it seems to only manipulate the `Issue` in this case

dimitarOnGithub:
@adehad sorry for the delay with this, here's the updated functionality with having a `with_lookup` flag in the JIRA's init.

Not sure about the linter errors in the automatic checks, the value provided is a `str` which is one of the expected variable types.

Issue #1875, opened 2024-07-16
Erroneous Warnings When Creating Issue Links


ssamsel-rtx:
### Bug summary

When creating issue links, I get a warning log informing me that the issue link type I specify is not present in the list of link types, even if it is in fact present. This is due to the equality of the `IssueLinkType` class and strings not working. Even when passing in the type argument as the appropriate `IssueLinkType`, this warning still occurs as the `translate_resource_args` decorator converts it to a string. This can be fixed by changing the line `issue_link_types = self.issue_link_types()` to `issue_link_types = list(map(lambda x: x.name, self.issue_link_types()))` in the `JIRA.create_issue_link` method in `client.py`

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### jira-python version

main

### Python Interpreter version

3.10.12

### Which operating systems have you used?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance
jira: JIRA
# 2. When I create an "is closely related to" issue link:
jira.create_issue_link("Closely Related To", "KEY-1", "KEY-2")
# 3. I get the erroneous warning
Warning: Specified issue link type is not present in the list of link types
```

### Stack trace

```python
n/a
```

### Expected behaviour

I expect no warning, as the issue link is created correctly, and is in fact in the set of issue links returned by `jira.issue_link_types()`

Lenormju:
I have the same issue.

KiloFoxtrotPapa:
This fix won't work. You'll break the for loop later when it tries to access `.outward` and `.inward`
Fix should probably go on the `if` line?
```
        if type not in [lt.name for lt in issue_link_types]:
```

Ket3r:
Same issue here. I even tried fetching the link types instead of using a string:

```py
def _create_issue_links(issue_key, issue):
    created_issue = get_jira().issue(issue_key)
    # Get link types from jira
    link_types = get_jira().issue_link_types()
    verification_link_type = next(link for link in link_types if link.name == "Verification")

    for target in issue.link_verifies_keys:
        get_jira().create_issue_link(
            verification_link_type, inwardIssue=created_issue.key, outwardIssue=target
        )
```

But this still results in this warning log:

```
WARNING  jira:client.py:2935 Warning: Specified issue link type is not present in the list of link types
```

adoniadi-stm:
[...]

PR #1876, opened 2024-07-17
jira/client.py: Map IssueLinkType array to string array

ssamsel-rtx:
Equality check between IssueLinkType and string always returns false. Function decorator casts IssueLinkType parameters to string, so a warning is always logged. This fixes that issue.

Closes: #1875

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2929,7 +2929,7 @@ class JIRA:
             Response
         """
         # let's see if we have the right issue link 'type' and fix it if needed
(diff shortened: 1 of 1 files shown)


adehad (approved):
Excellent, thanks for this. I want to have a think of a test case to make sure we don't cause regressions on this moving forward, so will have a think before merging

kohtala:
An IssueLinkType as parameter is supposed to be working according to the parameter typing, but it won't work with this patch. I created another PR #2406.

PR #1905, opened 2024-10-16
Add additional parameters for add_worklog()

LevBaranov:
The Jira Docs has method [add worklog](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-worklogs/#api-rest-api-3-issue-issueidorkey-worklog-post) with additional attributes in the request body. But this lib doesn't support this feature.

I think use `**kwargs` is a good idea to solve the problem.


LevBaranov:
@adehad, hi!

I don't know how to fix the errors. I think my changes are small and can be merged without problems. How can I help to make it faster?

Issue #1904, opened 2024-10-02
TokenAuth can stop working if session cookie becomes invalid


gmishkin:
### Bug summary

The Jira client initially works with token auth but eventually begins failing with a 401. A key part of the error response is

>os_authType was 'any' and an invalid cookie was sent.

We ruled out any situation on our end where there might be multiple web servers involved setting separate session cookies and a load balancer not setting sticky session. There is only on web server so any session cookies should still be valid.

So we contacted Atlassian support and were given the suggestion: _Disable cookies_ on outgoing requests: those are not necessary. The `Authorization` header is enough for authenticate REST requests.

>There are two mechanisms involved here:
> 
> - The Personal Access Token (PAT) that is used for authentication of your integration's requests to Jira.
> - The session cookie `JSESSIONID` that is created when a user session is established after the user has been authenticated. Every session has a `JSESSIONID` cookie regardless of authentication method. When that user session is closed (or times out) the session ID is removed from the active sessions list and any attempts to use the same session id from that on will be rejected.
> 
>So when you use the PAT to authenticate the normal flow is like this:
> 
>1. Jira receives the request with the PAT and checks its validity to either authenticate the request or reject it.
>2. If the PAT authentication was successful, a session is established and a `JSESSIONID` cookie with a session id is sent back in Jira's response.
> 
>As you then have already established a session, you can then use the `JSESSIONID` cookie in subsequent requests to hook on to the same session without having to authenticate every single request. This is the same as for a user request in the UI.
>If you, on the other hand, send a stored `JSESSIONID` cookie with the _first_ request to the REST API, it will be interpreted as that you want to hook on to a previously established session.
> 
>This session handling is taking place in the Tomcat web server, i.e. in the step _before_ Jira, so the `JSESSIONID` will be checked first and if that given session id is not in the active sessions list (maintained by Tomcat) it will reject it.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

9.16.1

### jira-python version

3.8.0

### Python Interpreter version

3.12.1

### Which operating systems have you used?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. Given a Jira client instance, which initially works
        jira_client: JIRA = JIRA(
            XXXX,
            token_auth=os.environ["JIRA_TOKEN"],
        )
# 2. Within a few minutes time period, when I call the function with argument x
[...]

PR #1910, opened 2024-10-28
refactor 401 retrying into both auth methods

gmishkin:
See #1904 for details on why.

This change refactors 401 retrying into the common base class, and for token auth the retry is implemented by clearing the session cookie and retrying with the same PAT as recommended.


gmishkin:
@ssbarnea not sure what to do about the check failures, can you advise?

Re4zOon:
Hi @adehad!

Can you please review this PR?

Thank you!

Re4zOon:
Hi,

This is now impacting multiple workflows on our side.
Can you please merge this?
Seems like a simple change (compared of my PRs :D ).

jira/client.py
```diff
@@ -310,53 +310,35 @@ def _sort_and_quote_values(self, values):
...
+    def __init__(self, session: ResilientSession | None = None):
         self._session = session
-        self._session_api_url = session_api_url  # e.g ."/rest/auth/1/session"
-        self.__auth = auth
         self._retry_counter_401 = 0
         self._max_allowed_401_retries = 1  # 401 aren't recoverable with retries really
 
+    def init_session(self):
+        """Auth mechanism specific code to re-initialize the Jira session."""
+        raise NotImplementedError()
```
adehad, line 323:
```suggestion
        raise NotImplementedError()
```

let's make this an abstract base class and define this as an abstract method. To allow us to pick up errors in static analysis rather than runtime.

jira/client.py
```diff
@@ -310,53 +310,35 @@ def _sort_and_quote_values(self, values):
...
-        self, session: ResilientSession, session_api_url: str, auth: tuple[str, str]
-    ):
-        """Cookie Based Authentication.
-
-        Args:
-            session (ResilientSession): The Session object to communicate with the API.
-            session_api_url (str): The session api url to use.
-            auth (Tuple[str, str]): The username, password tuple.
-        """
+    def __init__(self, session: ResilientSession | None = None):
```
adehad, line 316: Let's remove the default of `None` here, and then remove as many if not all the plain asserts.

If there do end up being situations where the session is None when passing to the constructor of the subclasses then let us move to a subclass `__init__` method override, and rather than asserts, either use warnings, or if it should really fail, let's use an Exception with suitable message.
I want to avoid `asserts` in the production codebase.

adehad (changes requested):

gmishkin:
@adehad I agree with both of your requested changes but just wanted to give you a heads up that we have our internal Jira 10 upgrade planned for next month, and if you look at the linked issue, it appears that someone found an [upstream issue](https://jira.atlassian.com/browse/JRASERVER-76340) that matches the symptoms I was seeing here, that's fixed in Jira 10.

So, I may not get back to this PR, and honestly, you would be forgiven for closing this PR outright, since it's just a workaround for a bug that's resolved in the latest supported major version of the non-cloud Jira software (I don't know what your support policy is for older versions of the Jira server).

adehad:
Thanks @gmishkin for this update, I didn't catch that.

Sorry for the delay in getting this merged when it would have made the most impact to you. 

I think there is still value in keeping it open and even merging, as I can imagine there are individuals affected that won't get the change to migrate as soon. Happy to leave that decision to you or any other budding contributors and happy to facilitate.
(Based on this it may also be helpful to bring attention to that information in the code to help someone understand why we might have gone down this route, maybe pointing to the upstream issue / this PR)

I've just come back to open source contributions after a long pause, but now that I have time available I should be able to accelerate merging if it is ready.

gmishkin:
@adehad I don't know if this is a thing that exists on GitHub, but if you're able to you should "pin" [this comment](https://github.com/pycontribs/jira/issues/1904#issuecomment-2730455264)

Adding a comment to the code describing what it's working around makes sense too 👍

Issue #1911, opened 2024-10-28
worklog does not support expand parameter


MarcelWilson:
### Problem trying to solve

Jira API [Get Worklog](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-worklogs/#api-rest-api-2-issue-issueidorkey-worklog-id-get)  supports requesting additional worklog properties with a parameter `expand` which accepts the value `properties`.  Any additional worklog properties set will then be returned.

It would be nice if the jira client request `jira.worklog()` supported the use of the expand parameter.

### Possible solution(s)

```python
jira = JIRA('https://jira.atlassian.com')

worklog = jira.worklog(ticket, worklogid, expand="properties")
value = worklog.properties[0].value.my_unique_property_key
```

PR #1912, opened 2024-10-28
Adding support for expand parameter to worklog fetches

bandophahita:
Addresses #1911

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -2770,17 +2770,18 @@ class JIRA:
         return worklogs
 
     @translate_resource_args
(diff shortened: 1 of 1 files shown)


bandophahita:
I'm not sure what I can do about the errors.  I can't add a label to the PR to get the ack step to pass.
and I'm not sure what I can do about tox/check error:
`Codecov: Failed to get OIDC token with url: https://codecov.io./ Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable`

PR #1918, opened 2024-11-08
fix: Wrong tox env for packaging after rename in 39d89ef2

wulmer:
Hi, this PR fixes the wrong tox env name (which was not renamed with the rest of 39d89ef2) and which led to the latest release packaging failures, e.g. https://github.com/pycontribs/jira/actions/runs/11516626042


adehad:
Thank you for this contribution, merged as per #1924 where the access tokens are set correctly to test.

PR #1920, opened 2024-11-08
Fix partial update for sprint should use POST

marcotollini:
Partial updates should use POST instead of PUT:
https://docs.atlassian.com/jira-software/REST/9.17.0/#agile/1.0/sprint-partiallyUpdateSprint

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -5391,7 +5391,7 @@ class JIRA:
             payload["goal"] = goal
 
         url = self._get_url(f"sprint/{id}", base=self.AGILE_BASE_URL)
-        r = self._session.put(url, data=json.dumps(payload))
+        r = self._session.post(url, data=json.dumps(payload))
 
         return json_loads(r)
 


studioj:
did you encounter issues? just wonderingas our tests dont really show failures with putting

marcotollini:
@studioj , yes, with PUT Jira Data Center is expecting you to define all the fields. Instead, with POST I can only define the fields that I intend to change

studioj:
@marcotollini can you then add a test so that the two behaviors become apparent?

Thank you

adehad (commented):
https://developer.atlassian.com/cloud/jira/software/rest/api-group-sprint/#api-rest-agile-1-0-sprint-sprintid-put

As documented here, `POST` is correct for a partially updated Sprint, where if the value is not set then the fields are not updated.

If someone wanted to unset the value it would be interesting. Perhaps, we can add a additional argument called `partial` which is default `False` (to preserve original behaviour), but can be set to `True` to do the true `partial` update.

jira/client.py
```diff
@@ -5391,7 +5391,7 @@ def update_sprint(
             payload["goal"] = goal
 
         url = self._get_url(f"sprint/{id}", base=self.AGILE_BASE_URL)
-        r = self._session.put(url, data=json.dumps(payload))
+        r = self._session.post(url, data=json.dumps(payload))
```
adehad, line 5394:
```suggestion
        method = self._session.post if partial else self._session.put
        r = method(url, data=json.dumps(payload))
```

(with corresponding addition of `partial` to args

PR #1921, opened 2024-11-08
Feat support autoStartStop in sprint create and update

marcotollini:
Support autoStartStop and incompleteIssuesDestinationId in create_sprint and update_sprint.
Documentation:
https://developer.atlassian.com/server/jira/platform/rest/v10000/api-group-sprint/#api-agile-1-0-sprint-post


jira/client.py
```diff
@@ -5374,11 +5376,13 @@ def update_sprint(
...
             state: (Optional[str]): The state of the sprint
             goal: (Optional[str]): The goal of the sprint
+            autoStartStop: (Optional[bool]): Start and complete sprint automatically
+            incompleteIssuesDestinationId: (Optional[int]): After sprint completes, move open issues to this sprint id, -1 for backlog
 
         Returns:
             Dict[str, Any]
         """
-        payload = {}
+        payload : dict[str, Any] = {}
```
marcotollini, line 5385: mypy autoinferms dict[str, str], and it would not be compatible with bool nor int

PR #1923, opened 2024-11-14
fix: fix release by using proper toxenv

mojicae:
Noticed that pkg is what probably is needed as that is what is defined in tox.ini, whereas packaging is not defined which is why the release workflow is currently failing


adehad:
Thank you for this contribution, merged as per #1924 where the access tokens are set correctly to test.

Issue #1913, opened 2024-10-30
Dont replace jira url


dimanoga:
### Bug summary

Hi there,

In our company, API access is provided through a white list system.
To "fix" IP addresses, we use Nginx as a proxy server for accessing Jira.
However, we have noticed that when we use the connection.issuel), in the dictresource method, the URL is replaced with one that Jira returns, rather than the one from which the request originated. Additionally, when updating issues using the issue.update method, the URL used is not the one that the authorization was initiated from, but rather the one provided by the Jira server.

Due to this, when attempting to update a task, we receive a 401 unauthorized error. We are hoping to resolve this issue and ensure that the correct URL is used for all API requests.

Could you please assist us in resolving this issue? We would greatly appreciate any assistance you can provide. Thank you in advance for your attention.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Jira Instance type

Jira Server or Data Center (Self-hosted)

### Jira instance version

8.20.0

### jira-python version

3.5.2

### Python Interpreter version

3.11

### Which operating systems have you used?

- [X] Linux
- [X] macOS
- [X] Windows

### Reproduction steps

```python
#1 connect to jira via nginx url
connection = JIRA(options={'server':"https://my-nginx.url/', 'verify': False}, auth=(login, pass))
#2 get issue
issue = connection.issue(key='JIRA-1')
#3 try to update issue
issue.update(fields={'description':'test'}) # the url is already used here = jira.company.com
```

### Stack trace

```python
File "/PycharmProjects/portal_back/venv/lib/python3.11/site-packages/jira/resources•py", Line 852, in update
super (). update(async_=async_, jira-jira, notify=notify, fields=data)
File "/PycharmProjects/portal_back/venv/lib/python3.11/site-packages/jira/resources-py", Line 328, in update
r = self._session.put (self.self + querystring, data-json.dumps (data))
Fetur se. euest Por, urt, vtadta, uergs) se-packages/requests/sesstons.py". Line 649, in put.
File "/PycharmProjects/portal_back/venv/lib/python3.11/site-packages/jira/resilientsession-py", line 247, in request elif raise_on_error (response, **processed_kwargs):
File "/PycharmProjects/portal_back/venv/lib/python3.11/site-packages/jira/resilientsession-py", Line 72, in raise_on_error
raise JIRAError(
jira.exceptions.JIRAError: JiraError HTTP 401 url: https://jira.company.com/rest/api/2/issue/15832118
[...]

PR #1925, opened 2024-11-22
Handle when jira is behind a proxy

skiedude:
Fixes https://github.com/pycontribs/jira/issues/1913

When your JIRA is behind a proxy, the urls returned in `self.self` on a Resource use the non-proxied URL which causes certain actions to timeout/fail. 

This new helper function compares the base URL of the returned `self.self` with the configured `server` parameter in the options. If they are different, the server parameter wins.

### DIFF
--- a/jira/resources.py
+++ b/jira/resources.py
@@ -293,17 +293,8 @@ class Resource:
         options.update({"path": path})
         return self._base_url.format(**options)
 
-    def _validate_self_self_url(self):
-        """In the case of a proxy, use the configured base URL
-            and not the one returned from JIRA.
-
-        Args:
-            self (Resource): self
-
-        Returns:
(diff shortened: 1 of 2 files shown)


studioj:
pls check the linting, 
https://github.com/pycontribs/jira/actions/runs/11979873297/job/33402963996?pr=1925#step:7:42

studioj (approved):

studioj:
> @studioj any updates?

Seems @ssbarnea solved quite a bunch of CI issues, I'll rebase your branch and see how it goes. Seems like i'm not allowed to do merges. But I'll rereview and provide some extra input if needed.

studioj (approved):

adehad (approved):

PR #1926, opened 2024-11-22
fix: release.yml

studioj:
based on the error here https://github.com/pycontribs/jira/actions/runs/11980093961/job/33403659776#step:5:22
im running the same system deps as here https://github.com/pycontribs/jira/blob/main/.github/workflows/tox.yml#L56C9-L59C108

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: 3.9
-    
+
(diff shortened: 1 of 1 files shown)


studioj:
@wulmer @mojicae :-) all eyes help in a review

wulmer (approved):
Makes sense, will likely fix the error, however I don't see the connection to the removal of `docker-compose`, but installing `libkrb5-dev` seems ok.

adehad (approved):

PR #1931, opened 2024-12-05
Handle unnecessary api call to /rest/api/2/field when all fields are requested

Abbas707:
- Previously, an unnecessary API call to fetch fields/rest/api/2/field was made in `search_issues` function even when all fields were requested (fields = ["*all"]).

- Introduced an extra check to prevent redundant API calls when the default case for all fields is specified.


Abbas707:
Closing due to no tag found

PR #1933, opened 2024-12-19
Allow users to search issues without retrieving fields

beliaev-maksim:
PR allows users to perform a search without specifying fields. While this is already possible with the current implementation, the service still makes an unnecessary API call to cache fields. This PR resolves that issue by validating the fields cache only when the user explicitly requests fields.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3533,7 +3533,7 @@ class JIRA:
               If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)
             validate_query (bool): True to validate the query. (Default: ``True``)
             fields (Optional[Union[str, List[str]]]): comma-separated string or list of issue fields to include in the results.
-              Default is to include all fields.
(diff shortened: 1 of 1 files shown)


beliaev-maksim:
let me know if you are interested in PR, then I can rebase it

jira/client.py
```diff
@@ -3533,7 +3533,7 @@ def search_issues(
               If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)
             validate_query (bool): True to validate the query. (Default: ``True``)
             fields (Optional[Union[str, List[str]]]): comma-separated string or list of issue fields to include in the results.
-              Default is to include all fields.
+              Default is to include all fields. If you don't require fields, set it to empty string ``''``.
```
studioj, line 3536: search issues has been deprecated and slightly reworked
i'd suggest implementing it at enhanced_search_issues.

PR #1937, opened 2025-01-25
implemented CommentProperty

zerwes:
handle comment properties analogue to the issue properties

example
```python
is_internal_comment = False
try:
    p = jira.comment_property(425201, 'sd.public.comment')
    is_internal_comment = p.value.internal.lower() == 'true'
except JIRAError:
    pass
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+*.swp
 *.bak
 *.egg
 *.egg-info/
--- a/jira/client.py
+++ b/jira/client.py
@@ -55,6 +55,7 @@ from jira.resources import (
     Attachment,
     Board,
     Comment,
+    CommentProperty,
     Component,
     Customer,
     CustomFieldOption,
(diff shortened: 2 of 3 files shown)


adehad (changes requested):
Excellent PR, let's get the pipeline passing and a test in

zerwes:
Hello @adehad 

> I think we can add a test similar to your example in the description in a file like `tests/resources/test_comment.py`

do you mean something like
```diff
diff --git a/tests/resources/test_comment.py b/tests/resources/test_comment.py
index 1dafc84..515c768 100644
--- a/tests/resources/test_comment.py
+++ b/tests/resources/test_comment.py
@@ -9,9 +9,10 @@ class CommentTests(JiraTestCase):
         self.issue_1_key = self.test_manager.project_b_issue1
         self.issue_2_key = self.test_manager.project_b_issue2
         self.issue_3_key = self.test_manager.project_b_issue3
+        self.issue_4_key = self.test_manager.project_b_issue4
 
     def tearDown(self) -> None:
-        for issue in [self.issue_1_key, self.issue_2_key, self.issue_3_key]:
+        for issue in [self.issue_1_key, self.issue_2_key, self.issue_3_key, self.issue_4_key]:
             for comment in self.jira.comments(issue):
                 comment.delete()
 
@@ -80,3 +81,10 @@ class CommentTests(JiraTestCase):
         comment.update(body="updated! without notification", notify=False)
         assert comment.body == "updated! without notification"
         comment.delete()
+
+    def test_comment_property(self):
+        comment = self.jira.add_comment(self.issue_4_key, "comment for property test")
+        comment_prop_data = {'internal': 'true'}
+        self.jira.add_comment_property(comment, 'sd.public.comment', comment_prop_data)
+        assert self.jira.comment_property(comment, 'sd.public.comment').value.internal.lower() == 'true'
+        comment.delete()
```

adehad:
@zerwes Yes exactly something similar to this. I might add an additional assertion with an `isinstance` check

You don't need to create the issue 4 at the test setup level, reusing an existing issue should be fine.
Or also within the test function itself making a new comment is also perfect.

zerwes:
here we go: 37d7247c570ff48610dec6a7911ad61398558735
I preferred to use a own fct. w/ issue and comment to keep things clearly separated ... if you think this is to heavy, let me know, I can integrate it in another comment test.

tests/resources/test_comment.py
```diff
@@ -9,9 +10,10 @@ def setUp(self):
         self.issue_1_key = self.test_manager.project_b_issue1
         self.issue_2_key = self.test_manager.project_b_issue2
         self.issue_3_key = self.test_manager.project_b_issue3
+        self.issue_4_key = self.test_manager.project_b_issue4
 
     def tearDown(self) -> None:
-        for issue in [self.issue_1_key, self.issue_2_key, self.issue_3_key]:
+        for issue in [self.issue_1_key, self.issue_2_key, self.issue_3_key, self.issue_4_key]:
```
adehad, line 16: Please reuse an existing issue for your test.

This issue key does not exist, all issues are created
https://github.com/pycontribs/jira/blob/main/tests%2Fconftest.py#L317

adehad (changes requested):
Excellent test, minor comment regarding the issue used.

tests/resources/test_comment.py
zerwes: sorry @adehad - I considered the issue creation more generic.
I hope f206010 fixes my mistake ...

PR #2327, opened 2025-03-18
GHA Maintenance

ssbarnea:


.github/workflows/tox.yml
```diff
@@ -23,7 +28,7 @@ jobs:
     steps:
       - name: Determine matrix
         id: generate_matrix
-        uses: coactions/dynamic-matrix@v3
+        uses: coactions/dynamic-matrix@v4
```
adehad, line 31: default python version `3.9` here is different from the `3.10` used with release

Additionally as `--notest` is used throughout the `check` build step fails as it finds any coverage reports

dbaxa (approved):

PR #2334, opened 2025-04-07
Allow users to search issues without retrieving fields

beliaev-maksim:
use advanced search instead of https://github.com/pycontribs/jira/pull/1933

PR allows users to perform a search without specifying fields. While this is already possible with the current implementation, the service still makes an unnecessary API call to cache fields. This PR resolves that issue by validating the fields cache only when the user explicitly requests fields.

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3695,7 +3695,7 @@ class JIRA:
               Total number of results is available in the ``total`` attribute of the returned :class:`ResultList`.
               If maxResults evaluates to False, it will try to get all issues in batches. (Default: ``50``)
             fields (Optional[Union[str, List[str]]]): comma-separated string or list of issue fields to include in the results.
-              Default is to include all fields.
+              Default is to include all fields If you don't require fields, set it to empty string ``''``.
             expand (Optional[str]): extra information to fetch inside each resource.
             reconcileIssues (Optional[List[int]]): List of issue IDs to reconcile.
             properties (Optional[str]): extra properties to fetch inside each result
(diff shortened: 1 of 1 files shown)


jira/client.py
```diff
@@ -3705,19 +3705,23 @@ def enhanced_search_issues(
         Returns:
             Union[Dict, ResultList]: JSON Dict if ``json_result=True``, otherwise a `ResultList`.
         """
-        if isinstance(fields, str):
-            fields = fields.split(",")
-        elif fields is None:
+        if fields is None:
```
Siddhant791, line 3708:
```suggestion
        fields = (
    ["*all"] 
    if fields is None 
    else fields.split(",") 
    if isinstance(fields, str) and fields  # Non-empty string check
    else []
)
```

Siddhant791: This is more cleaner

adehad (approved):
Great backwards compatible addition, I'm sure there is a two branch option for the if statement regarding the fields, but clear and well commented for now.

jira/client.py
```diff
@@ -3705,19 +3705,23 @@ def enhanced_search_issues(
...
             fields = ["*all"]
+        elif fields and isinstance(fields, str):
+            fields = fields.split(",")
+        else:
+            # this is required only for mypy validation
+            fields = []
```
adehad, line 3714: (non-blocking) in general for these simple comments that don't put us above the line limit, I would push them onto one line, as these comments can get missed and become out of date when not associated with a line of code.

```suggestion
            fields = []  # help mypy
```

PR #2339, opened 2025-04-16
Fix CI GitHub Action Workflow

adehad:


.github/workflows/jira_ci.yml
```diff
@@ -10,23 +11,17 @@ on:
 
 jobs:
   server:
+    uses: ./.github/workflows/jira_server_ci.yml
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
adehad, line 15: conditional here is changed correctly in #2340

PR #2347, opened 2025-04-29
fix: fixes #2346 fields input argument for search should not change behaviour

### LINKED ISSUE #2346: enhanced_search_issues() does not support a list of fields to be provided
### Bug summary

I believe your commit in https://github.com/pycontribs/jira/pull/2334/files#diff-442d14b20d591afd0c3de830d2803e30d4816eedc708f9ad129316aa7e5a4a7cR3714 is causing an unexpected bug

Specifically, [fields](https://github.com/pycontribs/jira/blob/87a697566849daac3556b0ca8bed0ebd66823078/jira/client.py#L3713C24-L3713C52) supports a list of strings but the logic that was added to `enhanced_search_issues()` empties this argument if a list is actually provided:
```
        if fields is None:
            fields = ["*all"]
        elif fields and isinstance(fields, str):
            fields = fields.split(",")
        else:
            # this is required only for mypy validation
            fields = []
``` 

### Jira Instance type

### Jira instance version

_No response_

### jira-python version

3.10.0

3.12

- [ ] Linux
- [x] macOS
- [ ] Windows

### Reproduction steps

```python
I can access `fields` object if I use jira v3.8.0, if I manually access the new API outside of this library and if I use a comma separated list of values to pass to `enhanced_search_issues()`, but not if I provide a list of values
```

### Stack trace

```python
N/A
```

### Expected behaviour

something along the lines of:

```
        if fields is None:
            fields = ["*all"]
        elif fields and isinstance(fields, str):
            fields = fields.split(",")
        elif fields and isinstance(fields, list):
            fields = fields
        else:
            # this is required only for mypy validation
            fields = []
```

where you accept fields being a list and don't just empty it

### Additional Context

_No response_

studioj:
this changed the behvaiour of the api thx for reporting @glenn-sq

just as a reference I'll tag the original creator, so they get informed (its git blame, not bug blame ;-) ) thx for the contributions either way @beliaev-maksim they were helpful

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -3721,13 +3721,10 @@ class JIRA:
         Returns:
             Union[Dict, ResultList]: JSON Dict if ``json_result=True``, otherwise a `ResultList`.
         """
-        if fields is None:
-            fields = ["*all"]
-        elif fields and isinstance(fields, str):
+        if isinstance(fields, str):
             fields = fields.split(",")
-        else:
-            # this is required only for mypy validation
-            fields = []
+        elif fields is None:
+            fields = ["*all"]
 
         untranslate = {}  # use to add friendly aliases when we get the results back
         if fields:


jira/client.py
```diff
@@ -3721,13 +3721,10 @@ def enhanced_search_issues(
...
-        if fields is None:
-            fields = ["*all"]
-        elif fields and isinstance(fields, str):
+        if isinstance(fields, str):
             fields = fields.split(",")
-        else:
-            # this is required only for mypy validation
-            fields = []
+        elif fields is None:
+            fields = ["*all"]
```
studioj, line 3727: this is the behaviour of the old search_issues

PR #2365, opened 2025-08-19
Remove unreliable libmagic dependency and broken use of mimetypes

### LINKED ISSUE #1801: Make dependency on pillow optional

kohtala:
The magic package for libmagic support has many different implementations that conflict with each other and I believe do not work on Windows.

The fallback attempt to use `mimetypes` on a `tempfile` does not work because on Linux the temporary file is created with `O_TMPFILE` and the file `name` is integer for the file descriptor. On Windows it works no better because it relies on the file name extension and a temporary file does not have an extension set based on the content.

For the avatar only JPEG, GIF and PNG are supported in [Jira Could REST API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-avatars/#api-rest-api-3-universal-avatar-type-type-owner-entityid-post) and also BMP and WBMP for [Data Center](https://docs.atlassian.com/software/jira/docs/api/REST/9.14.0/#api/2/project-storeTemporaryAvatar). Because we have only these types to recognize, it is easier for us to decide which it is ourselves than using a dependency.

This replaces solutions in #1847 #1867, #1891 and closes #1801 which should have been closed already.


jira/client.py
```diff
@@ -4631,49 +4623,32 @@ def _find_for_resource(
...
-            return mimetypes.guess_type(f.name)[0]
-        except (OSError, TypeError):
-            self.log.warning(
-                "Couldn't detect content type of avatar image"
-                ". Specify the 'contentType' parameter explicitly."
-            )
-            return None
+        # We assume the image is one of supported formats
+        # https://docs.atlassian.com/software/jira/docs/api/REST/9.14.0/#api/2/project-storeTemporaryAvatar
+        if buff[:3] == b"\xff\xd8\xff":
```
studioj, line 4637: did you investigate teh filetype package?

kohtala: Yes. It is mentioned in #1847 with a concern about it's maintenance. Also #1867 raises several concerns. On #1867 @ssbarnea commented on a plan to review and remove dependencies.

Jira supports a limited set of file types that are quite simple to tell apart. I expect any other MIME-types to be rejected by Jira. If Jira does not validate input, it could be a potential security issue on Jira.

Therefore I find that any dependency with wider support for file types does not add anything beyond this simple MIME-type discovery.

In my opinion, when use of a dependency is a larger and more complex problem than the root problem being solved, the solution does not need to be delegated to the dependency.

Do you know of any Jira instance that would accept any other MIME types to warrant a dependency that supports lots of MIME types?

kohtala:
I updated the description to be clearer on what are the problems in the current implementation and why this solution covers all the use cases we can support.

PR #2376, opened 2025-09-13
enable CI on cloud again (server seems to be pretty hard tbh)

studioj:


.github/workflows/jira_cloud_ci.yml
```diff
@@ -39,7 +44,8 @@ jobs:
     steps:
       - uses: actions/checkout@v5
         with:
-          ref: ${{ github.event.inputs.ref }}
+          # Use correct ref for PRs and pushes
+          ref: ${{ github.head_ref || github.ref_name || github.event.inputs.ref }}
```
kohtala: `github.head_ref` is the pull request source branch, not the code that is requested for merge. Seems to me you are checking out the wrong version. Do you really need this or the `inputs.ref` at all? I'd expect it goes for the correct commit without the `ref`. If the input is needed, it's better available from [inputs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#inputs-context) context as `inputs.ref`, no need to go for the `github.event.inputs`.

.github/workflows/jira_cloud_ci.yml
kohtala: You can add [concurrency](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idconcurrency) to prevent this workflow from running concurrently.

```
concurrency:
  group: jira-cloud-tests
  cancel-in-progress: false
```

.github/workflows/jira_cloud_ci.yml
```diff
@@ -1,6 +1,11 @@
 name: cloud
 
 on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - "*"
```
kohtala: Isn't this the same as not specifying branches at all? Or do you plan to change it for `main` only?

```
on:
  push:
```

.github/workflows/jira_cloud_ci.yml
studioj: ah thx for your review, I hope with your feedback i can get this ball rolling again :-)

.github/workflows/jira_cloud_ci.yml
studioj: i think I'll better do it on pull request only. New to the GHA world :-) Jenkins is my original thing. Thx

.github/workflows/jira_cloud_ci.yml
studioj: some stuff is rolling, but it seems other stuff was broken ... this will need a lot more tlc :S

adehad (dismissed):
I think it may make sense to not try and run the server tests for now.

If we are able to push a local image that runs, that might be an option?

Or maybe a preferred option would be to ask Atlassian to similarly get a server instance setup for us? Likely being able to use: https://hub.docker.com/r/atlassian/jira-software - if they give us a license key that we can use in our CI or at some free tier in a self hosting platform?

lipkau:
Hey.

I am the maintainer of a similar packages but for Powershell and wanted to draw inspiration from your repository on how to test on Jira Datacenter.
Unfortunately you didn't have the answer for me, but a long sessions with Opus found the answer and I want to share it with you: https://github.com/AtlassianPS/JiraPS/pull/623

I managed to get it to work with `moveworkforward/atlas-run-standalone:${JIRA_IMAGE_TAG:-jira-11}`

Good luck 🤞

PR #2383, opened 2025-10-01
feat: add notify option for add_worklog

yajoud-gr:
(cherry picked from commit 2306fde34f99710c560f21b1d16ce10defe74d3d)


yajoud-gr:
Adds a ’notify’ option for ’add_worklog’

PR #2396, opened 2025-10-31
chore: Adds some missing type annotations

schuylermartin45:
Adds a few missing type annotations I found while working on another project. `mypy` flagged most of these with `no-untyped-call`. The others I found while updating the functions flagged by `mypy`.


schuylermartin45:
Okay, I'm unclear what part I'm missing in the `Conventional Commits` doc that is preventing `chore:` from being recognized

Issue #2382, opened 2025-09-29
JIRA api /rest/api/2 depreciated. Needs path updated


rdsim8589:
### Bug summary

I got this error message when trying to access the JIRA end point

`response text = {"errorMessages":["The requested API has been removed. Please migrate to the /rest/api/3/search/jql API. A full migration guideline is available at https://developer.atlassian.com/changelog/#CHANGE-2046"],"errors":{}}`

>  What is changing?
> 
> We’re deprecating the following Jira Platform REST endpoints:
> 
> GET /rest/api/2|3|latest/search -  [Search for issues using JQL (GET)](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get)
> 
> POST /rest/api/2|3|latest/search -  [Search for issues using JQL (POST)](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-post)
> 
> POST /rest/api/2|3|latest/search/id - [Search issue IDs using JQL](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-id-post)
> 
> POST /rest/api/2|3|latest/expression/eval - [Evaluate Jira expression](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-jira-expressions/#api-rest-api-3-expression-eval-post) 
> 
> There is a deprecation period of 6 months, and these endpoints will be removed after May 1, 2025.
> 

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Jira Instance type

Jira Cloud (Hosted by Atlassian)

### jira-python version

main, 3.10.5

### Python Interpreter version

3.12

### Which operating systems have you used?

- [x] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
# 1. using the `search_issues` method of the JIRA client.py file. The trace stack was triggered.
```

### Stack trace

```python
File "/home/rsim/.local/share/virtualenvs/new_srampy-n6qhSQiD/lib/python3.12/site-packages/jira/client.py", line 3557, in search_issues
    issues = self._fetch_pages(
             ^^^^^^^^^^^^^^^^^^
  File "/home/rsim/.local/share/virtualenvs/new_srampy-n6qhSQiD/lib/python3.12/site-packages/jira/client.py", line 817, in _fetch_pages
    resource = self._get_json(
[...]

PR #2405, opened 2025-12-12
ISSUE-23820 update default rest api version to 3

SamuelLKane:
Related to https://github.com/pycontribs/jira/issues/2382

Started running into this issue as part of our development cycle today. Discovered that the default option for `rest_api_version` was still set to `2`. This PR updates it to the required version `3` as specified in the migration document from [Atlassian](https://developer.atlassian.com/changelog/#CHANGE-2046)


SamuelLKane:
Actions bot appears to be confused about conventionalcommits too 😢

PR #2416, opened 2026-03-28
fix: createmeta issuetypes endpoint incorrectly blocked on Jira Cloud

### LINKED ISSUE #2353: createmeta issuetypes not supported with Jira Cloud server
### Bug summary

In https://github.com/pycontribs/jira/blob/main/jira/client.py#L2130:

A check for _is_cloud() rejects accessing the createmeta issutypes endpoint.
According to https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-group-issues:
the Cloud server should support that endpoint.

Are we sure we need to reject access here?

We are using the code in https://github.com/HeinleinSupport/check_mk_extensions/tree/cmk2.3/notifications/jira_heinlein and we find it does not work with jira cloud:

versionNumbers: 1001, 0, 0
developmentType: "Cloud"
buildNumber: 100283

### Jira Instance type

### Jira instance version

1001.0.0.0

### jira-python version

main

3.10

- [x] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
jira: JIRA
jira.the_function(x)
# 3.
...
```

### Stack trace

```python
CRITICAL;jira_heinlein;Unable to query meta info, JIRA response code None, Unsupported JIRA deployment type: Cloud or version: (1001, 0, 0). Use 'createmeta' instead.;Unable to query meta info,
```

### Expected behaviour

The probably works fine with a recent cloud server.

### Additional Context

_No response_

star7js:
## Summary
- `_check_createmeta_issuetypes()` was blocking Jira Cloud deployments entirely with `self._is_cloud or version < (8,4,0)`
- The [Jira Cloud REST API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-createmeta-projectidorkey-issuetypes-get) supports this endpoint, so Cloud should be allowed
- Changed to `not self._is_cloud and version < (8,4,0)` — Cloud always passes, only old Server/DC (< 8.4.0) is blocked
- This fix applies to all 4 callers of the check: `createmeta_issuetypes`, `createmeta_fieldtypes`, `project_issue_types`, and `project_issue_fields`

Closes #2353


star7js:
Hi maintainers — CI is failing because the `fix` label wasn't auto-applied by the conventional-release-labels action. Could someone please add the `fix` label so the ack check passes? Thanks\!

PR #2441, opened 2026-07-29
fix(client): support bytes in add_attachment() (#1735)

### LINKED ISSUE #1735: Regression in add_attachment(), fails on bytes object
### Bug summary

When a `bytes` object is passed as the attachment argument to `add_attachment()` it raises an AttributeError. This was not the case before #1366 (first seen in v3.3.0).

### Jira Instance type

### Jira instance version

v8.9.0#809000-sha1:4ceb90a

### jira-python version

3.5.2

Python 3.11.4 (main, Jun  9 2023, 07:59:55) [GCC 12.3.0] on linux

- [X] Linux
- [ ] macOS
- [ ] Windows

### Reproduction steps

```python
from jira import JIRA

jira = JIRA(basic_auth=("admin", "admin"))
jira.add_attachment(
    issue="QQ-1",
    attachment=b"example content",
    filename="example.txt",
)
print("ok")
```

### Stack trace

```python
  File "/home/bvl/dev/scratch/./test_regression_1366.py", line 4, in <module>
    jira.add_attachment(
  File "/home/bvl/dev/scratch/jirra/venv/lib/python3.11/site-packages/jira/client.py", line 123, in wrapper
    result = func(*arg_list, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bvl/dev/scratch/jirra/venv/lib/python3.11/site-packages/jira/client.py", line 1032, in add_attachment
    encoded_data, request_headers = generate_multipartencoded_request_args()
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bvl/dev/scratch/jirra/venv/lib/python3.11/site-packages/jira/client.py", line 1009, in generate_multipartencoded_request_args
    attachment_io.seek(0)
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'bytes' object has no attribute 'seek'
```
```


### Expected behaviour

The attachment will be present on the issue, with the given content and filename.
```bash
$ python ./test_regression_1366.py
ok
```

### Additional Context

_No response_

Mukller:
## Problem

Fixes #1735.

Passing a `bytes` object to `add_attachment()` raises `AttributeError` since v3.3.0:

```python
jira.add_attachment(issue='QQ-1', attachment=b'example content', filename='example.txt')
# AttributeError: 'bytes' object has no attribute 'seek'
```

## Root Cause

The regression was introduced in #1366. Before that PR the attachment was passed directly to `requests`. After #1366, the upload goes through a `MultipartEncoder`, and `generate_multipartencoded_request_args()` calls `attachment_io.seek(0)` to reset the stream position before encoding.

When `attachment` is `bytes`, the `else` branch sets `attachment_io = attachment` (the `bytes` object itself), and `bytes` has no `.seek()` method.

## Fix

Add an explicit `bytes` branch that wraps the content in `io.BytesIO`:

```diff
-from io import BufferedReader
+from io import BufferedReader, BytesIO

 def add_attachment(self, issue: str | int,
-                   attachment: str | BufferedReader,
+                   attachment: str | bytes | BufferedReader,
                    filename: str | None = None) -> Attachment:
     close_attachment = False
     if isinstance(attachment, str):
         attachment_io = open(attachment, 'rb')
         close_attachment = True
+    elif isinstance(attachment, bytes):
+        attachment_io = BytesIO(attachment)
+        close_attachment = True
     else:
         attachment_io = attachment
         ...
```

`BytesIO` is seekable, so `attachment_io.seek(0)` works. `close_attachment = True` ensures the `BytesIO` is closed in the `finally` block (we created it, so we're responsible for it).

## Notes

- When passing `bytes`, a `filename` argument is required: `BytesIO` has no `.name` attribute, so the auto-name fallback (`os.path.basename(attachment_io.name)`) is skipped. This was already the case for any non-`BufferedReader` file-like object.
- `BytesIO` is accepted by `MultipartEncoder` for the same reason `BufferedReader` is: both are file-like objects implementing `read()` and `seek()`.

## Verification

```python
from jira import JIRA

jira = JIRA(basic_auth=('admin', 'admin'))
att = jira.add_attachment(
    issue='QQ-1',
    attachment=b'example content',
    filename='example.txt',
)
print(att)  # Attachment: example.txt — no more AttributeError
```

### DIFF
--- a/jira/client.py
+++ b/jira/client.py
@@ -25,7 +25,7 @@ import warnings
 from collections import OrderedDict
 from collections.abc import Iterable, Iterator
 from functools import cache, wraps
-from io import BufferedReader
+from io import BufferedReader, BytesIO
 from numbers import Number
 from typing import (
     Any,
@@ -1119,7 +1119,7 @@ class JIRA:
     def add_attachment(
         self,
         issue: str | int,
-        attachment: str | BufferedReader,
+        attachment: str | bytes | BufferedReader,
         filename: str | None = None,
     ) -> Attachment:
         """Attach an attachment to an issue and returns a Resource for it.
@@ -1139,6 +1139,9 @@ class JIRA:
         close_attachment = False
         if isinstance(attachment, str):
             attachment_io = open(attachment, "rb")  # type: ignore
+            close_attachment = True
+        elif isinstance(attachment, bytes):
+            attachment_io = BytesIO(attachment)
             close_attachment = True
         else:
             attachment_io = attachment
--- a/jira/client.py
+++ b/jira/client.py
@@ -1137,8 +1137,9 @@ class JIRA:
             Attachment
         """
         close_attachment = False
+        attachment_io: BytesIO | BufferedReader
         if isinstance(attachment, str):
-            attachment_io = open(attachment, "rb")  # type: ignore
+            attachment_io = open(attachment, "rb")  # type: ignore[assignment]
             close_attachment = True
         elif isinstance(attachment, bytes):
             attachment_io = BytesIO(attachment)
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -413,3 +413,31 @@ def test_experimental_non_200_not_404_405(
 
     assert ex.value.status_code == status_code
     assert isinstance(ex.value, JIRAError)
+
+
+class _BareAttachmentClient(jira.client.JIRA):
+    """A client without __init__ wiring, enough for add_attachment unit tests."""
+
+    def __init__(self, session):
+        self.log = logging.getLogger("test_add_attachment")
+        self._options = dict(jira.client.JIRA.DEFAULT_OPTIONS)
+        self._options["server"] = "https://mocked.jira.invalid"
+        self._session = session
+
+
+def test_add_attachment_accepts_bytes():
+    content = b"example content"
+    mock_session = mock.Mock()
+    mock_session.post.return_value.status_code = 200
+    mock_session.post.return_value.json.return_value = [
+        {"id": "10001", "filename": "example.txt", "size": len(content)}
+    ]
+    client = _BareAttachmentClient(mock_session)
+
+    attachment = client.add_attachment("QQ-1", content, "example.txt")
+
+    assert attachment.size == len(content)
+    posted_kwargs = mock_session.post.call_args.kwargs
+    body: bytes = posted_kwargs["data"].read()
+    assert content in body
+    assert posted_kwargs["headers"]["X-Atlassian-Token"] == "no-check"


Mukller (commented):
## Code Review

### Bug Trace

```
add_attachment(issue='X', attachment=b'data', filename='f.txt')
  close_attachment = False
  isinstance(b'data', str) → False  # not a file path
  else:
      attachment_io = b'data'        # bytes object assigned

  generate_multipartencoded_request_args()
      attachment_io.seek(0)          # AttributeError: bytes has no seek()
```

### Fix Correctness

```python
elif isinstance(attachment, bytes):
    attachment_io = BytesIO(attachment)  # seekable; MultipartEncoder accepts it
    close_attachment = True              # we created it, we close it
```

| Case | Before | After |
|---|---|---|
| `str` (file path) | Works ✓ | Unchanged ✓ |
| `bytes` | `AttributeError` ✗ | Wrapped in `BytesIO` ✓ |
| `BufferedReader` | Works ✓ | Unchanged ✓ |
| Other file-like | Works ✓ | Unchanged ✓ |

### `close_attachment` for BytesIO

We own the `BytesIO` (we created it), so setting `close_attachment = True` is correct — the `finally` block will close it. This mirrors the treatment of the `str` (file path) case.

### `fname` Behaviour

`BytesIO` has no `.name`, so the auto-name fallback:
```python
if not fname and isinstance(attachment_io, BufferedReader):
    fname = os.path.basename(attachment_io.name)
```
is not triggered (because `BytesIO` is not a `BufferedReader`). This is unchanged from the existing behaviour for all non-`BufferedReader` file-like objects; callers must supply `filename` when passing bytes.

### Scope

- `BytesIO` added to existing `from io import ...` line.
- Function signature: `bytes` added to the union type.
- 2 lines of logic added in `add_attachment`. No other changes.

Mukller:
Fixed the mypy failure:

The `attachment_io` variable was implicitly typed as `BufferedReader` (from the `open()` call in the first branch), so mypy rejected the `BytesIO` assignment in the `elif isinstance(attachment, bytes)` branch.

Added an explicit type annotation `attachment_io: BytesIO | BufferedReader` before the if/elif/else chain and narrowed the `# type: ignore` comment on the `open()` call to `# type: ignore[assignment]`.

Mukller:
Added a unit test for this in 8d409a2 so the fix no longer relies on the integration suite:

`test_add_attachment_accepts_bytes` builds a bare client over a mocked session, calls `add_attachment(issue, b"example content", "example.txt")`, and asserts that the bytes payload actually reaches the multipart body (`content in posted.data.read()`) along with the `X-Atlassian-Token: no-check` header.

Proof it's a proper regression test: it fails on current main with exactly the reported `AttributeError: 'bytes' object has no attribute 'seek'` and passes on this branch. The rest of `tests/test_client.py` stays green locally (the 7 collection errors are the usual live-server tests gated on `CI_JIRA_URL`).

